PR #3, opened 2016-05-27
Make all compile with cmake_modules in both locations

jochym:
This is done by making CMAKE_MODULE_PATH a list with both (in-dir and at the top) locations present.
All drivers compiled on test system within separate build directory.
Using:
cmake . ../3rdparty/
and
cmake . ../libindi/


knro:
Thanks!

PR #6, opened 2016-05-28
Telescope Scripting Gateway added

polakovic:


knro:
This is interesting, it is useful in case there is no native INDI driver? or what case scenarios?

Also, please update /usr/local/share/indi to /usr/share/indi, any "local" stuff will only introduce problems.

polakovic:
Hi Jasem,

yes, this is for Greg Schwimer and situations like this:
http://indilib.org/forum/mounts/1267-bisque-paramount-indi-driver-for-tsx-tcp-api.html?start=12#8382
.

I'll change default location later, for OSX is /usr/local the only choice,
but can be different for Linux.

Peter

polakovic:
Jasem, is it possible to remove properties defined by base class (e.g. port
or baud rate is not really needed in this case).

PR #7, opened 2016-05-28
New temp file name for each status check, /usr/shared/indi/scripts us…

polakovic:
…ed for linux

### DIFF
--- a/libindi/drivers/telescope/telescope_script.cpp
+++ b/libindi/drivers/telescope/telescope_script.cpp
@@ -28,7 +28,6 @@
 #include "telescope_script.h"
 
 #define	POLLMS      1000
-#define STATUS_FILE "/tmp/indi-telesope-gateway.status"
 
 enum { SCRIPT_CONNECT = 1, SCRIPT_DISCONNECT, SCRIPT_STATUS, SCRIPT_GOTO, SCRIPT_SYNC, SCRIPT_PARK, SCRIPT_UNPARK, SCRIPT_MOVE_NORTH, SCRIPT_MOVE_EAST, SCRIPT_MOVE_SOUTH, SCRIPT_MOVE_WEST, SCRIPT_ABORT } scripts;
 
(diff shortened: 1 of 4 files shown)


libindi/drivers/telescope/telescope_script.cpp
```diff
@@ -28,7 +28,6 @@
 #include "telescope_script.h"
 
 #define	POLLMS      1000
-#define STATUS_FILE "/tmp/indi-telesope-gateway.status"
```
jochym: Wouldn't that be better somewhere in the /var/run? This is essentially a communication point, isn't it?

libindi/drivers/telescope/telescope_script.cpp
```diff
@@ -79,7 +78,11 @@ const char * ScopeScript::getDefaultName() {
 bool ScopeScript::initProperties() {
   INDI::Telescope::initProperties();
   
+#ifdef OSX_EMBEDED_MODE
   IUFillText(&ScriptsT[0], "FOLDER", "Folder", "/usr/local/share/indi/scripts");
+#else
+  IUFillText(&ScriptsT[0], "FOLDER", "Folder", "/usr/share/indi/scripts");
```
jochym: I have an issue with sticking things in /usr and not in /usr/local. I know that @knro prefers the first but I don't know why. It is against Debian guidelines (?), and it messes up with package management. Jasem, would you care to comment?

knro:
I am very aware it is against Debian guidelines. But users who install INDI stuff in /usr/local **always** end up in conflict with what is installed in /usr, and they often forget the installation in /usr/local at some point.

libindi/drivers/telescope/telescope_script.cpp
knro: How about INDI FIFO? I often create it in /tmp, would it better in /var/run ?

libindi/drivers/telescope/telescope_script.cpp
knro: Regarding the location as well, INDI won't be shipping any scripts, they're always user-provided. So in reality maybe it doesn't matter where they end up, and the user can change the folder anyway.

jochym:
You obviously have more experience with this. But right now we are not hardcoding the path to the installation.  Making it os-dependent adds to the confusion IMHO. Maybe we should use the installation location instead ?

libindi/drivers/telescope/telescope_script.cpp
jochym: As I understand both locations are OK. But /var/run tends to be used more recently. It is often in memory and non-persistant.

libindi/drivers/telescope/telescope_script.cpp
polakovic: That's right, any script itself should be prepared by the user, they definitely have persistent nature and /var is not a right choice. But this is just default value and I suppose it will be changed by the user anyway.

knro:
Let me merge this as I have immediate pending fixes as well

polakovic:
Installation location can't be used for anything, what can be modified on OSX, whole bundle is digitally signed.

knro:
Use of tmpnam(NULL) is generating a gcc warning "Use of tmpnam is dangerous", is there a better solution?

jochym:
Ok. So maybe it was good idea by @polakovic to have user-supplied scripts in /usr/local... Maybe we have grown large enough to have some config in /etc ?

knro:
Config already stored under ~/.indi

polakovic:
Use of tmpnam is dangerous, but in this case it is under control :) ~/.indi is an option, but I have a feeling, that in this case it is systemwide and not personal setting.

PR #10, opened 2016-05-29
Shelyak eShel driver

sholmbo:
Tested and working.

### DIFF
--- a/3rdparty/indi-shelyak/CMakeLists.txt
+++ b/3rdparty/indi-shelyak/CMakeLists.txt
@@ -1,0 +1,23 @@
+cmake_minimum_required(VERSION 2.4.7)
+project(indi_shelyak CXX C)
+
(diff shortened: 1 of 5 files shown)


knro:
Thanks! The driver XML file is missing. Can you also write some details we can put in the INDI devices page with screenshots..etc? Something like this: http://indilib.org/devices/focusers/focuslynx.html

jochym:
It will be also useful if you would add the driver build to .travis.yml - that way your driver will be automatically tested before merging - it builds on pull request.

sholmbo:
@knro I'll look at it, but it won't be as detailed as that example.
@jochym I don't know this Travis-CI. Am I right when I think that I just have to add it to 3rdparty/CMakeLists.txt? By the way, I think line 40 should be `option(WITH_ARDUINO "Install Arduino Driver" On)`, right?

jochym:
@sholmbo Just check if it builds in standard way (the same the other 3rdparty packages build). I will add the build myself to travis.yml

PR #24, opened 2016-06-02
New version of Pulsar 2 driver (still beta)

ptanga:
Much improved version  of the Pulsar 2 driver. Functional, but still work in progress.


knro:
Thanks, you should probably send a message upon startup to warn the user that is is Beta.

knro:
I updated the driver to use the INDI Standard Property TELESCOPE_PIER_SIDE

PR #29, opened 2016-06-12
Introduce the beginnings of unit test framework

A-j-K:
Brings in new INDI::TTY interface.
Baader dome first to receive refactor to consume this IF using DI (see driver class constructor).

Adds test directory and first worked examples.

Adds Google Test and Google Mock to automated build (see extension work done on Travis-CI)

https://travis-ci.org/A-j-K/indi/builds/137110031


libindi/indidrivermain_gbls.c
```diff
@@ -0,0 +1,40 @@
...
+
+#define MAXRBUF 2048
+
+ROSC *roCheck;
+int nroCheck;			/* # of elements in roCheck */
+
+int verbose;			/* chatty */
+char *me;				/* a.out name */
+LilXML *clixml;			/* XML parser context */
+
```
A-j-K: Just in case you are wondering why this was done. First up, over time I will refactor to remove the use of globals, they make UT more complex. However, to keep it simple I left them. However, all drivers need these four globals. But they are defined in the same file that defines main(). This is a problem as each UT created by Google Test already defines main(). In order to be able to still compile the drivers and have these globals available but use the UT's main() they need to be defined in a file other than indidriver.c otherwise we get two main()

knro:
Thanks that makes sense. How about the INDI::TTY in INDI::Telescope? It is not used in any telescope driver or in INDI::Telescope?

A-j-K:
I started with the Baader dome as a demonstrator. Now this is merged the idea is to slowly go through each driver a refactor to use INDI::TTY and slowly remove indicom.c entirely.

PR #31, opened 2016-06-16
Update to latest ZWO provided SDK and add support for newer cameras.

goodwink:
I've pulled in the latest SDK from ZWO and added support for the following cameras:

- ASI1600M(M|C)(-Cool)?
- ASI290M(M|C)(-Cool)?


goodwink:
I have only tested this on linux x64, but the libraries are provided upstream and I haven't changed any other code, so it shouldn't make any difference.

knro:
I believe ZWO is working on fixing an issue on Raspberry PI as the image is not downloaded on full native resolution via libusb. Is this issue fixed with this binary? Also please make sure that all the text files have unix EOL. use dos2unix tool to convert all files to unix end-of-line first.

goodwink:
I have not tested on arm, I will have to check with Sam.  I'll run dos2unix on ZWO's header file.

knro:
ZWO already submitted a PR.

goodwink:
@knro that PR did not update the udev rules, so the new cameras will only be usable by root.

knro:
Please submit a new PR for the udev rules.

knro:
Nevermind, I applied the diff to the udev rules.

PR #34, opened 2016-06-20
NSE: Corrected guiding rate scale and simulator for development

jochym:
This PR contains small fix to the NSE driver and a hardware simulator complete enough for the tablet celestron application (SkySafari 4) to work as with the real mount. The simulator is a developers tool, and is not intended for the end-users. It is written in python and requires python 3.5 (uses async/await asyncio syntax). All commands used by the vendor application are identified and implemented. Also the auto-detection protocol is implemented in the simulator. The implementation of all the corresponding parts in the driver (autodetection, settings and mount monitoring) will follow shortly.


knro:
The simulator can't be built-in in the driver like EQMod and other drivers do it?

jochym:
The device is an intrinsically networking one - and the important part of the simulation is the actual network protocol (it uses at least two ports and UDP  as well as TCP packets). Build-in simulator may implement this, but will be much more complicated since we will need to implement two asynchronous network servers into it - in c++ it is not entirely trivial task, and will unnecessarily complicate the driver for no end-user benefit. I may port the command-parsing part of the simulator into the driver at the later time - right now the separate python implementation is so much easier to experiment and adapt that there is really no comparison. If you dislike it very much I can remove it from the tree and develop it separately - but it *is* quite useful for development. Remember that NSE driver is a reverse-engineered one. Having separate simulator allows one to run it against the vendor applications and learn from them - actually I was able to decipher many unknown AUX commands that way. I plan to add the rest of my development tools to the tree as well - that way there is at least a chance that somebody except me will start hacking this driver. 
What are the arguments in favour of build-in simulator?

knro:
We had a long discussion on Gitter on this very topic. The plan is to use Google test and mock frameworks to unit test the code and create mocks to check if the code confirms to the vendor specifications. 

Would it be possible that your driver fully makes use of them and it would serve as an example for other drivers in the future?

jochym:
Not a bad idea. The driver is actually quite simple. Much simpler then eqmod, due to the high "inteligence" of motor controllers in celestron scopes and relatively high level of AUX command set. The most of the heavy lifting is done by the alignment module. It is beta, so would be probably a good place for such development. The problem is, I am not familiar with this testing framework and I am pretty rusty C++ coder - most of my day-to-day work is python or fortran (!). I do not feel I can implement this by myself alone. My current priorities are bringing the driver to the point I can recommend it's use with clear conscience. Right now I finally have full set of protocol specs I need for this. The immediate goal is to implement all known mount settings and connection auto-detection/setup which I finally discovered. Next will be review of the communication layer to improve its robustness and probably factoring it out into separate abstraction layer to facilitate extending the set of target devices to mounts with other types of connections to the AUX port (direct serial, hand controller 0x50 bridge command, additional SkyFi dongle, others ?).

A-j-K:
NexStarAUXScope.h is an ideal target for Google Mock so the structure of this 3rd party module is reasonably ideal to add unit tests. However, unit tests aren't a development tool like the simulator. So, on the one hand you would have a pretty good opportunity here to demonstrate unit testing. On the other hand, however, that unit testing doesn't replace the simulator that other developers can use. They are two different tools.

jochym:
I do understand the difference. I am familiar with the *idea* of unit tests - only not with this particular framework (I have followed the discussion). I agree that the simulator is quite a different tool with different use-case. The question is @A-j-K, are you or somebody else willing to help me with this?

jochym:
@knro Are we/you considering merge of the PR in present form? If no i will resubmit it as just a fix to the driver. But let me just say I would rather prefer to include the tools in the repo to facilitate possible development of the driver by other developers.

knro:
Yes sure. Your project is a good candidate for the testing/mock frameworks now that we're working on it.

A-j-K:
@jochym yes, willing to help

jochym:
Great. Maybe we should move the discussion to the gitter: https://gitter.im/jochym/indi-unittests?utm_source=share-link&utm_medium=link&utm_campaign=share-link

PR #35, opened 2016-06-21
Refactor/unittest

A-j-K:
The is the beginnings of adding a Unit Test suite based on Google Test Framework. Please note, as you will see I started with the lowest hanging fruit, base64. It will be a good exercise you you guys to say add test_fq.cpp to test the fifo code as a first go.

Note, all tests are C++ and not C. So "Cisms" such as:-

`char *foo= malloc(5);`

will fail to compile as C++ type-checks, this is what you need:-

`char *foo = (char*)malloc(5); // <-- you need to cast`

I added a note about building for unit tests to the end of the README.md but please see .travis.yml and the travis-ci/ directory.

https://travis-ci.org/A-j-K/indi/builds/139331740

The first test I wrote uses Google Test. I will introduce Google Mock as we get more into it but as you will see, GTEST is pretty powerful in it's own right. GMOCK will probably come when I begin the driver tests as a first example.

https://github.com/google/googletest
https://github.com/google/googletest/blob/master/googletest/docs/Primer.md

[edit] It should be noted that, by default, if GTEST suite is not installed then the previous build behaviour remains the same for to ensure builds for most devs not involved so far are not inconvenienced. Likewise, after GTEST is installed you still need to provide the cmake -D switch to enable UT builds. This should ensure we don't trample any developers current work status if they rebase from master after this merges.


knro:
Sounds great. Thanks for the fantastic effort!

PR #55, opened 2016-08-15
indilib make base64 encoding 2x faster

rumengb:
Complete rewrite of to64frombits() - gives 2x the performance of the old implementation.
Here are some test results on RPI3:
memcmp of two encoded buffers: 0 (len1 = 53333336 len2 = 53333336)
Base64 encode (INDI) took 18955ms
Base64 encode (FAST) took 8155ms
FAST took 43.0% of the time that INDI used.

### DIFF
--- a/libindi/base64.c
+++ b/libindi/base64.c
@@ -2,6 +2,9 @@
     INDI
     Copyright (C) 2003 Elwood C. Downey
 
+    Complete rewrite of to64frombits() - gives 2x the performance
+    of the old implementation (Aug, 2016 by Rumen G.Bogdanovski)
+
     This library is free software; you can redistribute it and/or
     modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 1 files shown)


knro:
Wow impressive work Rumen! Will give it a go

knro:
Is it possible to obtain any performance gains from from64tobits as well?

knro:
Actually what do you think of https://github.com/aklomp/base64 ?

rumengb:
I think it is... I will have to try. One solution i know of will require a bit larger LUT and one can hardly beat it performance wise but it will require 128kB LUT. I can check what. Probably we will get 2x from decode too but at the price of 128k in RAM. Which is not a problem i believe if we deal with > 20mb images. i have no idea of  https://github.com/aklomp/base64, it looks quite optimized for dofferent architectures but I think LUTs will be faster anyway :) We can test it...

PR #56, opened 2016-08-15
libindi: fixed fast base64 encoding - the last byte was wrong sometimes

rumengb:
Jasem there was a wried corner case in the code when one "=" had to be added the last byte was off by 2... I think now it is fixed. I wrote  a rigorous auto tests which does not fail any more. The good news is that the fix does not affect the performance at all...


knro:
Great! Will give it a real-field test today as well!

PR #57, opened 2016-08-18
filter same exptime and gain settings

qhyccd-lzr:


knro:
Hi Lee the commit conflicts with master branch. Please sync your branch to master first, then you must change all the lines to UNIX End-Of-Line. Use dos2unix tool to covert all text files (*.h, *.cpp), do not apply it to binary files like .hex and .img

Then resubmit another Pull Request which hopefully should be without conflicts.

PR #61, opened 2016-08-26
greatly improve the performance of from64tobits()

rumengb:
* Removed "\n" from BLOBs to maximize decoding performance.
* Replaced fprintf with slightly faster fwrite when writing BLOBs as we do not need formating.
* Added  from64tobits_fast()  which requires BLOB length (for max performance)
* Kept from64tobits() interface which has a bit reduced performance compared to from64tobits_fast().  
* Ekos seems to call  from64tobits() directly (maybe needs to change to _fast version). 

IMPORTANT NOTE! The current client library will not work with old servers (from64tobits() does not allow "\n" any more to maximize performance). However the new server will work with old clients.


rumengb:
This is what I get on my computer:

rumen@rumen-VirtualBox:~/indi-rgb/libindi/build/test$ ctest -V
UpdateCTestConfiguration  from :/home/rumen/indi-rgb/libindi/build/test/DartConfiguration.tcl
UpdateCTestConfiguration  from :/home/rumen/indi-rgb/libindi/build/test/DartConfiguration.tcl
Test project /home/rumen/indi-rgb/libindi/build/test
Constructing a list of tests
Done constructing a list of tests
Checking test dependency graph...
Checking test dependency graph end
test 1
    Start 1: test_base64

1: Test command: /home/rumen/indi-rgb/libindi/build/test/core/test_base64
1: Test timeout computed to be: 9.99988e+06
1: Running main() from gtest_main.cc
1: [==========] Running 2 tests from 1 test case.
1: [----------] Global test environment set-up.
1: [----------] 2 tests from CORE_BASE64
1: [ RUN      ] CORE_BASE64.Test_to64frombits
1: [       OK ] CORE_BASE64.Test_to64frombits (0 ms)
1: [ RUN      ] CORE_BASE64.Test_from64tobits
1: [       OK ] CORE_BASE64.Test_from64tobits (0 ms)
1: [----------] 2 tests from CORE_BASE64 (0 ms total)
1: 
1: [----------] Global test environment tear-down
1: [==========] 2 tests from 1 test case ran. (0 ms total)
1: [  PASSED  ] 2 tests.
1/1 Test #1: test_base64 ......................   Passed    0.00 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.00 sec
rumen@rumen-VirtualBox:~/indi-rgb/libindi/build/test$

rumengb:
Finally figured out what was wrong... and fixed it

knro:
Need to figure out a way to be backward-compatible with older servers as well.

rumengb:
It is easy... I can do that... check if base64[71] == "\n" if so tun another version of decode (~15%) slower. else run the fast. I will do so...  but it is a kludge. We can do it this way and eventually drop the slower version in the future.

(I am still waiting for the on exposure complete exec in Ekos. I will need it in 2 days... Can you do this gr me?)

rumengb:
By the way do we really need this backward compatibility? 
If one upgrades the server only, there will be no problems. We can ask users to upgrade libindi on the server if they intend to upgrade libindi on the client. I am afraid that all these kludges will degrade the performance we gain with the new algorithm. We already have one by keeping the old interface of the decoder that calculates the BLOB length and now we need to add one more on top of that. We may end up reducing the performance by 25-30% in some cases (still faster than the original) which makes all this effort next to useless...

BTW is there any protocol version negotiation performed that we can use to decide which decode function to use?

knro:
I implemented the post script stuff in Ekos, you can find in under KStars options. Please try to implement the start/stop/dither commands for lin_guider whenever you can.

The protocol version is unchanged at 1.7, this change does not affect the protocol per se, but an implementation detail. 

How is this going to degrade performance by 25%, isn't it a single check per blob?

rumengb:
Well it is one check per BLOB and this reduction comes from figuring out the blob len + if there is \n we need to skip them and calculate the BLOB length differently so we need to provide old and new interface to decode that can handle "\n" and no "\n" so this leads to two wrappers for the 2 implementations of the decode. Do not get me wrong, it is easy to do, but the code will become bloated and  the benefit in some cases will be next to none. Ok I will do it. But testing will be a kind of difficult as we have 4 different scenarios that have to be set up. I will do it anyway :)

Is there somewhere a daily build of Kstars?

about in_guider I will implement start and stop or Andrew will, actually I asked him as I am working in my spare time on optimizing indi and also working with ZWO guys trying to make the linux support as stable 
better and I have to say there is a breakthrough there too :)

PR #62, opened 2016-08-29
Remove buffered write of BLOBS some and ASI optimized

rumengb:
* remove buffered write of BLOBS  - did not test the performance compared to buffered version, but fwrite() does an extra copy into the stdio buffer which written at a later time with write() syscall,  so using only write() should be faster - removed one unnecessary copy.

* indi_asi_ccd: increase usbfs_memory_mb to 200mb so that ASI1600 can work with other cameras on the same Linux host, new SDK is on the way in a day or two. ZWO guys did a good job on this.


knro:
Great work Rumen! I need to give it a test today to make sure all in order.

PR #65, opened 2016-08-31
resored '\n' - indi is safe and sound :)

rumengb:
but slower! -yet faster than it was

### DIFF
--- a/libindi/base64.c
+++ b/libindi/base64.c
@@ -98,6 +98,9 @@ int from64tobits_fast(char* out, const char* in, int inlen)
 	uint16_t* inp = (uint16_t*)in;
 
 	for( j = 0; j < n; j++ ) {
+		if (in[0] == '\n') in++;
+		inp = (uint16_t*)in;
+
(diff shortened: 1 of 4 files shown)


rumengb:
I am sorry for breaking the protocol. I understand this is a criminal offense :)
In my defense  I want to show you why I did it:
base64 decode test:
---------------
Base64 decode (INDI) took 7420ms
Base64 decode (FAST) took 1750ms
Base64 decode (FAST no LEN) took 2268ms
Base64 decode (FAST '\n') took 3320ms
Time reduction (FAST) 23.6%
Time reduction (FAST no LEN) 30.6%
Time reduction (FAST '\n') 44.7%

conclusion: instead of 4x times we get 2x...
Enjoy the protocol :)

knro:
Thanks for restoring it. I understand the performance restrictions but we must comply with the INDI Wire protocol. Even 3rd party programs that might nest INDI XML messages (e.g. RTML) can continue to do so because they can read all the content in a standard compliant method. 

Still a 200% performance increase is great! So thanks for working on this!

PR #71, opened 2016-09-11
Build alignment module on all platforms

seanhoughton:
Currently the AlignmentDriver library is only built for "Linux" platforms. However, it's a dependency and must be linked on all platforms causing linker errors. I've removed the platform condition and it builds and links on OSX with no additional changes.


seanhoughton:
Note: there's a forum thread about this at http://indilib.org/forum/mounts/977-major-changes-in-synscan-driver.html?start=6

PR #73, opened 2016-09-13
Add initial support for DSI Color III.

bgilsrud:
This should add support for the DSI Color III. I don't have a DSI Color III to test with, but I have verified that this doesn't break support for my DSI Pro I.


knro:
Anyone who has DSI Color III tested it?

bgilsrud:
Someone in the thread at http://indilib.org/forum/ccds-dslrs/1408-meade-dsi-3.html?start=12#9921 has one, but they use an RPi 3 and I don't have a cross compile enviroment set up. My hope is that they use your ppa and can test now that it's committed. At the very least, DSI III colors work as well as they did before this merge ;).

PR #74, opened 2016-09-14
Remove sign from LX200 set longitude command

bgilsrud:
This change reverts a recent change made for OnStep support. The OnStep firmware was recently changed to support the standard LX200 set longitude command (:SgDDD:MM#). The previous sign change may have broken support for other LX200-protocol mounts that use this code (though I don't have any of these devices to test with).

Adding @azwing in case he has input.


knro:
Roger, the Meade command set does not specify +/- sign. This slipped past me.

PR #77, opened 2016-09-22
Get 9x9 binning mode work for SBIG CCD's

ashindin:
Some little changes to get 9x9 binning work for SBIG STF-8300M CCD


knro:
Thanks, did the 9x9 binning work OK when tested? no issues?

ashindin:
For my STF-8300M it works but I don't really know about the other models

PR #82, opened 2016-10-07
Add a feature to the base indidome to know if the telescope is parked

dokeeffe:
Hi, this is a feature I'd like to have in the base indidome for my roll-off driver.
Basically I'm paranoid about the roof closing and the telescope not being parked. This will be an extra safety check allowing me to use isTelescopeParked() from my driver.

### DIFF
--- a/libindi/drivers/dome/roll_off.cpp
+++ b/libindi/drivers/dome/roll_off.cpp
@@ -222,6 +222,11 @@ IPState RollOff::Move(DomeDirection dir, DomeMotionCommand operation)
             DEBUG(INDI::Logger::DBG_WARNING, "Roof is already fully closed.");
             return IPS_ALERT;
         }
+        else if (dir == DOME_CCW && isTelescopeParked() == false)
+        {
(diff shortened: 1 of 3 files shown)


knro:
This **assumes** there is a valid telescope that can be listened to for parking, and assumes a telescope can park. If the first assumption is false, then telescope parking is always assumed to be false all the times. Same thing for second assumption. Furthermore, it assumes telescope not being parked is problem for the dome driver, which it isn't. I have a roll-off roof and I can safely park the roof without requiring the mount to be parked. Not to mention dome drivers are not affected by this.

You can perhaps move all this logic to rolloff driver without changing INDI::Dome itself, but you need to make sure that unless a telescope indicates that it is UNPARKED then it is assumed to be parked. This way if there is even no scope to listen to, closing the dome would work fine.

knro:
You can also introduce a property to control this behavior. Have it by default to prevent dome closing when mount is parked, but if it is set to OFF, then park the dome anyway.

dokeeffe:
Correct yes, I guess its up to the individual's roof driver if they use it in their implementation or not. 
I'll move it into the roll-off simulator driver for now and add a property too.

PR #83, opened 2016-10-07
Add 'scope park aware' feature to the rolloff simulator

dokeeffe:
Hope this is a bit better than my last effort. Maybe it can help someone else who might be writing a driver based on the roll-off simulator. These changes are going to be added to my own custom roof driver anyway.
Let me know what you think, no worries if not suitable.


knro:
Excellent, this works great (from reading the code), did you test it?

dokeeffe:
I tested manually with the simulator and it works.
Tested happy path and also tested while connected to roof but not scope and with option on and off etc.
I'm applying the same changes to my roof driver right now :-)

PR #92, opened 2016-10-21
Initial support for GigE machine vision cameras.

hbeijeman:
This commit enables initial support for Point Grey machine vision
cameras, and in general, for any camera supported through Project
Aravis.

These cameras are mainly geared towards high speed video streams, and do
not typically support very high exposure times. This driver attempts to
operate the camera in the most manual mode possible. Because most
cameras slave the programmable exposure time to the configured
framerate, which cannot be set lower than 1.0 FPS, out of the box these
cameras do not support longer exposure times, unless the frame rate is
manually disabled.

Unfortunately, this is a vendor-specific operation, here implemented for
the "BlackFly" family from Point Grey.
However, typically, it is not so difficult to set disable the frame
rate. Look in arvcamera.c (in Project Aravis) or BlackFly.cpp for clues
if you want to adapt this for your own camera.

Tested with kstars/EKOS and PHD2, with a BlackFly BFLY-PGE-31S4M camera.

### DIFF
--- a/3rdparty/CMakeLists.txt
+++ b/3rdparty/CMakeLists.txt
@@ -46,6 +46,7 @@ option(WITH_ASICAM "Install ZWO Optics ASI Driver" On)
 option(WITH_DSI "Install Meade DSI Driver" On)
 #option(WITH_QHY "Install QHY Driver" Off)
 option(WITH_GPSD "Install GPSD Driver" On)
+option(WITH_GIGE "Install GiGE machine vision Driver" On)
 
 if (WITH_QHY)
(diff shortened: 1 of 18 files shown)


knro:
Thanks merged, we probably need some documentation for http://indilib.org/devices/ccds.html

If you can send me documentation for the camera along with screenshots..etc that would be great.

knro:
Also, in order to make a debian/ubuntu package, we need the debian directory stuff, please take a look here:

https://github.com/indilib/indi/tree/master/debian

And let me know if you can adapt any of the drivers files to your driver.

PR #94, opened 2016-10-23
Add Optec Filter Wheel driver

mainate:
Change drivers.xml
Change CMakeLists.txt in 3rdParty
Add a new 3rdparty folder  "indi-optec-wheel" with required files


knro:
If you want to add this as a 3rd party driver, then it needs to have its own XML file. But I recommend that this is added directly to libindi (under drivers/filter_wheel) and not as 3rd party since it is a pretty straightforward driver. So just added the files to libindi and then resubmit the PR.

knro:
also please make you pull or rebase from master before you submit the Pull Request to make sure there are no conflicts.

PR #95, opened 2016-10-23
Add optec wheel

mainate:
Change to integrate directly to libindi.
Hope to PR is correct.


knro:
Almost there! Still one problem remains, you submitted 3rdparty/CMakelists.txt which is not supposed to be changed for this pull request. Please do a 'git checkout' for the file and then re-submit you pull request.

PR #96, opened 2016-10-24
Proposal new filter wheel driver Optec IFW

mainate:
add only ifwoptec.* to indi/libindi/drivers/filter_wheel as new driver 
Hope this is correct this time, as ou can see I'm not feel confident with github


knro:
Well, that is missing the change the libindi/drivers.xml but fine I'll add it myself. The one I asked to be removed was 3rdparty/CMakelists.txt change. 

At any rate, great job, git takes a while to get used to!

PR #105, opened 2016-11-08
Temma test

macsleep:
need to set TrackState during goto so EQUATORIAL_EOD_COORD property stays busy
(otherwise no abort with client).

### DIFF
--- a/libindi/drivers/telescope/temmadriver.cpp
+++ b/libindi/drivers/telescope/temmadriver.cpp
@@ -70,6 +70,7 @@ void ISNewBLOB (const char *dev, const char *name, int sizes[], int blobsizes[],
   INDI_UNUSED(names);
   INDI_UNUSED(n);
 }
+
 void ISSnoopDevice (XMLEle *root)
 {
     INDI_UNUSED(root);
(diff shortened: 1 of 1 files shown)


macsleep:
Hi Jasem, they generate empty nameless properties and have no use in the code as far as I can tell.

knro:
I just tried it and I don't see any nameless properties. I can see a "Motion Control" tab under which the guide properties are present. Can you attach a screenshot of what you get?

macsleep:
I can see them on the Macintosh using Peter Polakovics "INDI Control Panel" and the "INDI Device Controller" in PixInsight (see attached images).

screen shot 2016-11-09 at 08 38
screen shot 2016-11-09 at 08 35

knro:
Thanks! I informed Gerry to take a look at this.

knro:
Can you please run the server with -vvv, need to find out which properties are responsible for the "empty" device?

knro:
Thanks, it was GuideRateNP which was defined but not initialized. I disabled it for now. Fixed pushed.

macsleep:
Thanks Jasem :-) .

PR #108, opened 2016-11-13
Add OSX TravisCI configuration

seanhoughton:
With the current push to get kstars running on OSX I thought it would be a good time to get indilib CI running on an OSX Travis instance.


seanhoughton:
I'm looking at the gtest failure on OSX now, should have a fix shortly.

PR #126, opened 2016-12-17
Quantum Filter Wheel merge with mainstream INDI

rkaczorek:


rkaczorek:
It will be resubmitted to core libindi instead to 3rdparty drivers

PR #127, opened 2016-12-17
Prevent infinite loop in indi-asi and also set bandwidth to 40 on connect

dokeeffe:
There are 2 changes in this PR. Not sure if you want to take both but it seems to have helped with the exposure looping on my PC with my camera. Although I'm not 100% convinced that this solves the problem...

1) **Prevent an infinite loop** in the while(true) loop that waits for exposure to complete.
I was able to create an infinite loop by setting EKOS to take continuous 2 sec exposures with the camera using the focus module 'framing' option. Then disconnect the driver and re-connect. It gets stuck in this while loop. I added a debug statement in the loop and could see it being logged continuously.

2) **Set the the usb bandwidth to 40** (auto) on connect. I noticed a strange problem on my PC. Not sure if you also see this.... To replicate: Set debug log on, take a few exposures, now go to the driver's controls tab and try to set the bandwidth to auto or change the value. I get an error. 
```
2016-12-17T20:52:26: ASISetControlValue (BandWidth=40) error (16) 
2016-12-17T20:52:26: ISNewSwitch->SetControlValue 6 40.00 
```
Looks like 16 corresponds to` ASI_ERROR_GENERAL_ERROR,//general error, eg: value is out of valid range`
So I'm not sure its setting the value anywhere while the driver is used... This is why I'm defaulting to 40 on connect. Looks like this call does not fail for some reason. I'm not sure if this cures my looping issue but I seem to be able to reliably use the camera now on my desk.

### DIFF
--- a/3rdparty/indi-asi/asi_ccd.cpp
+++ b/3rdparty/indi-asi/asi_ccd.cpp
@@ -1076,7 +1076,7 @@ void ASICCD::TimerHit()
                       SetTimer(POLLMS);
                       return;*/
 
-                      DEBUGF(INDI::Logger::DBG_DEBUG, "ASIGetExpStatus failed (%d) after retry. Restarting exposure...", errCode);
+                      DEBUGF(INDI::Logger::DBG_DEBUG, "ASIGetExpStatus failed (%d). Restarting exposure...", errCode);
                       InExposure = false;
                       usleep(100000);
                       StartExposure(ExposureRequest);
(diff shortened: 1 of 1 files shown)


knro:
I actually made different changes here, pleas pull and test. It works fine without any looping on x86_64. Of course, it's always warm inside.

dokeeffe:
Thanks, yes I just pulled and built. Seems ok, no looping (in my living
room). Will be traveling back to the observatory later today and connect
everything else up and test again. Hopefully it all still works.

PR #131, opened 2016-12-23
Fixing DMX21/V4L2 corrupted frames

TallFurryMan:
Worked in the V4L2 driver around the problem of corrupted frames on my DMX21.
I tested with a few other webcams, but of course this would need more test samples.
I tried very hard to conform strictly to the V4L2 spec in my changes. There are portions of the code (which I didn't touch) indicating much fiddling with buggy drivers :) I don't think I broke anything there, but well, as long as the problematic device isn't tested...
There's much to do here to clarify the code, perhaps I'll return to this soon.
I have another rework for cropping, but the feature isn't much used probably so that can wait.

### DIFF
--- a/libindi/drivers/video/v4l2driver.cpp
+++ b/libindi/drivers/video/v4l2driver.cpp
@@ -707,8 +707,11 @@ bool V4L2_Driver::StartExposure(float duration)
 {
     if (is_exposing)
     {
-        DEBUG(INDI::Logger::DBG_ERROR, "Can not start new exposure while exposing.");
-        return false;
+        /* Clicking the "Expose" set button while an exposure is running arrives here.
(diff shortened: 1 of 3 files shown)


knro:
This caused compile failure on Raspbian which apparently doesn't have the latest V4L2 headers which support the "flag" bit. See http://indilib.org/forum/general/1779-compile-error-from-the-last-github-src.html#13354

I made a quick fix but apparently that didn't work and I was on my winter vacation so I couldn't check further. Please submit a PR to fix the issue.

TallFurryMan:
Opened [#137](https://github.com/indilib/indi/issues/137).
Issue occurs with older (3y+) kernel headers. I'm investigating how to make the compressed-frame check compatible.

PR #150, opened 2017-02-05
10Micron mount driver. Supports ethernet. Extends lx200generic.

d33psky:
This driver is based off of lx200generic and for now only adds connecting to ethernet.
The 10Micron LX200 command protocol extensions can later be added to this driver.


knro:
Please make driver name in lower case. Also no need to put any comment regarding micron in INDI::Telescope

PR #153, opened 2017-02-08
New release (0.3.0) of NexStar Evo driver

jochym:
This is a new version of nexstarevo driver. There are many internal changes and clean-ups in the code. Here are most important ones:
 * Much better tracking performance
 * Always approach from the direction of tracking
 * Telescope autodetection implemented
 * Telescope simulator added with stellarium monitoring connector
 * Dropped dynamical memory allocations in the driver 
 * Improved stability (still a bit problematic though)
 * Improved and simplified command processing logic
 * Some code clean up and refactoring
 * Constant approach distance on encoders
 * Periodically update target position on long goto
The resulting driver is much more stable and seams to work better.

This is a stepping stone to more work on polishing the interface and implementing niceties like mount monitoring and setup (battery, lights etc) and extensions like serial communication with the mount (AUX protocol) via hand controller or direct serial connection. This version may be regarded as beta quality and is suitable for careful end-user testing.


jochym:
Would you find some time to review and possibly merge this new version @knro  ?

knro:
Done thanks! Perhaps you want to inform users to test it now?

jochym:
Thanks. I will post on the forum. I think it is time o include this driver in the standard build. I have performed long tests and it is working much better. Not 100% stable yet (there is still one elusive allocation bug I cannot trace) but recently it ran for 20h driving the simulator and real hardware without a hitch. I managed to crash it only after five clients were controlling/monitoring the server running on RPi2 at the same time. Inclusion within the standard set would help to expose any bugs still left there.

knro:
Sure, I'll work on releasing a package for it. Can you please create a detailed documentation for this driver here? http://indilib.org/devices/telescopes/celestron.html

You already see the existing driver, you can edit that to copy/paste the structure and then add "New Listing". Please be as detailed as possible. If you have screenshots, send me over and I'll upload them.

jochym:
I will. It will take some time. Thanks for pointing it out. The packaging for debian is already added to /debian - you need to just review it and integrate it into the build process.

PR #157, opened 2017-02-18
Fix Skywatcher Virtuoso mount detection, highspeed query and add slew support in skywatcherAPIMount driver

kecsap:


knro:
AFAIK, this driver was never used before (except for Roger the original developer). I presume you tested this on your mount? Have you tried EQMod?

kecsap:
Hi,

I try to repair this driver. I also tried EQMod, but none of them work correctly with my Alt/Az mount. I also read this discussion:

https://sourceforge.net/p/indi/mailman/indi-devel/thread/527E15F4.5060808%40beardandsandals.co.uk/#msg31618221

And I came to the conclusion that it would be too much effort to modify the EQMod driver to be operational with Alt/Az mounts. Since the skywatcherAPIMount driver seems to me generic and Alt/Az detection was there originally, I thought I would fix it and add also wedging support. If you look into my current commits, the main ideas came from the EQMod driver so far.

kecsap:
Thanks for the merge. Don't you know what kind of mount was used by the original author (Roger)? His email address is dead.

PR #178, opened 2017-03-17
Faster travis builds for separate drivers

jochym:
A faster (approx 5x) path for driver developers to use travis.
If the branch has a name drv_drivername the setup will build 3rdparty/indi-drivername exclusively.
And the build is run using libindi from the PPA.

The developer needs only to name the branch drv_drivername to have it activated.
No other actions are required.

Additionally this limits deb packages building to package branch to save some time.

This works for linux only for now but can be extended to osx if someone will create homebrew tap
with current libindi (like PPA for osx) to use in the process.


jochym:
I think it is ready to be merged @knro .

PR #181, opened 2017-03-18
QHY driver fixes and enhancements

kecsap:


kecsap:
I rechecked with the newest QHY SDK in the repo and new the indi QHY driver does not crash on reconnection, but it still fails. I am not sure if it is only related to my camera or all QHY camera. My proposed changes however make the driver more robust by checking if the camera is still connected to the USB or not before initiating the connection.

knro:
Just tested with QHY5LII-M and it works fine between multiple connects and disconnects, so might be an issue related specifically to your camera.

PR #183, opened 2017-03-20
Migrate Quantum Filter Wheel to updated connections plugin framework

rkaczorek:
For some reason I cannot set default port with serialConnection->setDefaultPort("/dev/ttyACM0") but it does not impact driver's functionality. It works as expected.


knro:
Thanks, I just tested, and the default port was indeed set to /dev/ttyACM0 so it is working as expected.

rkaczorek:
OK, thanks.

21 mar 2017 07:02 "Jasem Mutlaq" <notifications@github.com> napisał(a):

PR #184, opened 2017-03-22
Set a certain USB speed for QHY 5P-II C camera by default

kecsap:
This is the best approach so far for QHY 5P-II C. I look into the driver if I can fix and make it better.

### DIFF
--- a/3rdparty/indi-qhy/qhy_ccd.cpp
+++ b/3rdparty/indi-qhy/qhy_ccd.cpp
@@ -590,9 +590,13 @@ bool QHYCCD::Connect()
         if(ret == QHYCCD_SUCCESS)
         {
             HasUSBSpeed = true;
-          // Force the speed to 0 on initialization of QHY5PII-C to avoid stuck transfer
-          if (isQHY5PIIC())
-            SetQHYCCDParam(camhandle, CONTROL_SPEED, 0);
+
+            // Force a certain speed on initialization of QHY5PII-C:
+            // CONTROL_SPEED = 2 - Fastest, but the camera gets stuck with long exposure times.
+            // CONTROL_SPEED = 1 - This is safe with the current driver.
+            // CONTROL_SPEED = 0 - This is safe, but slower than 1.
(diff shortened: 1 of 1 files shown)


knro:
So driver still not working?

kecsap:
The official driver has the following issues with QHY 5P-II C:
- For longer exposures (>9 sec), the initial exposure time gets insanely slow. E.g 2 minutes for 10 seconds.
- If the SPEED is set to 2, the initial exposure may get complete stuck and the driver must be killed.

I am in contact with QHY and help to fix these problems.

knro:
So maybe we should hold off this PR until the remaining issues are solved? Maybe you wouldn't even need it at that point?

kecsap:
Hi,

unfortunately, it is maybe by hw design. I don't know if it can be fixed. If you check the official page:

http://www.qhyccd.com/QHY5II.html

...it says on the bottom specs sheet for QHY5P-II:

"Note(*)In some computer can not get 2592*1444 at high speed readout, in these comptuer it works only in 2048*1944 at high speed readout".

I suppose this is exactly the case here, max. resolution with CONTROL_SPEED = 2 and a workaround is needed for this model. After I played a few days, I found out what is in the comments:

- CONTROL_SPEED = 2 - Fastest, but the camera gets stuck with long exposure times.
- CONTROL_SPEED = 1 - This is safe with the current driver.
- CONTROL_SPEED = 0 - This is safe, but slower than 1.

Would it be better to restrict the CONTROL_SPEED parameter range to [0, 1] for this particular model instead of this commit?

knro:
In the commit you already set it to 1, any additional benefits from 0?

kecsap:
If I understood your question correctly, the benefit of this commit:
- Clear description of why 1 is chosen as a default value now.

or 

If you want, I can drop this commit and modify other parts of the driver to disable the possible CONTROL_SPEED = 2 case for this particular QHY model on the UI side.

knro:
No it is fine now as is. Merged.

PR #186, opened 2017-03-23
Make the image download in QHY driver async

kecsap:
The grabImage() function was called in TimerHit() function which may not return shortly and I experienced stability issues with the communication from a Raspberry Pi indiserver to my Virtuoso mount. This change makes the image download async and it does not block the TimerHit() function.

Please review this PR carefully because it is a change to a basic function of this driver. I used this change so far in my setup and I did not experience any side-effects. Don't merge immediately.

### DIFF
--- a/3rdparty/indi-qhy/qhy_ccd.cpp
+++ b/3rdparty/indi-qhy/qhy_ccd.cpp
@@ -1064,7 +1064,11 @@ int QHYCCD::grabImage()
   if (ExposureRequest > POLLMS * 5)
       DEBUG(INDI::Logger::DBG_SESSION, "Download complete.");
 
+  PrimaryCCD.setExposureLeft(0);
+
   ExposureComplete(&PrimaryCCD);
+
+  InExposure = false;
 
   return 0;
 }
@@ -1074,6 +1078,10 @@ void QHYCCD::TimerHit()
   if (isConnected() == false)
     return;
 
+   if (!InExposure && asyncCapture.get())
+   {
+     asyncCapture.reset();
+   }
    if (InExposure)
    {
        long timeleft = calcTimeLeft();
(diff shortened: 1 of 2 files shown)


knro:
I can see how this can be quite problematic. What if the next exposure starts before grabImage() is completed? That could lead to nasty effects.

kecsap:
How the next exposure can be started before the previous frame capture is finished and downloaded?

These conditions are explicitly set only before grabImage returns.

PrimaryCCD.setExposureLeft(0);
ExposureComplete(&PrimaryCCD);
InExposure = false;

And only one grabFrame() can be called once because its call is guarded in a unique pointer (asyncCapture) in TimerHit(). This variable is cleaned if InExposure already set to false. If you want I can add additional check also with asyncCapture->wait_for() if the grabImage() call is completely finished.

asyncCapture.get()

I am also not very familiar on the overall architecture of the Indi <-> Ekos <-> KStars system. Is not there any side-effect if a driver does not return in a reasonable amount of time from TimerHit()? Can it impact the communication of other drivers in any way?

knro:
The INDI driver is self contained, it shouldn't depend on how external clients like KStars behave. You have 20+ INDI clients, not just KStars/Ekos. A simple script can be issuing continuous fast captures each 0.1 seconds for example.

How does a blocking QHY affect mount communication? Is INDI server being blocked somehow? Each driver lives in its own separate process space.

kecsap:
Is the drivers TimerHit() callback is connected in any way to the indiserver? I saw that indiserver is connected with pipe to the drivers, but I did not look into the low-level communication. My perception was that the async call improved the communication with the mount, but I might be wrong.

kecsap:
I brought back this PR if I see value. Now I drop it.

PR #190, opened 2017-03-26
Minor Fixes

mp77:


mp77:
Just a fix for the default baudrate to get the driver working again. Will work on the variable types later.

knro:
All appear OK, but this:

```
if (isConnected() == false)
{
        SetTimer(POLLMS);
        return;
  }
```

SetTimer is again called by INDI when a connection is re-established. If there a reason this is needed here?

mp77:
I get alot of error messages if I do not check for connection and manually disconnect the focuser. Looks like the TimerHit() function is called even when the connection is terminated. Whats the right method to stop the timer on disconnection?

knro:
The timer is a single-shot timer anyway, if no one calls SetTimer(POLLMS) then it is not invoked. I just tested in other focuser and TimerHit() is not invoked when the device is disconnected.

PR #195, opened 2017-03-29
Fix BLOB size passed when block is compressed

TallFurryMan:
BLOB length `bloblen` must be coherent with the actual length of the block, while `size` should report the actual size of the raw data.
Make Astrometry driver use field `bloblen` to avoid a crash reading invalid memory.


TallFurryMan:
Crash confirmed fixed with a few use cases.

Unfortunately, this results in the solver detecting a problem with the FITS structure when the frame is uncompressed. Thus, regression or misuse in there.

TallFurryMan:
Fixed by b82021

PR #197, opened 2017-04-02
fix some typo and correct home button state at reset by changing type…

mainate:
fix some typo and correct home button state at reset by changing type to ISR_1OFMANY. Remove from second focuser connection options to avoid end user confusion
Remove local portFD as already exist in original indi library


knro:
This code fails to compile. You need to include serial and tcp connections here. i.e.
`
#include "connectionplugins/connectionserial.h"
#include "connectionplugins/connectiontcp.h"
`

PR #198, opened 2017-04-04
Add alternate aperture/focal length configuration options to skywatch…

kecsap:
…erAPIMount driver

### DIFF
--- a/libindi/drivers/telescope/skywatcherAPIMount.cpp
+++ b/libindi/drivers/telescope/skywatcherAPIMount.cpp
@@ -21,6 +21,7 @@
 #include <memory>
 #include <cmath>
 #include <limits>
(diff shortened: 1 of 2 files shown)


knro:
This problem shouldn't be resolved at the driver level but at the client level. This way each driver will implement its own settings for this.

kecsap:
I am not sure how much need is there for other people apart from that one recent forum topic and I don't know if this is the best approach. I just hacked in, it works, but I will test it how practical it is. For a general implementation, the UI and other things should be much more well-thought. At least the Virtuoso mounts were advertised with standalone DSLR and telescope usages. I don't know how common is this feature for other mounts, for me it is very easy to swap during an imaging session.

PR #202, opened 2017-04-08
Adding explicit RTS Line control

Vignesh1230:
Added code to explicitly raise and lower RTS line when the serial port is open.


knro:
I wonder how would this affect existing users who use the serial shutter normally under Linux?

knro:
Ok I guess one way to find out is to test it with other users.

PR #210, opened 2017-04-25
failed execvp() fixed in scripting gateways

polakovic:


TallFurryMan:
Note that the debug log output might corrupt another log sent by the driver issuing the fork. Perhaps you should use a return code and keep the debug log in the parent driver.

PR #213, opened 2017-04-28
Remerge from indilib master

grbsystems:


grbsystems:
Ignore - got this the wring way round again!

Issue #212, opened 2017-04-28
indi_gpsd: reports GPS fix obtained repeatedly without actual fix


xsnrg:
In the logfile, the lines:

` 2017-04-28T16:25:44: GPS fix obtained. `

are repeated every second even though the fix mode in the GPSD tab of the INDI control panel properly says NO FIX.

the actual gpsd process does not have location data at this point, and is attempting to lock on to more satellites with the gps unit.

Additionally, once this is scrolling in the log file, the gpsd service can be completely stopped, and  this message continues to be added to the log.  Pressing the Update button finally gives the proper GPSD read error, and the messages stop.

jochym:
Indeed. It seems the actual behavior of gpsd is a bit different from the docs. The status indicates we have a fix but there is no actual fix data yet. I have added a check for actual fix data (https://github.com/indilib/indi/pull/214) and moved time read before checkin the fix. 
It seems to be working now.

PR #214, opened 2017-05-01
gpsd driver fix lock detection

jochym:
Fix gps fix detection (https://github.com/indilib/indi/issues/212) 
Switch cfitsio to licfitsio on osx travis build.


xsnrg:
That was quick.  Thank you both.

PR #215, opened 2017-05-01
Feature  v4l2ccd exposure left

TallFurryMan:
This pull request should allow users of USB CCDs to better manage exposures.

- Added exposure left report to client.
- Fixed various start/stop/restart/abort situations which would lock the driver logic up.
- Implemented CCD_CAN_ABORT logic over those fixes.
- A few log changes to make driver behavior clearer.

Basically, it allows the end-user to abort that 20-minute exposure launched by mistake.
It also makes starting and stopping streaming and exposure more robust.


knro:
Amazing work, thanks!!!

PR #216, opened 2017-05-01
DomeSync correction

fcasarramona:
Correction to OpticalCenter and OpticalVector and simplification of
calculus.


knro:
Thanks a lot! I'll ask the user who reported the issue to test this patch!

Issue #228, opened 2017-05-17
indi-gpsd has improper DST offset


xsnrg:
Discussion: http://www.indilib.org/forum/general/2004-help-with-gpsd.html?start=12#16787

timedatectl returns the proper system settings and DST offset.  indi-gpsd performs a check against the current timezone and changes the offset when it does not need to.  This might be needed on other systems?  On Linux it is handled and the math should not be performed.

    // adjust offset for DST
    if (local->tm_isdst)
        offset -= 1;

credit to:jpaana

knro:
@jochym What do you think?

jochym:
He is right. We are correcting for DST twice. (!) 
I'll prepare PR in few moments.

xsnrg:
Thank you @knro and @jochym

xsnrg:
Confirmed fixed with #229

PR #229, opened 2017-05-17
GPSD: Fix double DST correction

jochym:


jochym:
Should fix problem with incorrect DST offsets (#228)

rkaczorek:
I believe that the change proposed is not complete. All the following lines should be removed from the code:
    // get utc_offset
    double offset = local->tm_gmtoff / 3600;

    // adjust offset for DST
    if (local->tm_isdst)
        offset -= 1;

    // convert offset to string
    sprintf(ts,"%0.0f", offset);
    IUSaveText(&TimeT[1], ts);
 
The offset is calculated and saved in the lines:
188    snprintf(ts, sizeof(ts), "%4.2f", (local->tm_gmtoff/3600.0));
189    IUSaveText(&TimeT[1], ts);

jochym:
Why the save lines? I do not get it.

rkaczorek:
The save is in line 189: IUSaveText(&TimeT[1], ts);
New you're double saving the offset by leaving the code below you have proposed to remove.

jochym:
You are right. Removed

Issue #242, opened 2017-05-25
indi_v4l2_ccd crash


jakugab:
Hello!
I have a Raspberry Pi 2 running a freshly installed Ubuntu MATE 16.04 for the RPi (firmware version: 4.1.19 v7+). I attached the dedicated RPi NoIR V2 camera board, which works fine with the camera app (I can take photos with raspistill command from terminal), and streaming program, like Cheese (using bcm2835_v4l2 module). I installed the latest indi-full and kstars-bleeding from the ppa successfully. In Ekos I made a profile with only a CCD (I chose V4L2 CCD from CCD tab). When I start INDI, the ccd is trying to connect, but fails, Kstars says: "Kstars detected, that indi_v4l2_ ccd driver crashed, please check server log in the Device Manager ". Here is the Kstars-log of the latest crash:

 2017-05-17T09:30:53.652 - DEBG - Welcome to KStars 2.7.7
2017-05-17T09:30:53.677 - DEBG - "Setting clock:  UTC: Sze máj. 17 07:30:53 2017 GMT  JD: 2Â 457Â 890,81"
2017-05-17T09:30:54.135 - DEBG - Opened the DSO Database. Ready!
2017-05-17T09:30:54.637 - DEBG - Opened the User DB. Ready.
2017-05-17T09:30:59.893 - DEBG - Processing  "unnamedstars.dat" , HTMesh Level 3
2017-05-17T09:30:59.894 - DEBG -   Sky Mesh Size:  512
2017-05-17T09:31:00.074 - DEBG - Loaded DSO catalog file:  "unnamedstars.dat"
2017-05-17T09:31:00.081 - DEBG - Processing  "deepstars.dat" , HTMesh Level 3
2017-05-17T09:31:00.082 - DEBG -   Sky Mesh Size:  512
2017-05-17T09:31:00.082 - DEBG - Loaded DSO catalog file:  "deepstars.dat"
2017-05-17T09:31:02.366 - WARN - "Star HD61421 not found."
2017-05-17T09:31:02.382 - WARN - "Star HD10700 not found."
2017-05-17T09:31:02.409 - WARN - "Star HD20794 not found."
2017-05-17T09:31:02.435 - WARN - "Star HD2151 not found."
2017-05-17T09:31:02.468 - WARN - "Star HD190248 not found."
2017-05-17T09:31:02.505 - WARN - "Star HD142860 not found."
2017-05-17T09:31:02.515 - WARN - "Star HD1581 not found."
[...]

PR #245, opened 2017-05-29
V4L2 integer menus

geehalel:
I've added V4L2 integer menus (specifically for the rpi camera).
Also try to stop capturing/recording when device is disconnected.
Finally check if a V4L2 control is mutable before setting it (it can be grabbed, inactive,....)

### DIFF
--- a/libindi/libs/webcam/v4l2_base.cpp
+++ b/libindi/libs/webcam/v4l2_base.cpp
@@ -689,8 +689,11 @@ int V4L2_Base::stop_capturing(char * errmsg)
             // long time ago. I recently tried taking this hack off, and it worked fine!
 
             type = V4L2_BUF_TYPE_VIDEO_CAPTURE;
-            IERmCallback(selectCallBackID);
-            selectCallBackID = -1;
+	    if (selectCallBackID != -1)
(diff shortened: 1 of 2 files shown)


TallFurryMan:
Are you able to reproduce #242 with your setup? There's a debug log causing a segfault somewhere in check_device.

geehalel:
I don't have a raspberry camera and the ones I use were working with a
quiet recent version of libindi.
I would suggest to modify Indi::Logger to test for null string as I think
the raspberry driver may not fill every fields (cap.driver, cap.bus_info or
input names for instance).

TallFurryMan:
I think a null string will be properly handled as "(null)". Might be a non-init string in an out-of-bounds structure record instead. I'll slice that big function into smaller ones and the issue should pop up while refactoring. Thanks :)

geehalel:
Yes, you're right. Reading the v4l2 doc:
"Storing strings in fixed sized arrays is bad practice but unavoidable
here. Drivers and applications should take precautions to never read or
write beyond the end of the array and to make sure the strings are properly
NUL-terminated."
suggests to test every string returned from v4l2 drivers...

PR #247, opened 2017-06-02
Replace NULL with nullptr

kecsap:


TallFurryMan:
Isn't this "restricting" the build to C++11 compilers ?

knro:
We've been on C++11 for a while now, see CMakelists.txt

PR #250, opened 2017-06-04
Fixed warning requiring "*" format specifier to be of type "int".

TallFurryMan:
This should remove warnings issued during the x64 build.


TallFurryMan:
Should be good to go with commit e4d5875. Sorry for the mess.

PR #253, opened 2017-06-08
Use Clang Format for code indentation

kecsap:


knro:
As it is, it affects many good indentations we have for #defines and enums among others, so can't apply as is.

PR #262, opened 2017-06-18
Migrate the recent improvements to the 3rdparty plugins

kecsap:


jochym:
Why we have clang in linux build? This is not a standard compiler for linux. Is this required for some reason?

kecsap:
@jochym: Because clang catches different bugs (warnings) compared to gcc and it is a pretty drop-in replacement for gcc on Intel platforms.

PR #272, opened 2017-06-26
Silence a warning for gcc-7

kecsap:


TallFurryMan:
Now we're getting bleeding edge. I note that some drivers are set to Off. Do you plan to raise issues for them

TallFurryMan:
and let authors fix?

kecsap:
@TallFurryMan Which drivers do you mean?

TallFurryMan:
There are two options in libindi and three in 3rdparty. I can manage the build for the MGen driver with gcc 4, 5, 6 and 7 (though I don't know if my system will be able to slot all these simultaneously), but do we consider others deprecated?

kecsap:
@TallFurryMan: I think Jasem mentioned that the disabled drivers have no maintainer and we can't test it. On the other hand, if we enable any of these unstable drivers, Travis CI already build all gcc versions+clang 4.0 with warnings turned to errors now.

TallFurryMan:
Understood. How can I trigger a Travis build? Should I open a PR with for instance the MGen enabled, and fix repeatedly afterwards? Or is there another way?

kecsap:
@TallFurryMan: Exactly. Otherwise, you have to manage your local builds of gcc, clang yourself.

TallFurryMan:
About Travis, I observe the distribution is Ubuntu trusty. This causes a dependency for the MGen driver (libftdi1-dev) not to be readily available (libftdi1 is though). Do I import the source and build the dev resources or do we move to a later distribution for the worker? (isn't it 16.04 which is officially supported by indi?)

knro:
Travis doesn't support 16.04 thus far unfortunately. It's odd but that's their policy.

TallFurryMan:
I'll try to add the ppa for libftdi1 to the Travis config. Fairly new to this tool, sorry :)

PR #277, opened 2017-06-30
Unity Build support

kecsap:


jochym:
linux/joystick.h is not going to be supported on osx... We probably need to switch off this driver for OSX buillds?

kecsap:
@jochym The new version fixed. It was a mistake in my changes.

PR #279, opened 2017-07-03
Build libftdi for MGen driver in Travis CI

kecsap:


TallFurryMan:
I would have picked the package from the ppa, but this is pretty neat! Thanks!

PR #302, opened 2017-07-17
remove libnova dependency for indiclient and fix some MSVC compile warnings/errors

agalasso:


agalasso:
I meant to do this a two separate pull requests.

The first piece is 16134bd which removes the libnova dependency from libindiclient and fixes some MSVC compile errors and warnings.

The second piece, 6d17c27 +  5f467cc is independent, fixing the Windows runtime errors caused by the incompatible setlocale calls.

PR #313, opened 2017-07-24
Caught one more location where CONNECTION property is being used

mikefulb:
This file doesn't seem to be built (?)  but I though I'd catch the one remaining place the string "CONNECTION" is being used.


knro:
It's not built by default.

PR #318, opened 2017-07-26
Fixes and adds for indidetector

iliaplatone:
I added a new tutorial for indidetector and fixed a bug on blob generated fits for this interface.


knro:
I don't see simulator detector file in there?

iliaplatone:
finished.

knro:
does it make sense for detectors to have shutters? coolers?

iliaplatone:
These capabilities are for photon detectors, which are 1-px very fast CCDs essentially

PR #324, opened 2017-08-02
Fix DEC custom rate scaling and set OK state when tracking is enabled

mikefulb:
DEC rate needs to be scaled to sidereal.
Tracking state switch never set to OK when tracking mode turned on.


knro:
TrackState is already set is line 1263 and for declination changes, the values are in arcsec/s and are NOT related to the sidereal rate. did you measure the actual tracking rate? (Final DEC - Initial DEC)/Time should give you the tracking rate so you can test if it is actually moving in the specified rate.

knro:
I actually take that back regarding declination, I just re-read the documentation. I updated the rate now to 
double APDERate = deRate / TRACKRATE_SIDEREAL;

PR #325, opened 2017-08-02
Fix compiler warnings

kecsap:


knro:
Opss, can you resubmit the PR against master?

PR #329, opened 2017-08-08
Fix macOS compilation in Travis CI

kecsap:


knro:
hmmm same error occurs again.

PR #331, opened 2017-08-11
Fix for missing FILE_PATH property in gphoto ccd

pludov:
The property FILE_PATH failled to appear in gphoto when Upload was set to Local or Both.
As a result, a notification was received by clients for a unknown property.


knro:
Thanks for the find! I implemented another solution to fix this issue. It should be working now!

PR #338, opened 2017-08-18
Fixed many issues that was causing the NStep driver to not function.

eric-vickery:


knro:
In the future, please use DEBUG(....) macro instead of IDMessage(...)

eric-vickery:
Looking at the code again most of the IDMessage(...) calls were in the original version of the driver.

knro:
Hi Eric, thanks for the PR! I fixed most of these issues already and pushed it. Let me know if you have any issues with the newer revision. Your contribution is greatly appreciated!!

PR #341, opened 2017-08-24
Fix geographic coordinate comparison

pchev:
At the moment if you load an AlignData.xml file, you can never save it again because of the rounding error in the floating point comparison.
Worst the file is cleared because it is opened for write before this test.

This patch fix the problem by comparing the coordinates to 1E-4 (10 meters) and opening the file after the test.


TallFurryMan:
Good spot!

PR #344, opened 2017-08-26
[3rdparty] added librtk

iliaplatone:
I added a small library. librtk gives high precision GPS positioning.
librtk can give sub-meter positioning precision.


iliaplatone:
There is a licensing problem: the older librtk versions were released with GPL3.0 licenses.

PR #348, opened 2017-08-29
Disable EQ mounts in Skywatcher API Mount

kecsap:


knro:
I guess it's time for a change in the driver name?

kecsap:
That is also a subsequent commit in the queue. I wait for this merge and then I will push more to review.

PR #350, opened 2017-08-30
Rename Skywatcher API Mount driver to Skywatcher Alt-Az Mount

kecsap:


knro:
Do not forget drivers.xml

PR #356, opened 2017-08-31
Prefer the saved telescope config over the configuration

kecsap:


knro:
What problem is this trying to solve exactly?

kecsap:
After I implemented the scope configs, I noticed that sometimes some config is lost. I figured it out that the configuration loading/saving and the config loading/saving interferes. Basically, if "Config #1" is already saved, there is no too much sense to try to load/save anything from the configuration because "Config #1" will be loaded by default when the INDI panel is created.

PR #361, opened 2017-09-03
Updates to dome sync math

grozzie2:
I found another error in the dome sync math, as well, base dome is not getting side of pier updates from the telescope.  Both corrected in this PR


knro:
@fcasarramona FYI

PR #364, opened 2017-09-04
Eqmod PHD2 issue

ken-self:
PHD2 was segfaulting with a null pointer to a property being removed  with EQMOD driver. Specfically noted the SIMULATOR properties
Changes the Eqmod simulator class to add a property isEnabled to hold the enablement state of the simulator so that in updateProperties, properties are only removed if they had been previously defined
Made equivalent changes to eqmod, align and scope-limits to hold connection status.
With these changes phd2 no longer receives removal of properties that have not been defined.
Existing tests on property pointers were ineffective as they were always set in the initProperty methods


knro:
I made a simpler change to the code so that simulator properties are not defined. Didn't test it on a real mount yet.

PR #366, opened 2017-09-09
Check current connection state before updating properties

ken-self:
This addreses continued segfaults in PHD2 caused by Eqmod driver - currently resolved with a patch in Phd2. On receipt of a DISCONNECT command, the state of the driver is checked and if not connected then updateProperties is not called so that undefined properties are not removed.


knro:
Thanks, I made changes to the same effect. Can you please recompile and check now?

ken-self:
Looks good. I checked with eqmod and no issues. Also checked with my other devices and no adverse effects on those either.

PR #382, opened 2017-10-01
[indi] changed symlinks' paths, [inovaplx] corrected ST4 timing

iliaplatone:
Fixed ST-4 timing and symlinks fixes:
When building with a different installation path, symlinks to some libraries and drivers point to absolute paths, I changed them to point to local relative paths, since link targets are usually in the same directory. This way it's possible to move the whole INDI library after a custom installation (ex. custom ARM jailed environment) without losing link references


knro:
Can you please break this into 2 PRs? The symlinks needs to be reviewed separately.

PR #383, opened 2017-10-01
[indi] changed symlinks' target to local relative paths

iliaplatone:


knro:
Looks like you submitted all in this PR, not just symlinks.

PR #385, opened 2017-10-02
fixed paths in generated symlinks

iliaplatone:


knro:
Shouldn't 3rdparty/indi-mi be included as well?

PR #387, opened 2017-10-04
New driver ioptronHC8406 base on lx200gotonova

nachoplus:
indi_ieq45_8406_telescope: fix compling isues.Very basic driver only slew and 

ioptronHC8406: New driver base on lx200gotonova. Ioptron 8406 Hand Control is very buggy. This driver implement almost all the funcionality that works and hide the commands than don't. 

WORKS:
Slews
Guiding commands and move by cursor commands
Change tracking speed (but the mount has to switch to track manually first)
Read/wrtite time/date lat/lon
Cursor move speeds (centering, guiding)

DO NOT WORK:
Start/stop tracking (there is not command for that)
Slew rate (:RS#, :RS0#..)


knro:
Thanks Nacho for the update! Few comments:
1. Fix Travis errors.
2. Remove IOptron documentation, these do not go in the repo.
3. Remove .RHistory
4. libindi/drivers/telescope/ieqprodriver.cpp has an unnecessary change, revert that.
5. What's difference between ioptron HandController 8406 and OLD ioptron HandController 8406?

nachoplus:
Hi Jasem! 

OK with all your suggestions. I'll fix them and make a new pull request this nigth..

The old indi_ieq45_8406_telescope driver was a very base one, only slew and track rate were soported. The new one include more funcionality.  I fixed and added just for emotional/historical reasons. In my machine compile without problems but  I saw that the Travis error are related to this driver so **I'll remove**.

Anyway I think that is a good thing to have  tecnical information such the command set documents or testing results on the official repository.. May be not in this place but it could be very handy and secure (i.e. I wasn`t be able to find the "HC8406 command set" pdf on internet, I was to search on my computer/email to recover..).

nachoplus:
Jasem , Sorry for the mistakes and thanks for the final tunning!

PR #389, opened 2017-10-07
checkLX200Format() to fix ioptronHC8406 driver RA/DEC precision issue

nachoplus:
Bad RA/DEC set/get precision . Calling checkLX200Format() function fixed it. 

One more issue have to be solved:
.- SetTrackMode() is not invoqued until a first slew is done. I am not sure but I think is related with IsNewNumber() on the INDI::Telescope base class and could affect other drivers also...


knro:
You do not need to define SetTrackEnabled() as it is not called by INDI for your driver. Also, not sure what you mean by SetTrackMode.. it is only made when a user makes a change.

nachoplus:
Two different issues:

1.- When you just start the driver and the change the track mode switch using the GUI  not :RT0 (or :RT1...) is sent to the mount. Then you do a "on set" command even without filling the RAC/DEC fields and start to work. Apparently SetTrackMode() is not called until you do the first slew/sync.

2.- The other issue is not a cmd in the  HC8406  command set to start/stop tracking. The CMDs :RT0,RT1 and so on, only affect tracking speed when tracking is on but don't start tracking itself. So you have to start tracking manualy using the HC. This is very bad news because it prevent using this mount remotly. I include setTrackEnabled() function as a places holder locking for a workarround.. actually do nothing..

Anyway, talking about track/slew/sync switch, I understand that the diferences between   track/slew options are that track means "slew+start tracking at selected speed" and slew means "slew only and stop tracking" so I understand that SetTrackEnabled(on/off) has to be call after ":MS#". hasn't it?

knro:
Ok, I think there is a bit of confusion here.

1. Yes, it is only called once the GOTO is finished and tracking starts. You can define another property like EQMod does "TRACK_DEFAULT" that selects the default tracking mode _before_ the mount begins the GOTO and then engages tracking. In your situation, it might be easier to check in Goto(double,double) if the TrackModeSP has changed from previous value, and if it did, then make the actual call. The solution is up to you.

2. Since you didn't define TELESCOPE_CAN_CONTROL_TRACK as part of the mount driver, SetTrackEnabled(bool) is _never_ called for your driver, and therefore you don't need to override, it is never called.

nachoplus:
Hi Jasem,

Thanks for your help. I fixed my 1) issue with the information you provided me.

Relating second 2) issue, I asked to ioptron people about the way to start/stop tracking by software. I think that this is crucial for a remote operate telescope. So I need a little more time to finish this work.

In the mean time **is up tu you to merge this pull request** that only fix the precision issue. 

Nacho

PR #395, opened 2017-10-18
USB_Dewpoint driver

jpaana:
This pull request adds driver for USB_Dewpoint (http://www.usb-foc.us/) dew controller to change settings and monitor temperatures. It's originally very loosely based on the USB_Focus_V3 driver, but all functionality is obviously different. Communication protocol is fairly similar as well and was quite easy to reverse engineer from the Windows driver debug log, which nicely prints all the commands and responses to the device. No dependencies to any external libraries or anything, just basic serial port communication.

### DIFF
--- a/libindi/CMakeLists.txt
+++ b/libindi/CMakeLists.txt
@@ -939,15 +939,6 @@ IF (UNIX AND NOT APPLE)
     INSTALL(FILES ${CMAKE_CURRENT_SOURCE_DIR}/drivers/auxiliary/99-flipflat.rules DESTINATION ${UDEVRULES_INSTALL_DIR})
 ENDIF ()
 
-########### USB_Dewpoint Driver ###############
-
-SET(usb_dewpoint_SRC
-    ${CMAKE_CURRENT_SOURCE_DIR}/drivers/auxiliary/usb_dewpoint.cpp)
-
-add_executable(indi_usbdewpoint ${usb_dewpoint_SRC})
-target_link_libraries(indi_usbdewpoint indidriver)
-install(TARGETS indi_usbdewpoint RUNTIME DESTINATION bin)
-
 ########### Sky Quality Meter ###############
 
 SET(sqm_SRC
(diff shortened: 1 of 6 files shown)


knro:
Thank you for your contribution! Since this driver is weather related, it needs to be based on the INDI::Weather class. Just yesterday I added MBox Meteostation driver (https://github.com/indilib/indi/blob/master/libindi/drivers/weather/mbox.cpp) you can base your driver on it and modify it as necessary. I hope it's not too much work!

jpaana:
Sure no problem, I wasn't sure if it was weather related enough to warrant using Weather class as it's a fairly simple dew heater controller with just a few temperature and humidity sensors so I made it just aux device instead. I'll update the pull request when ready.

knro:
Oh stop! a dew heater! Nevermind then it does not qualify as a weather device! It's just auxiliary. I should've checked the product first sorry.

jpaana:
Oops, did the change already, it did actually clean up things a bit, but I can revert it as well, yes/no?

knro:
Okay, since it has temperature and humidity sensors, It can be technically used as a weather-reporting device. I guess we can add it twice, once in weather and once in auxiliary (with two different labels perhaps).

jpaana:
Would a suitable compromise be that I move it back to auxiliary directory, but keep the usage of INDI::Weather for the weather related parts? There are temperature sensors in almost all focusers too but I wouldn't look for them in the weather section :)

knro:
I think that would create more confusion. Just revert it back to a generic auxiliary device and call it a day.

jpaana:
Ok, done :)

knro:
Can we get some documentation and screenshots for this driver so I can an add an entry to INDI devices page?

jpaana:
Something like this http://s2.org/~jpaana/usb_dewpoint/?

knro:
That's prefect! One question though: regarding the udev rule, is it specific to USB_dewpoint or that's some FTDI vendor id?

jpaana:
Not FTDI but Primax, but yeah, probably similar generic USB-to-serial-chip:
Bus 001 Device 004: ID 0461:9e41 Primax Electronics, Ltd 

I also have USB_FocusV3 focuser, which also uses different Primax chip:
Bus 001 Device 009: ID 0461:0033 Primax Electronics, Ltd 

so I don't know if the rule is specific enough, but haven't had conflicts myself, with FTDI I've had to use serial numbers to distinguish between devices...

knro:
Thanks, I added the documentation here: http://indilib.org/devices/auxiliary/usb-dew-point.html

jpaana:
Didn't look carefully enough at lsusb output before but it's actually possible to make device specific udev-rules for these:

SUBSYSTEM=="tty", ATTRS{product}=="USB_Dewpoint", SYMLINK+="usb_dewpoint", ENV{ID_MM_DEVICE_IGNORE}="1"

Similar rule for USB_Focus would be:
SUBSYSTEM=="tty", ATTRS{product}=="USB_Focus", SYMLINK+="focuser", ENV{ID_MM_DEVICE_IGNORE}="1"

PR #405, opened 2017-10-26
Driver for Gemini Snapcap 

jpaana:
This pull request adds driver for Gemini SnapCap (http://www.geminitelescope.com/snapcap-motorized-tube-ota-cap/) motorized OTA cap and flat light. It's mostly the same driver as for Flip Flat but with some serial protocol and behavior differences. The controller is Arduino based and uses the same USB ids as any other Arduino Uno so specific udev rules are unfortunately not possible, but otherwise something like this works:
`SUBSYSTEM=="tty", ATTRS{idVendor}=="2a03", ATTRS{idProduct}=="0043", SYMLINK+="snapcap", ENV{ID_MM_DEVICE_IGNORE}="1"`

Also the current firmware version doesn't provide very good way of identifying the device and whether it has a light or not so I just read the firmware version and assume the device is there if the response looks valid and that it has a light.

### DIFF
--- a/libindi/CMakeLists.txt
+++ b/libindi/CMakeLists.txt
@@ -939,6 +939,15 @@ IF (UNIX AND NOT APPLE)
     INSTALL(FILES ${CMAKE_CURRENT_SOURCE_DIR}/drivers/auxiliary/99-flipflat.rules DESTINATION ${UDEVRULES_INSTALL_DIR})
 ENDIF ()
 
+########### SnapCap Driver ###############
+
+SET(snapcap_SRC
+    ${CMAKE_CURRENT_SOURCE_DIR}/drivers/auxiliary/snapcap.cpp)
+
+add_executable(indi_snapcap ${snapcap_SRC})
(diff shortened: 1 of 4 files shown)


knro:
Thanks! This is great I thought FlipFlat was the only device of its kind available commercially until this. Please fix the travis error and it would be great if you can provide excellent documentation for it like you did before!

jpaana:
Yes, it's good to have more options. I wrote some documentation at http://s2.org/~jpaana/snapcap/ and removed the unused variable so Travis should pass again.

jpaana:
Forgot to mention the controller has built-in bluetooth serial port that can also be used instead of USB cable, added that to the documentation.

knro:
Thanks, I modified the documentation a bit here to clarify more details: http://indilib.org/devices/auxiliary/gemini-snapcap.html

Can you please inform Gemini Telescopes about this driver so that they mention INDI support on their page with a link to the above documentation? Thanks!

jpaana:
Thanks, done :)

PR #408, opened 2017-11-03
Losmandy Gemini fixes

eric-vickery:
Fixed some issues that were causing the Gemini driver not to work correctly. Added new code to deal better with ethernet connections.


knro:
Great work Eric! Lots of changes to that may affect other drivers as well. For set/get calendar date, Meade uses '/' as date separators and not '-', so you might want to make a function for Gemini, or pass the separator to the function.

Is the UDP communication in indicom Gemini specific?

eric-vickery:
I will change the date separators to a parameter but when I looked at the spec for the LX200 from Meade's site it said that the date would have a '-' as the separator.

The general UDP support is generic and could easily be used by other drivers. It is just if you call tty_set_gemini_udp_format(int enabled) that it will do some things specific to the Gemini. I added that to the common indicom code because the changes were minimal and I figured it was better not to duplicate the code into the Gemini driver for a very few changes.

knro:
I think the date separators could be treated as wild-card maybe something like this:
sscanf(str, "%d%*[^0-9]%d%*[^0-9]%d", &a, &b, &c);

knro:
What's the status? I also updated LX200 driver to fix an issue with the time format as well so there is conflict now.

eric-vickery:
Sorry I have not had the time to look at it. Hopefully I will have some time this weekend. I will merge master into my branch and fix any conflicts.

eric-vickery:
I merged the latest master with my branch and resolved the merge conflict. I tested it out last now and it worked great. Everything should be good to go now.

knro:
There are still Travis error, can you fix these please?

libindi/libs/indicom.c
```diff
@@ -347,6 +354,22 @@ int tty_write(int fd, const char *buf, int nbytes, int *nbytes_written)
...
+    {
+        buffer[0] = ++sequenceNumber;
+        buffer[1] = 0;
+        strncpy((char *)&buffer[2], buf, nbytes);
+        // Add on the 8 bytes for the header and 1 byte for the null terminator
+        nbytes += 9;
+    }
+    else
+    {
+        strncpy(byteBuffer, buf, 255);
```
jpaana: This line broke my dome driver as strncpy ends at the first zero byte and it's not using a text based protocol but purely binary. Should be memcpy or the whole copy in itself is totally unnecessary in the non-gemini case.

libindi/libs/indicom.c
```diff
@@ -435,26 +433,47 @@ int tty_read(int fd, char *buf, int nbytes, int timeout, int *nbytes_read)
...
+            // Not the right reply just do the read again.
+            return tty_read(fd, buf, nbytes, timeout, nbytes_read);
+        }
+
+        *nbytes_read -= 8;
+        strncpy(buf, readBuffer+8, *nbytes_read);
+    }
+    else
+    {
+        strncpy(buf, readBuffer, *nbytes_read);
```
jpaana: Strncpy here is a no-no too, same thing as above with writes

knro:
This is long closed, but recently a user reported problems connecting to the mount and then he said he used port 11110. Should this be default in the driver?

libindi/libs/indicom.c
knro: Sorry I just saw this, was this addressed?

eric-vickery: Yes those issues were fixed in another PR. I don't remember which one though.

PR #415, opened 2017-11-11
Add Slew Accuracy to Sky Sensor 2000 PC

kbahey:
This change adds a configurable slew accuracy parameter to the SkySensor 2000 PC hand controller. 

The previous default difference between current and target RA/DEC was 0.01, and that caused slews to not end. Work around was to Abort the slew manually. 

Now there is no need for this workaround. One can set the slew accuracy to a different value than the default 3 arc minutes. On my Vixen GPDX with the Sky Sensor 2000 PC ROM version 2.05, a value of 4 arc minutes seems to work well indoors.

### DIFF
--- a/libindi/drivers/telescope/lx200ss2000pc.cpp
+++ b/libindi/drivers/telescope/lx200ss2000pc.cpp
@@ -24,12 +24,12 @@
 #include "lx200driver.h"
 
 #include <cmath>
-#include <cstring>
+#include <string.h>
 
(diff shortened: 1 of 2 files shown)


knro:
Thanks Khalid! Can you make corrections against master since this PR conflicts with the existing tree?

kbahey:
I think they are resolved now. 

Please check and let me know.

knro:
Please fix the Travis errors.

kbahey:
All good now.

PR #417, opened 2017-11-12
Correct setting of latitude and longitude on Sky Sensor 2000 PC

kbahey:
The setting of latitude and longitude on Sky Sensor 2000 PC is different from the generic LX200. First, a space is required between the Sg and St command and its arguments. Also, seconds can be used unlike the generic LX200.


knro:
Great work! merged.

PR #418, opened 2017-11-13
Better logging and engage tracking after sync correctly (Synscan Mount)

kecsap:


knro:
I would suggest you get rid of all IDLog and use DEBUG and DEBUGF respectively.

knro:
Great! There is no need to use if (isDebug()) at all since the DEBUGX macros take care of that.

DEBUG(INDI::Logger::DBG_SESSION, "Synscan Mount not responding");

This would probably be DBG_WARNING or DBG_ERROR

PR #422, opened 2017-11-18
Fix tty_write and tty_read for non-text protocols

jpaana:
Commit bbbead91e73ba04e65b219c84b02b9913bbf7a17 introduced a special case for Gemini UDP mode in tty_write and tty_read and broke them for any writes or reads that had zero bytes in them due to use of strncpy which terminates at the first zero byte. My dome driver happens to be one so I bisected what broke it in the last few days. The change also introduced a potential buffer overflow in case someone writes or reads more than about 256 bytes at a time. In this change it's only a problem for the Gemini special case, but should preferably be cleaned up properly.

### DIFF
--- a/libindi/libs/indicom.c
+++ b/libindi/libs/indicom.c
@@ -362,21 +362,17 @@ int tty_write(int fd, const char *buf, int nbytes, int *nbytes_written)
 #ifdef _WIN32
     return TTY_ERRNO;
 #else
-    int buffer[66]={0};
-    char *byteBuffer = (char *)buffer;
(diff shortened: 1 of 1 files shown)


TallFurryMan:
@knro I discussed with tech sales at Synopsis recently. They offer free static code analysis for open source projects, with no obligation in return (they say this is an opportunity to train their analyzer). I could set that up for indi if you want.

knro:
Function should migrate to tty_read_section_n which is the safe alternative.

PR #427, opened 2017-11-24
Lx200ap setguiderate

mikefulb:
This modification allows changing the guide rate used by the lx200ap driver to 0.25x, 0.50x, or 1.0x.  I've tested changing the guide rate and doing a calibration run with PHD2 and the guide rate appears to change as expected.


knro:
Does this fix the issue reported for GTOCP2?

mikefulb:
I am not sure yet - I need to talk to Howard @ AP and will not be able 
to until next week.  I still don't understand the exact problem.

What this change does do is allow the user to set the guide rate which 
might allow a work around.

Michael

PR #432, opened 2017-11-29
Lx200ap fix detect time format

mikefulb:
The default time format (12/24 hr) check in LX200Generic does not work on AP mounts.  Following the 10Micron driver we replace getBasicData() with one custom to lx200ap.


knro:
Ok there are Travis errors. There is no need to override getBasicData as it already does not define sites. What I did with timeFormat is that it is initialized to -1, and if the value is -1 then LX200Generic will inquire for its value from the controller, but if it is already set, then it will not. I think this should fix this issue. All code pushed. Please test.

PR #436, opened 2017-11-30
Fix the OSX build issue on travis

jochym:
The reason was just cmake already present in the Xcode 8.3 image


jochym:
I expect this to still fail since current master is failing but it could be checked that the source tree builds fine on OSX (see: https://travis-ci.org/jochym/indi/builds/309501859). The next build was cancelled to save resources on travis.

PR #440, opened 2017-12-01
Start moving linux builds to CircleCI

jochym:
@knro : Please, do not marge this until I remove WIP: from the title. 
The basic setup for CCI is done I have moved gcc5 build over and removed redundant default build (it was building with gcc5 for the second time). It is much faster. The rest may be moved later. We can even have separate docker images for different environments (eg. debian/ubuntu/centos/fedora). 
Let us check this for a few days.


.travis.yml
```diff
@@ -6,12 +6,13 @@ sudo: required
...
- - osx
-
 matrix:
   include:
+    - os: osx
+
+# Default config (on trusty gcc4)
+    - os: linux
+
     - os: linux
```
jochym: Moved to make it explicit. When we accept CircleCI builds we should drop at least some builds here to lower the load on travis.

jochym:
We should drop most of travis builds when we accept CCI config. I have restored full set to make the transition slowly.

jochym:
@knro : I think this is ready to merge into master. The only changes outside of CircleCI config are corrections of build-dependencies of several packages to make them build on trusty. They are verified in my PPA and non-intrusive. I will wait for you to review this but it is fairly non-intrusive patch. We can remove redundant builds from travis later.

knro:
Thanks! This is major improvement to the CI build of INDI! Let's give it a go live.

PR #441, opened 2017-12-03
Added 3 new dependancies for 3rd party build

hdevillepoix:
I had to install those in order to build all 3rd party packages on Debian GNU/Linux 9 (stretch)


knro:
Good catch, thanks!

PR #450, opened 2017-12-06
Adapt Indi Alignment Subsystem for EQ mounts and use it in indi-eqmod

geehalel:
Main updates to IAS are:
- use an ha telescope reference frame in place of the altaz frame used with altaz mounts
- compute a rotation matrix when having only one sync point
- use a nearest algorithm when outside the current convex hull or when having only 2 sync points
Besides, the original alignment code in indi-eqmod has been deactivated.


TallFurryMan:
Do you have some test prerequisites and scenarii with which we can test these change thoroughly?

geehalel:
No, it should behave as before, when using the original alignment code.

jochym:
Please review the errors - I believe they are warnings treated as errors (e.g. unused variables etc).

jochym:
Wow. That is strange. It seems to me to be some missing header (str). I have to go right now. But what have changed in this file? (strange that clang4/5 compails it fine ...)

libindi/libs/indibase/alignment/MathPlugin.cpp
```diff
@@ -18,5 +18,59 @@ bool MathPlugin::Initialise(InMemoryDatabase *pInMemoryDatabase)
     return true;
 }
 
+std::string MathPlugin::GetInternalDataRepresentation(std::string PluginDisplayName)
```
jochym: I'm not a huge C++ expert but std::string is defined in <string> but Common.h only includes <cstring>.
Is this the reason? Strange that gcc and clang4/5 on linux did not complain...

geehalel: Neither do I... On my linux gcc `<cstring>` includes `<string>` so that should do the job. I try with including `<string>` and `<cstdio>` too for `snprintf`.

PR #454, opened 2017-12-09
Apogee driver & streaming removed from mac build

polakovic:


knro:
Why is streaming removed? The whole purpose behind my recent work is to get streaming working on the Mac platform!!

polakovic:
Hm, maybe I missed something, but it produced a lot of errors.

knro:
Please check on your side because Travis already runs CI on MacOS and it's been building fine and also tested OK with several users on MacOS.

polakovic:
What is the configuration on Travis? If you use ports or homebrew layer it may work, but it is not native mac build.

knro:
@jochym your thoughts?

knro:
Please post the error you had on native Mac.

polakovic:
OK, pls. drop this pull request for now and I'll try do it next time with streaming or will post error messages. But anyway, what's the configuration in Travis?

polakovic:
The immediate reason was just lot of missing new dependencies, it can be probably solved.

BTW, your implementation of streaming will be totally incompatible with INDIGO streaming. Why did you do it this complicated way?

jochym:
Travis build uses homebrew. What you mean it is not a native build? 
It uses additional libraries naturally. Otherwise you would need to include in your packages a ton of external code not provided by apple. You are perfectly welcomed to modify osx build config on travis to make it better. I am not an expert on osx, I was thinking that homebrew is just a way to distribute the additional libraries. If we should go another way, please propose something.

jochym:
@polakovic : why not cooperate with Jasem on making it simpler and better on all platforms?
The travis setup is fairy simple:
```
brew update
brew tap 'homebrew/homebrew-science'
brew tap 'caskroom/drivers'
brew tap 'indilib/indi'
brew upgrade cmake
brew install dcraw fakeroot gsl libftdi libgphoto2 libusb gpsd libraw libcfitsio libnova librtlsdr libfftw3
brew cask install sbig-universal-driver
```
The build is almost identical on all platforms (see travis-ci directory). If you need any help with this just ask. I'll do my best.

polakovic:
Pawel, native build is sandboxed, sealed and signed macOS package without any external dependencies. Installed, upgraded or removed by drag & drop and executed with double click. This is how couple of hundreds Mac users runs INDI now. Homebrew on Mac is something like Wine on Linux.

jochym:
Ok. Now I understand, except I think wine is a bit more complicated then homebrew but that is beside the point. We should avoid forking the development even if there are multiple ways the soft is run on one platform. This will not be good for project as well as for your enterprise as well (IMHO). I do not know how majority of OSX indi users run the software, but I know that fragmentation is never good. So I guess my suggestion is to make the core INDI in INDIGO as similar to the upstream as possible and the travis build as dependence-free as possible. I will be happy to assist with the CI part, and I am fairly sure Jasem will cooperate with his part of the code. I believe this will be beneficial for all parties. Ideas welcomed.

polakovic:
Yes, you're right! But first of all we have to agree on the basic premise that INDI driver should be as simple as possible to be usable and portable. They are overloaded by many useless features and dependencies and many of them are not portable at all (e.g. what is plate solving in driver good for? what are math plugins in alignment module good for? all this is completely wrong and out of control!).

Another issue is INDIGO compatibility. Even if we have totally different approach, we invested a lot of effort to be initially compatible with INDI. I really don't think that it is a good idea to use different implementation for features in INDI if they are already implemented and proven to be functional in INDIGO (e.g. streaming).

knro:
The streaming stuff has been in INDI for almost a decade, there is nothing new in them (data wise) except more encoders and recorders. Recent work was made to make them work on Mac & BSD without additional libraries. Sure you can _optionally_ libtheora, but it is not required to get this functionality working. INDI streaming also works with PHD2 for a long time among other clients that depends on it.

polakovic:
It was and still is linux only feature and was available only for V4L2 driver for almost decade, it is not even mentioned in the list of standard properties. How many users really depends on it? :) Let's stop it, we're just wasting time...

jochym:
Yes we will be wasting time if we not move the discussion to technical points. Beside "mine is better" discussion  which is indeed a waste of time. What is your practical, technical proposition @polakovic ? To replace streaming in INDI by the INDIGO implementation? I believe that would be impossible unless you donate the code and someone would need to port it to the linux platform. I fully support the Keep-It-Simple attitude. It makes for better code and I don't think that fragmentation is a good idea. So let us work on technical level how we can make INDI better for the sake of both projects.

polakovic:
Technical point is, that there is just one additional property for streaming in INDIGO with two items (exposure time and frame count). It works exactly the same way as CCD_EXPOSURE with the exception, that multiple frames can be requested (for frame count >0) or frames are sent until CCD_ABORT_EXPOSURE is set (for frame count = -1). Property is optional and present only if hardware does support some kind of "video" mode (ASI, QHY and IIDC in our case).

There is no need to define mode, frame size, image format, etc for streaming. All settings are shared between single frame capture and streaming request. E.g. if it makes a sense to use JPEG for streaming (BTW, it actually doesn't, but we have it as well), it makes a sense to use it for single frame mode as well. From our experience much more useful is to add "raw" format (8 or 16 bit unsigned ints in little-endian bit order with minimal header) instead of FITS, it may be used directly by rendering engines on client machines or ffmpeg without endian change and rescaling on both sides.

I don't think there is too much code that can be shared between INDI and INDIGO, but this approach is so simple, that it can be added to base classes in a few hours. Nevertheless you can look at indigo_process_image(...) in

https://github.com/indigo-astronomy/indigo/blob/master/indigo_libs/indigo_ccd_driver.c

to see how it is implemented. Note that we don't use cfitsio and in the same buffer is used from USB transfer on driver side to rendering on multiple client sides to save time and resources. I don't know if "mine is better", but it is 10x faster in average if driver and client are in the same process :)

Property and item names are not important, INDIGO can translate them on per protocol version basis.

PR #456, opened 2017-12-11
BaseClient::isServerConnected

asiniscalchi:


knro:
I think you need to add the same function to baseclientqt

PR #466, opened 2017-12-19
Compilation fixes

kecsap:


knro:
probably do not need the one for QHY as I will push an update shortly for SDK v1.0.8

knro:
any of this still required? If yes, please update the PR.

kecsap:
Yes, since the problem was not fixed really, just a temporary. The next QHY SDK update will contain the same problematic header what is currently locally patched.

jochym:
I am not too enthusiastic about switching off warnings instead of fixing the problem... Is this the only way?

kecsap:
@jochym We have to be relaxed for 3rd party components because we don't want to maintain local patching. (And it is not a crucial warning.)

PR #467, opened 2017-12-20
Fix qhy compile errors

jochym:
- External functions cannot be static
- current libqhy does not support gcc < 4.9


knro:
There will be an update QHYSDK so we'll hold off this for now.

jochym:
Ok. But it would be better to have buildable master at all times...

PR #471, opened 2017-12-21
Fixes for OSX build

deisenlord:
This is a set of build fixes for the Apogee camera drivers in libapogee for OSX.  

Summary
Under recent versions of OSX the C++ Technical Report 1 includes have been deprecated, libapogee was using <tr1/memory>.

One of the networking files requires the inclusion of <sys/select.h> under OSX.  

Finally the shared library version number 3.0.3234 is incompatible with OSX's version numbering scheme which is a 64bit tuple of up to 8 components that range from 0-255.    In CMakeLists.txt I conditionally changed this to 3.0.3.2.3.4 for OSX.  This is ugly but works for any version number, an alternative would be 3.0.32.34.   I'm ambivalent.

I tested the build of libapogee under OSX, RedHat, Debian and Ubuntu.

David


knro:
So indi-apogee can build cleanly against this under MacOS now? Just to confirm.

jochym:
The apogee build is switched off on OSX in CI - this should be changed for this PR, otherwise we do not know if it builds in travis environment.

deisenlord:
@jochym this is my first pull, I'm not familiar with the CI system, if you point me in the right direction and I have the permissions to do so I can re-enable OSX.

deisenlord:
@knro, let me pull the latest indi-apogee tonight and rebuild under OSX just to double check but I believe so.

jochym:
@deisenlord essentially yes. Additionally check 3rdparty/CMakeLists.txt for switched off apogee driver.
It is easier for you to do that (it will be automatically added to the PR). Ping me again if you have trouble with this and I'll try to append appropriate changes to your PR.

jochym:
@deisenlord : I think you need to change setting for apogee in line 88 of 3rdparty/CMakeLists.txt

deisenlord:
@jochym, thank you!  I missed that.

jochym:
No problem. Remember to push to your branch to update PR. Without CMakeLists changes only library will  get build on OSX.

jochym:
Please add changes to CMakeLists to your commit @deisenlord - the last build tested only libapogee without the driver.

deisenlord:
@jochym  Yes just did that.  I was trying to test build all of indi under linux and had to come to grips with the two pass nature, libraries then drivers before I got indi-apogee to correctly build.    
Thanks!

jochym:
You can use docker images (https://hub.docker.com/u/jochym/indi-xxxx) for local tests see readme in docker subdir. They are pre-configured for building indi.

deisenlord:
@knro I built both libapogee and indi-apogee tonight under OSX with my changes, no problem.  Deployed these versions and fired up EKOS and my camera and took a couple of frames, all appears well.  
David

knro:
Thanks a lot for this work!! I hope your changes make it upstream as well!

deisenlord:
Your very welcome, you've built and amazing system in KStars/Ekos/Indi.



BTW, this morning I submitted our libapogee changes to Apogee/Andor/TheRiver in the hope they will incorporate them upstream.



David





-----Original message-----

rlancaste:
I tested this a few minutes ago.  It tries to install files to  /etc/Apogee/camera, which is a protected directory.  You would need to run as sudo to do that.  All my stuff for home-brew, INDI, KStars, astrometry etc is in the /usr/local folder.  So shouldn't that be placed in /usr/local/etc/Apogee/camera since my prefix is /usr/local?

rlancaste:
in cmakelists, it says set(CONF_DIR "/etc/Apogee/camera" CACHE STRING "Configuration directory")

Should that not have "${CMAKE_INSTALL_PREFIX} in it or something?

rlancaste:
I would prefer to not run the KStars or INDI build commands using sudo and would like to keep them in the /usr/local directory structure.

deisenlord:
All your comments make sense to me @rlancaste.  I noticed this last night but didn't think too hard about it.  We should definitely change it to use the install prefix.  I'm pretty ignorant of CMAKE at the current time but I can take a crack at this and report back.  

BTW do you know if there has been any progress on build_kstars.sh, I will have problems with the translations.

deisenlord:
Hum, there is a disconnect, don't understand how its been working previously, maybe it hasn't?   The code in ApgHelpers.cpp is hard coded to use /usr/local/etc as the configuration directory.  So copying files to /etc is broken anyway. 

I will move forward with changing the CMakeLists.txt

rlancaste:
I made a bunch of changes to it about a week ago. It should be fixed now.  The one thing is that you might need to uninstall kf5-stuff before you run it.  I decided to eliminate Jamie's hacks for using the QT version installed to ~/QT from the QT installer.  We were originally doing that because the brew QT took about 9 hours to install, but that isn't true anymore.  Recently there have been some home-brew issues with phonon and also it seems with the WIT dialog in kstars when using the non-brew QT.

rlancaste:
I haven't changed the emerge/craft version to use the brew qt yet, but might try that soon.  It might fix the WIT dialog in KStars. I'm not sure if that was fixed because it was the brew QT or because it was QT 5.10 instead of 5.9.  Phonon is fine in craft using either brew or system qt.

rlancaste:
but in answer to your previous question about how it was working before, I have not tried to build libapogee before. It was disabled in the OS X build via cmakelists.

rlancaste:
I also can only test building it, I don't have an apogee camera.

deisenlord:
I'm guessing in the heat of building/testing it I ended up copying the configs into /usr/local to get it working and just plain forgot.  I will check tonight.   

Thanks for the updates on the OSX build, I will give things a new shot this weekend.

rlancaste:
ok

rlancaste:
do you think the apogee build will be fixed today?  if so, I can test the build and see if any files/drivers need to be copied into the app for the dmg version.  Also if any environment variables need to be set.  This was true for both GPhoto and for QHY.

deisenlord:
I hope so, right now I can't push my changes to my branch, getting some strange errors.





-----Original message-----

deisenlord:
I've pushed my new branch up to github, how can I give you access?

rlancaste:
You don't want to do a pull request for Jasem?

deisenlord:
I can do that, just wonder if you would give it a try first?

rlancaste:
ok, what is the cloning link or GitHub page?

deisenlord:
https://github.com/deisenlord/indi.git

rlancaste:
It says that is even with master currently, did you make the changes to it?

deisenlord:
The changes are on branch fix-osx-config-location

rlancaste:
ok, I have to go get my oil changed.  I can test that when I get back

deisenlord:
Roger 👍

rlancaste:
Ok, the Apogee camera text files no longer cause a build error, now they are copied to /usr/local/etc/Apogee/camera

There was another error though:  file cannot create directory: /lib/udev/rules.d

udev rules are linux only.  So I assume this should not be attempted/installed.

rlancaste:
The issue is on line 10
set(UDEVRULES_INSTALL_DIR "/lib/udev/rules.d" CACHE STRING "Base directory for udev rules")
or perhaps line 63
install(FILES 99-apogee.rules DESTINATION ${UDEVRULES_INSTALL_DIR})

I assume either the first line should install to ${CMAKE_INSTALL_PREFIX}/lib/udev/rules.d or the second line should be for linux only and not run on OS X systems.  Both should fix it I think

rlancaste:
Anyway, I can still work with this now since you fixed the configuration file location.  I assume that the config files will need to be copied into the KStars.app and that we will need to set an environment variable in kstars so it knows where to find these files.

rlancaste:
I will see what I can do with that part if you can fix the cmakelists.txt

deisenlord:
Sounds good, I'll take a look.

deisenlord:
@rlancaste I made the changes to skip the udev deployment for OSX and tested it.  Created a pull request.  Thanks so much for helping me out on this, lots to learn.

knro:
It should be fixed now. Please git pull and test.

Issue #473, opened 2017-12-23
Sync fails for Celestron Nexstar driver


nesos:
Hi everyone,

when calling a sync command, the Celestron Nexstar driver will only transmit the first half of the sync command. 
For example:  If the telescope shall sync to the zero coordinates for RA and DEC it will only transmit

`s00000000,`

instead of

`s00000000,00000000`

The DEC coordinate is missing, causing a timeout on the hand controller.
I am using the latest version from the mutlaqja ppa.

Best regards, Ben

nesos:
I think I found the problem and fixed it in the merge request #474

knro:
Fix is merged.

PR #469, opened 2017-12-20
Fix a bug in Celestron protocol and reduce code verbosity

juanmb:
- I fixed a bug that happened when parsing responses from celestron mounts: `tty_read_section` doesn't work with some non-ASCII packets.
- I added a method for checking the firmware version that removes code duplication.
- New logging macros reduce the code verbosity an avoid many long lines.

PR #475, opened 2017-12-23
Fixed length error for precise sync commands

nesos:
This fix will take the set the command length according to the precision mode,
This solves the problem stated in Issue #473


knro:
There is another PR pending, so I informed the author of this. Check https://github.com/indilib/indi/pull/469

knro:
This will be address in a different PR, thanks!

PR #476, opened 2017-12-23
Fix osx config location

deisenlord:
@knro This contains fixes for the install of libapogee on OSX.  @rlancaste  found that the install was hardcoded to put camera config files in /etc which of course a privileged location. I subsequently found the code is hardcoded to look in /usr/local/etc for camera config files, so a disconnect.

@rlandcaste tested my changes and found an additional problem where the OSX build was trying to install a udev file which is unix specific.  I conditionally changed the CMakeLists.txt file skip this file for OSX.

Sorry for not finding all this on the first go, lots to learn.
David


knro:
Thanks for the PR. I just looked into it in detail and there was even a problem with Linux since apgHelpder.cpp used a hardcoded value which is wrong. I pushed a fix for this for both Linux & MacOS. I hope upstream switches to CMake and users INDI's work in libapogee for less headache in the future.

deisenlord:
Thanks for digging in, your fixes are so much better.
David

PR #480, opened 2017-12-26
Prevent Portage errors while installing on Gentoo

Ctibor:
This fixes sandbox ACCESS_VIOLATION errors while installing libqhy, libsbig and libinovasdk on Gentoo. Symlinks to .so are created in working directory instead of live system.


Ctibor:
I made a mistake in patch. Will create new PR.

PR #481, opened 2017-12-26
Prevent Portage errors while installing on Gentoo

Ctibor:
This fixes sandbox ACCESS_VIOLATION errors while installing libqhy, libsbig and libinovasdk on Gentoo. Symlinks to .so are created in working directory instead of live system.


jochym:
@Ctibor please check "maintainers can edit" on the right hand side of this PR page - that way we can make corrections to the PR.

Ctibor:
@jochym I see it checked in other words this option is enabled.

jochym:
Thanks.

knro:
You need to rebase and submit a new PR. Also how is this going to work exactly? Make .so in build directory then copy them over?

PR #482, opened 2017-12-27
WIP: Add pull requests to the filter (fix CCI build for PR)

jochym:


jochym:
It seems the mistake was not entirely mine. CCI, apparently, sometimes puts branch name into the variable an sometimes "pull/number" (I think maybe when the user has no CI account?). 
Maybe we should get rid of the branch filter altogether @knro ?

knro:
I'm not really familiar with the CircleCI (or Travis CI for that matter). As long as it builds from master after each push it should be fine. Not sure what the branch filter for?

jochym:
Right now only some branches are build and if the branch name is
drv_drivername only this driver is built - o make it much faster for driver
developers.
Maybe all branches should be built without any limits?

knro:
Hmmm, so if a commit pushed against libindi, then only libindi is compiled but if pushed against indi-eqmod then only that is compiled in the CI?

jochym:
Something like that but witch branch names as a selector. Look at my repo -
I have branches drv_gpsd and drv_nexstarevo when I push to them they are
only built in minimal setup - if I make PR out of them full build is
performed.

PR #485, opened 2017-12-31
Make unit tests work in debian/ubuntu-based containers under CircleCI

juanmb:
This commit should fix (in debian/ubuntu containers) the issue which cause unit tests not being run on CircleCI.
There are also some cosmetic changes, like removing trailing spaces and empty lines at the end of some files.


jochym:
Thanks @juanmb !

Issue #421, opened 2017-11-17
libapogee on gentoo fails with multilib-strict-check


Ctibor:
Hi,
building libapogee on Gentoo 64bit fails with:
```
Files matching a file type that is not allowed:
   usr/lib/libapogee.so.3.0.3234
 * ERROR: sci-libs/libapogee-1.4.1::brancik-overlay failed:
 *   multilib-strict check failed!
 * 
 * Call stack:
 *   misc-functions.sh, line 666:  Called install_qa_check
 *   misc-functions.sh, line 217:  Called source 'install_symlink_html_docs'
 *   80multilib-strict, line  46:  Called multilib_strict_check
 *   80multilib-strict, line  42:  Called die
 * The specific snippet of code:
 *              [[ ${abort} == yes ]] && die "multilib-strict check failed!"
 * 
```
Gentoo multilib wants libraries installed to /usr/lib64 and /lib64, not /lib and /usr/lib
It is a matter of changing CMakeLists.txt like this:
`install(TARGETS apogee LIBRARY DESTINATION lib)`
to
`install(TARGETS apogee LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR})`

PR #487, opened 2018-01-02
Fix build scripts for cleanly installing on Gentoo via portage

Ctibor:
Similar to fix in indilib/indi#421 this fixes error when installing libdspau on Gentoo. This GNU/Linux distribution recently introduced checks when running multilib profiles, that prevents installing 64bit libraries other than to /lib64 and /usr/lib64 directory.

${CMAKE_INSTALL_LIBDIR} is the right variable to determine correct location.

More information for the curious people is here: https://devmanual.gentoo.org/archs/amd64/index.html

Modifications to CMakeLists.txt in libsbig, libqhy and libinovasdk are also needed for smooth installation experience on Gentoo. When portage (Gentoo package manager) compiles package, it uses sandboxed environment. Inside this sandbox a working directory is created and cmake build is ran. The make install phase results into a image of a project, that is then transfered to the system. Writing files outside this sandbox (i.e. to live filesystem) while install is in progress results in error similar to this:

Install the project...
-- Install configuration: "Gentoo"
-- Installing: /var/tmp/portage/sci-libs/libsbig-9999/image/usr/lib64/libsbigudrv.so.2.1.1
-- Installing: /var/tmp/portage/sci-libs/libsbig-9999/image/usr/include/libsbig/sbigudrv.h
-- Installing: /var/tmp/portage/sci-libs/libsbig-9999/image/lib/firmware/sbigucam.hex
-- Installing: /var/tmp/portage/sci-libs/libsbig-9999/image/lib/firmware/sbiglcam.hex
-- Installing: /var/tmp/portage/sci-libs/libsbig-9999/image/lib/firmware/sbigfcam.hex
-- Installing: /var/tmp/portage/sci-libs/libsbig-9999/image/lib/firmware/sbigpcam.hex
-- Installing: /var/tmp/portage/sci-libs/libsbig-9999/image/lib/firmware/stfga.bin
-- Installing: /var/tmp/portage/sci-libs/libsbig-9999/image/lib/udev/rules.d/51-sbig-debian.rules
running /usr/bin/cmake -E create_symlink libsbigudrv.so.2.1.1 /usr/lib64/libsbigudrv.so.2  2>&1
 * ACCESS DENIED:  unlink:       /usr/lib64/libsbigudrv.so.2
failed to create symbolic link '/usr/lib64/libsbigudrv.so.2' because existing path cannot be removed: Permission denied
running /usr/bin/cmake -E create_symlink libsbigudrv.so.2 /usr/lib64/libsbigudrv.so  2>&1
 * ACCESS DENIED:  unlink:       /usr/lib64/libsbigudrv.so
failed to create symbolic link '/usr/lib64/libsbigudrv.so' because existing path cannot be removed: Permission denied

Proposed patch fixes those errors.


jochym:
I like the idea of expanding the spectrum of systems where INDI can be cleanly installed. The keyword is _expanding_. I have looked over the changes and they look fine, but we need to make sure this does not break any working builds (e.g. ARM or OSX). It seems to be fine but I think at least one (better two) more person with good knowledge of INDI build system should sign-off on this (@knro what do you think?)

knro:
In the last PR, it failed on Debian. I will have to check this again but got busy. I will definitely check it tomorrow and see if it raises issues for other platforms.

jochym:
The CCI builds seem to be ok - I inspected few of them. Particularly in the modified places. I did not check debian packaging builds - and they may be affected. Maybe we should add such build to the CCI? We are already above 4cpu in second stage of the workflow so this will not change the time to finish. So only packaging builds really need to be checked at this point.

Ctibor:
Just looking on the debian test build and it failed because apt-get could not install g++-5.

jochym:
You are talking about packaging builds? Can you point me to some log @Ctibor ?

Ctibor:
I meant this https://travis-ci.org/indilib/indi/jobs/324227438

jochym:
That may be due to the updated travis images. I will investigate this, as it should not happen.

jochym:
It seems they (travis) removed gcc-5 from their trusty image. It is actually ok. since gcc-5 is not supported on trusty. If we need gcc-5 build (xenial default compiler) we should move this to CircleCI, switch to deprecated image or disable and wait for xenial image on travis. @knro ? What will you like? I can implement any of this easily.

knro:
So I just tested this and it does NOT create symlinks. For example, in libsbig, on sudo make install, I get this:
Install the project...
```
-- Install configuration: ""
-- Installing: /usr/lib/x86_64-linux-gnu/libsbig.so.2.1.2
-- Up-to-date: /usr/include/libsbig/sbigudrv.h
-- Up-to-date: /lib/firmware/sbigucam.hex
-- Up-to-date: /lib/firmware/sbiglcam.hex
-- Up-to-date: /lib/firmware/sbigfcam.hex
-- Up-to-date: /lib/firmware/sbigpcam.hex
-- Up-to-date: /lib/firmware/stfga.bin
-- Up-to-date: /lib/udev/rules.d/51-sbig-debian.rules

```

knro:
Ok just tested it again and the symlinks are installed. Please update the SBIG library name since it is now called libsbig

Ctibor:
It is done.

PR #488, opened 2018-01-03
Make tests run on all platforms

jochym:
Also update dockerfiles to contain properly installed googletest framework.


jochym:
@knro - sorry for making a little mess with your commit ee8d14f ! I had to temporarily revert it to have a clean working tree. When I realised that I can do it without this revert, just in different branch it was too late. This is only noise in commit log - nothing changed in your code.
This PR should make the tests execute in all containers on CCI and on OSX build on travis (we can switch it on/off by simple env variable). The failure on clang4/clang5 is due to the problem in DSI driver (the same as on OSX) - you can check my builds before reapplying of ee8d14f (https://github.com/jochym/indi/commits/master). The PR is not intrusive it only changes dockerfiles and travis config.

knro:
Thanks, we have a patch for DSI ready so will apply it shortly.

PR #494, opened 2018-01-10
 update to 1.6

polakovic:


knro:
libapogee and indi-apogee are already compiled and working for MacOS and tested by at least one user with access to the camera. This change breaks this. Why is this required?

polakovic:
This build is tested with users as well.

My changes are two months old, you refused them because of streaming. I
spent whole day by merging it and really will not do it again :(

knro:
Can you ask your users to test the current driver please?

polakovic:
Jasem, if you mean driver in master repo, it will not work. I had a short
discussion with deisenlord to this issue couple of days ago, we made more
less the same mac specific changes, but he did it for homebrew and I did it
also for native build. It needs to be tweaked even much more to be properly
sandboxed, there are still some absolute paths pointing to
/usr/local/something, but I'll not do it until my repo will be in sync with
master again. Peter

knro:
Ok but your changes now _break_ the existing driver again. For example, consider this:

 // N.B. Prefer environment variable over compiled value if it exists		
 -    if (getenv("APOGEE_ETC_DIR") != NULL)		
 -        path = help::FixPath( getenv("APOGEE_ETC_DIR"));		
 -    else		
 -        path = help::FixPath( sysconfdir );

APOGEE_ETC_DIR was added there to make the path configurable at run time, but it was removed in your commit. Also, it seems your PR reintroduces <tr1/memory> when it was removed from the code.

What brew-specific things this library depends on exactly? libusb?

polakovic:
Yes, libusb is not in libusb-1.0.

<tr1/memory> is apparently needed on some linux versions and accepted on
all of them, why it was removed? It was correct.

APOGEE_ETC_DIR is a very bad idea, there is common variable INDIPREFIX
introduced years ago (see basedevice.cpp) and set by INDI Server or similar
containers on driver start. As it is this driver can't be used in native
mac application.

BTW, homebrew should be avoided at all, it compromises macOS security and
makes system potentially unstable. Forcing non-IT users to install it is
irresponsible!

knro:
<tr1> is causing problems for C++11 or later support, so it was removed and it's been building fine in the CI since then.

Regarding APOGEE_ETC_DIR, I'm trying to submit all these changes upstream so we cannot use any INDI references. So whatever changes you'd like to merge are fine as long as they do not break the existing library.

rlancaste:
So the goal of adding the APOGEE_ETC_DIR variable in this driver and similar environment variables in other drivers was to allow the firmware path to be configurable at runtime by a mac application so that all of the drivers, firmware, and related files can be wrapped up into the app bundle so it can be moved around.  It was mainly in support of KStars but could be used by any mac application.  INDIPrefix is the path to the indiserver directory which is inside the KStars app bundle in /Contents/MacOS/indi.  We did not place the firmware and plugins in the same directory though.

rlancaste:
The first set of drivers we had to do this for was GPhoto.  It has two environment variables that need to be set so that you can find its supporting files, IOLIBS and CAMLIBS.  These folders and environment variables were not designed by us, because of course GPhoto is a separate project, so we just used their existing environment variables.  We placed these folders into the KStars app bundle in /Contents/Plugins/libgphoto2_port and /Contents/Plugins/libgphoto2 respectively.

polakovic:
>
> So the goal of adding the APOGEE_ETC_DIR variable in this driver and
> similar environment variables in other drivers was to allow the firmware
> path to be configurable at runtime by a mac application so that all of the
> drivers, firmware, and related files can be wrapped up into the app bundle
> so it can be moved around.


Unfortunately it doesn't work this way. INDI Server or similar container
generally doesn't know there is libapogee or gphoto inside and that it
should set APOGEE_ETC_DIR to something.

 INDIPrefix is the path to the indiserver directory which is inside the
> KStars app bundle in /Contents/MacOS/indi


Exactly. Thus $INDIPREFIX/../Resources/ is the correct root for anything
you need to bundle with INDI.

rlancaste:
When we tried to get QHY, DSI, and Apogee to work in KStars on OSX, we followed the same scheme that worked so well when we did GPhoto.  QHY files ended up in /Contents/Plugins/qhy/firmware  Apogee files were put in /Contents/Plugins/Apogee/camera and finally DSI ended up in /Contents/Plugins/firmware/   These paths were based on the idea that we just wanted to put all the program files for KStars into /Contents/MacOS and the support files for KStars into /Contents/Plugins.  This includes not just support files for INDI drivers, but also for other Plugins and support files needed for KStars to run.

rlancaste:
sorry I wasn't done with my comments before you posted your next comment.  At runtime, in KStars, we set the environment paths to the various needed files at runtime so that the drivers and INDIserver knows where to find the files

rlancaste:
KStars has a separate path for the INDI drivers directory and the INDI server directory.  The driver needs to know where to look for the firmware files.  We made the environment variables so that we can define the path to the various components at runtime so the drivers know where to look for them.  We followed the basic scheme that was already followed by GPhoto.  And it turned out that this scheme also was good for GSC, for phonon, for Python, and even for dbus

rlancaste:
We were able to make KStars a completely modular app that can be dragged to different locations by wrapping everything up into the bundle and then defining any needed paths by the support programs like indi at runtime.

rlancaste:
It is possible that we could follow a different organizational scheme for the app bundle and this could reduce the number of environment variables needed.  We are flexible and can modify this to fit a new scheme if it would work better for you.  But we do want the paths to be configurable at runtime so that people can move the app and it will still work.  I understand you put them all into resources, but we didn't put them there because we have many more files that all need their paths defined.

rlancaste:
sorry I probably should have just posted one comment, but I found it easier on this computer to type it in separate comments so I could see what i was typing.  I apologize for that, this computer is not my mac.

rlancaste:
Oh I forgot to answer the other part of the discussions at the top.  When we modified the apogee indi driver, we removed the "rules" installation on OS X because it was trying to install the rules to a protected OS X folder, when we don't need rules on OS X anyway.  We also had made the firmware files install to /usr/local instead of /usr because /usr is also a protected folder on OS X.  For KStars it will not end up being there in the end, but we do need it to install someplace that we can copy it from when we put together the app bundle.  We do not expect most end users on OS X to use homebrew, but we do use homebrew while building the different components of the KStars app just because it is easier to do so.  So we needed the cmake build of indi to install the drivers, firmware, and other support files to a location that we can copy them from when we build kstars.  We also wanted to make sure that if a user still wanted to use indi in its traditional location at /usr/local/bin and the driver files in /usr/local/share/indi that our changes still allow that to work properly.  So no, we don't plan to have users install indi using homebrew, but we don't want to abandon support for that and we also currently use it to build the app bundle for KStars.

rlancaste:
Yes, I agree that we want it all to be packaged into the app, that's why we tried to make the path configurable so that it can go into /Contents, instead of having to be in /usr/local or in /usr.  That's the reason for the environment variable.  If it makes sense to call the environment variable something different that would be fine, but we do want the path to be configurable if the program needs it to be.  I might recommend APOGEE_FIRMWARE_DIR instead of the one Jasem used, but either way is fine.  There also would be no problem with the driver first looking for files in the same directory as the driver if the environment variable is not defined, and then finally looking in its default installation path on OS X as a last resort.  

I would like to make sure that the INDI drivers work for those using KStars, those using INDI Server, and those using INDI installed by homebrew.   (Believe it or not, there are still people who want to use a homebrew install of indi--personally I just use the indi version that we bundle into KStars).  But we do want them to work with whichever installation is being used.  

Yes I am aware that homebrew builds its files for the current OS X version and they are not backwards compatible and you cannot set the deployment target.  We ran into this issue awhile ago.  I actually build INDI and KStars on an El Capitan machine and then the dmg that we make will work on El Capitan, Sierra, and High Sierra.  At the moment the Mac OS X KStars dmg does not support older OS versions.

polakovic:
>
> I might recommend APOGEE_FIRMWARE_DIR instead of the one Jasem used, but
> either way is fine.


None is fine, it is still the same problem. INDI Server should set variable
specifically for apogee driver. Why? It will not work if somebody will
decide to change it for anything else. I'll not check every commit to INDI
repo just to make sure that nobody decided to introduce his own variable
for his driver today. Am I really the only one who think that it is wrong?

Believe it or not, there are still people who want to use a homebrew install


We have several hundreds downloads of INDI Server each month and 3 (three!)
requests for homebrew build since February 2015. So I do believe, but I
don't care :)

Yes I am aware that homebrew builds its files for the current OS X version
> and they are not backwards compatible


The problem is not just OSX version, but also processor type. There are
xeon, i7, i5, i3, core duo 2, core duo and maybe another machines out
there. E.g i5 code definitely doesn't work on core duo 2 (tested). And
there are subtle differences between different generations of the same
processor (different SSE instruction sets etc.).

>

rlancaste:
But anyway, focusing on the question at hand.  I would definitely want an environment variable to be available if the developer of an indi based program like KStars, INDI Server, or some other program wants to put the firmware/support files in a different location in the app bundle or someplace else on the system so that the driver can find it at runtime.  I definitely want to do this for organizational purposes because of the large number of such files.  As of now in KStars I placed each driver's firmware in a different folder in plugins.

But as you say, I don't think it should be required that you use the environment variable or need to know them in order for the driver to function in your program.  There should be a default location that makes sense.  I think astrometry.net and the location of the conf file is a good example.  It has an environment variable, but if that is not available then it first looks in the same folder as the executable first.  If the conf file is not specified and it is not in the same folder, then it checks in /usr/local/etc which is considered the default location for the file.  

I think that the indi drivers could all follow the same logic.  First, they could use the environment variable.  If its not defined, check in the same folder as the driver.  And if that doesn't work, check in the default OS X install location.  That way, KStars can use the first method, the INDI Server program you maintain can use the second method, and the home-brew users can use the third method.

Would that work for you?

rlancaste:
I know that all users who use INDI Server and most users who use Mac KStars will not be using the home brew version, but will be using the bundled INDI installed in each app respectively.  And while I do use home brew for building dependencies for KStars and INDI, I don't generally end up actually using INDI installed to /usr/local/bin except for testing purposes.  I too usually use the bundled INDI.  So you do have a fair point that in fact there are very few people using the home-brew installed INDI.  So maybe you do have a point that the fallback location with no environment variable should just be the same folder as the driver.  My big concern is that if somebody did want to use the home-brew version, all the support files such as hex files would go into /usr/local/bin since that is the default location for INDI on OS X.  And there are a large number of these files.

polakovic:
Well, not the same folder as the driver, but under the same common root as
the driver will probably work. Unfortunately I don't think there is
portable way how executable can discover its own path, at least I don't
know it. That's why I did it like this (see lines 125-137):

https://github.com/indilib/indi/blob/master/3rdparty/indi-qhy/qhy_ccd.cpp

rlancaste:
Yes, I have seen the QHY code, and it already is an excellent example of the plan I have been proposing, the QHY code includes the QHY_FIRMWARE_DIR environment variable we currently use in KStars as well as the fallback that will allow it to work in a home-brew install.  This is exactly what I had in mind.

In terms of your method for having the driver find its location, its a very ingenious idea.  That looks like it would work very nicely.  In KStars, I just set the environment variables before running the indiserver based on the location of the KStars application like this:
       QString environmentPath = QCoreApplication::applicationDirPath() + relativePath;
(where relative path is the path to the particular environment variable from the application path)
I have no idea if there is any similar method for non-qt programs.

My problem with using your method of putting all the files directly in one folder in KStars is the fact that there is a large number of driver support files.  Apogee has 386 text files, DSI has one hex file, QHY has 57 hex and img files, and Gphoto has about 70 files.  Its too many to put them in the same folder as all of the indi drivers and all the indi xml files as well.  Everything would get lost in there.  I would like to keep them separate.

I do have another idea if you don't like my previous idea (the one where QHY is the example).  We could set the cmakelists for installing apogee, dsi, qhy, and any others that come along like them so that they don't install to their current default locations on OS X by default, but instead install to a common firmware directory (like resources or plugins) as you propose, but in a subfolder of that directory unique to each driver.  So one idea (using Apogee as an example) could be a folder in the firmware directory to support the indi_apogee_ccd INDI driver called indi_apogee and that folder could contain all those text files.  One single environment variable could be used to define the firmware directory location instead of one for each driver.  And then each driver would append its own unique folder path to that firmware directory to know where to find its driver files.  For example, the Firmware environment variable might point to /Contents/Plugins/INDI/ or /Contents/Resources/ and then each driver's subdirectory would be in that location.  Like in the QHY code, If the environment variable doesn't exist and it is in the embed mode, it could default to resources and if it is not in embed mode it could default to something like  /usr/local/lib/INDI-firmware/

Thank you for discussing this with me and I hope we can come up with a good plan that will work well for everyone.

polakovic:
>
> My problem with using your method of putting all the files directly in one
> folder in KStars is the fact that there is a large number of driver support
> files. Apogee has 386 text files, DSI has one hex file, QHY has 57 hex and
> img files, and Gphoto has about 70 files. Its too many to put them in the
> same folder as all of the indi drivers and all the indi xml files as well.
> Everything would get lost in there. I would like to keep them separate.


OK, why not, INDI should provide just common root, driver or SDK can decide
what subfolder to use.

rlancaste:
Sounds good.  I will look at it probably on Wednesday.

rlancaste:
Hi Peter.  I just made some changes on my system in the INDI code that should make this work.  I sent a diff to Jasem so he can look over it to make sure I didn't do anything stupid.  Next I will send it to you so that you can check if it works for you.  Basically I modified the CMakeLists for DSI, QHY, and Apogee so that they install the firmware and camera text files into a subfolder in /usr/local/lib/indi/DriverSupport/ by default (for Homebrew).  In both KStars and INDIServer (and any other programs with INDI built in), the files would be in subfolders of /Contents/Resources/DriverSupport/.  This will allow us to get rid of the QHY_FIRMWARE_DIR, APOGEE_ETC_DIR, and DSI_FIRMWARE_DIR environment variables and just use a common root directory for all the files.  I also have no problem with putting them in Resources as long as they are in subfolders, so I thought it would be simplest to just use the same location as your INDIServer program.  Also it will get some things out of the KStars plugins folder which has quite a few directories in it already.  Hopefully this works for everybody.  

I was also looking through some of the other things you have in your pull request, many of them look like they are essentially already in master ( I mean the #include <memory> ones).  The changes to the Xcode project are of course not in there yet and they could be merged in.  There are some other changes you made that I assume could be merged in like the libusb changes.  One issue is that you reverted some changes in the libapogee CMakelists.txt which are actually needed for the home-brew and kstars build such as preventing the install of the rules file.  If you want I can try to merge some of your changes into my diff/patch before sending them to you so that you could apply it and test with hopefully all of your changes included.  Or I could send you my diff from master and you could see what else is needed.  Which would you prefer?

rlancaste:
Jasem said the code looked good, so I will await your response on my question and then I will get it ready to send to you.

Also would you prefer me to send it via email or to post it here?

rlancaste:
or possibly I could just submit a pull request so you can get it easier?

rlancaste:
I'll try this for now.  I am attaching a diff of my changes compared to the current master.  We can change it further to add your other changes.  Let me know what you think.

[changesForPath.txt](https://github.com/indilib/indi/files/1650912/changesForPath.txt)

polakovic:
Hi, definitely submit a pull request, I have no chance to look at it in
next few days. Peter

rlancaste:
Peter, I attempted to make a separate pull request.  Please take a look at it when you get a chance.  I tried to integrate as many of your changes as possible, though as I said many had already been done.  I did not include the Xcode project file since I think it would be better if you update that yourself.  There were a lot of changes to it and I don't want to mess it up.

knro:
Closing this, please follow up with this PR instead: https://github.com/indilib/indi/pull/505

PR #496, opened 2018-01-11
Asi ccd 20180107

wulfblat:
I avoided gratuitous changes, but the scope of moving the exposure handling from the TimerHit call back to the dedicated video thread was necessarily large.

As stated in the commit comments, the reason I moved the exposure handling into the thread is because I was seeing a consistent overhead of between 1.47 and 2.1 seconds in TimerHit handling the final phase of the exposure. All other ASI timer events were thus delayed by this overhead, including guiding (exposures and pulses).

TimerHit is now used only for cooler control and temperature monitoring. Separate call back functions are used for NS and WE guiding. Delays in the timer call backs are all kept at under 1ms to keep the other timers on-time (or close to it).

I have tested the changes over hundreds of exposures plus live video on an ASI294MC Pro and an ASI174mini.

I have not yet had a chance to test the guide port on the 174, but the primary change here was splitting the existing code from TimerHit into separate timer call backs. Once I have a clear night sky I will test this as well.

Comments and suggestions are very welcome.


wulfblat:
The two failures were for indi-dsi, which I did not modify.

knro:
It seems all functions are working except for guiding, so please check that.

jochym:
clang4 and osx errors are the same - unused variables warnings treated as errors in dsi. Didn't we deal with this recently @knro ?

wulfblat:
Thanks! I'll look over the guiding code in detail today.

knro:
@jochym I fixed the unused variables error, but if you check the first error is _not_ related to variables, but something else.

wulfblat:
I had forgotten to convert the pulse time from milliseconds to seconds when setting the pulse request duration variable.

knro:
Have you tested guiding now?

wulfblat:
We have had nothing but for and rain for the past week, and so I have not been able to test it under the stars. I did create a simple test to verify the guide pulses were working for the correct amount of time (by manually calling each of the four cardinal guide pulse functions with durations ranging from .001 to 1.5 seconds).

wulfblat:
We have a clear sky tonight and I was able to use the ASI174mini in an OAG to guide my Losmandy Titan (gemini controller). This was my first time guiding with Ekos, and I still need to tune the settings for my setup, but the guider was working.

knro:
Hello Leonard . Today I added a new feature to INDI::CCD Rapid Looping. So basically, instead of waiting for upload to be complete, it would immediately take the next image. This works fine in other drivers, but it's causing a deadlock in INDI ASI driver once "Rapid Looping" is enabled in options. 

I checked the code, and looks like the reason is because grabImage() eventually calls ExposureComplete() which then immediately triggers StartExposure() again, and then it gets deadlocked trying to obtain the mutex. I tried to find a solution but not sure what's the optimal way to fix this, introduce another state? or what? Thanks!

wulfblat:
Thanks for the notice Jasem.

I have the day off and can work on a solution.

Is the "Rapid Looping" on master, or is it branched or in a forked repo?

Leonard

knro:
Yes it's on master. To enable it, go to "Options" and it should be there. Try it with the CCD Simulator. You also need KStars master if you want to make use of that in Ekos.

wulfblat:
Thanks.

I have the latest kstars master source and so I can quickly build that.

At first glance it is a bit puzzling because we release the mutex before 
calling grabImage, but I'll find and fix the problem today.

Leonard

wulfblat:
Hi Jasem,

Did you see a deadlock in the debugger, or did you see the thread simply 
waiting?

If not an actual dead-lock, I see an issue where after calling 
grabImage() we call exposureSetRequest(StateIdle), which will clear the 
StateStartExposure value set in StartExposure() when called from 
grabImage().

If this call is moved before grabImage() it should solve this particular 
issue:

*** asi_ccd.cpp.orig    2018-02-19 07:03:37.032471167 -0800
--- asi_ccd.cpp    2018-02-19 07:04:07.380531678 -0800
***************
*** 1862,1870 ****
                   PrimaryCCD.setExposureLeft(0.0);
                   DEBUG(INDI::Logger::DBG_SESSION,
                       "Exposure done, downloading image...");
-                 grabImage();
                   pthread_mutex_lock(&condMutex);
                   exposureSetRequest(StateIdle);
                   break;
               }
               else if (status == ASI_EXP_FAILED)
--- 1862,1872 ----
                   PrimaryCCD.setExposureLeft(0.0);
                   DEBUG(INDI::Logger::DBG_SESSION,
                       "Exposure done, downloading image...");
                   pthread_mutex_lock(&condMutex);
                   exposureSetRequest(StateIdle);
+                 pthread_mutex_unlock(&condMutex);
+                 grabImage();
+                 pthread_mutex_lock(&condMutex);
                   break;
               }
               else if (status == ASI_EXP_FAILED)

I will test this out as soon as I get back in from taking care of the 
horses.

Leonard

wulfblat:
I tested this change this morning and can connect, enable rapid looping, 
and then capture a series of images. This is using an ASI174mini locally 
connected.

With rapid looping enabled I see in the imaging dialogue box:

    2018-02-19T09:02:10 Received image 5 out of 5.

    2018-02-19T09:02:09 Received image 4 out of 5.

    2018-02-19T09:02:07 Received image 3 out of 5.

    2018-02-19T09:02:06 Received image 2 out of 5.

    2018-02-19T09:02:05 Received image 1 out of 5.

    2018-02-19T09:02:03 Capturing image...

Which differs from the output when rapid looping is disabled:

    2018-02-19T09:00:04 Received image 5 out of 5.

    2018-02-19T09:00:03 Capturing image...

    2018-02-19T09:00:03 Received image 4 out of 5.

    2018-02-19T09:00:01 Capturing image...

    2018-02-19T09:00:01 Received image 3 out of 5.

    2018-02-19T08:59:59 Capturing image...

    2018-02-19T08:59:59 Received image 2 out of 5.

    2018-02-19T08:59:57 Capturing image...

    2018-02-19T08:59:57 Received image 1 out of 5.

    2018-02-19T08:59:55 Capturing image...

I will fork the repo and create a branch with the change this morning, 
although if you'd prefer you can simply apply the change from the diff I 
sent earlier.

However, when I connected remotely to an ASI294MC and try the same thing 
with rapid looping enabled image capturing goes into an infinite loop of 
capturing images (even with just "preview"), with each attempt popping 
up its own fits viewer with a message "Please hold, downloading FITS 
file" and the status never gets beyond 65%. I have to kill kstars to 
stop the loop.

I experimented and found this only happens with short exposures. The 
ASI294 image size in 8 bit mode is 11MB (as opposed to 2MB for the 174), 
and I suspect there is an error when the download time exceeds the 
exposure time that causes this infinite loop. I do not think it is in 
the asi driver, but I will investigate further today.

This does not occur with rapid looping disabled.

Leonard

knro:
Thanks for the testing. There is indeed other major issues. Since the upload function is in the same thread, startExposure can start but no updates are sent until the upload is complete. Furthermore, there are 2 pipes between the driver and INDI server, and the same upstream pipe used for INDI XML is also used for BLOBs so the pipe is saturated until the upload is complete. So what you said above coupled with these issues is going to make this more difficult than I initially thought, but I'll tackle it again tomorrow and it is disabled by default.

PR #497, opened 2018-01-11
Explore Scientific G11 PMC-Eight Telescope Driver

mikefulb:
I am submitted a new driver for the G11 PMC-Eight mount from Explore Scientific.  It was based on the IEQPRO driver.  I have been using it for a few months and it has been used for imaging with good success.  I still consider it under development so I have put warnings in the code and startup messages as well on the website documentation so users will be aware to be cautious when initially using the driver.  

I currently do not support use in the Southern Hemisphere as I have not been able to verify its behavior in that situation.  If we get someone knowledgeable who could test that I could remove the restriction.


mikefulb:
I believe I fixed the compile errors.

mikefulb:
Ok that should do it hopefully.

knro:
Ok, indicom.c should retain the Gemini changes, even though I think it should be perhaps done in a better way. But for now, you changes do not take it into effect.

PR #504, opened 2018-01-21
Lx200ap experimental 20171230

mikefulb:
This update contains improvements to the lx200ap_experimental telescope driver.  

Changes:

- checks for 'V' level of firmware and will not load if not present
- adds ability to choose from 4 park positions as defined by Astro-Physics
- better mount initialization


mikefulb:
I committed a change that should fix the build_clang4 issue.

PR #505, opened 2018-01-23
Path updates and OS X Changes

rlancaste:
This pull request will apply the OS X installation path changes for KStars, INDI Server, and home-brew INDI as was discussed by Peter and myself.  In a home-brew install, the firmware and other camera support files will install to subfolders in /usr/local/lib/indi/DriverSupport.  Currently in a home-brew install, these subfolders will be for DSI, Apogee, and QHY files.  For KStars and INDIServer, these files will be placed in a similar manner into the app bundle in /Contents/Resources/DriverSupport. Thus at runtime, if the INDIPrefix Environment variable exists on OS X such as in KStars and INDI Server, the drivers will look for their files in this folder in the app bundle.  But if it doesn't exist, such as in a home-brew install, then it will look in the default folder listed above.  This will eliminate separate firmware location environment variables for each driver and just use the one.  It will also ensure that each driver gets its own unique folder for its files (helpful for the embedded versions), and that they are all collected in one place (helpful for the home-brew version).  I also attempted to integrate some of the other OS X driver updates that Peter had integrated in his pull request, but others seemed to already have been done.  One item I did not include in this pull request that Peter had done that will still need to be updated is the Xcode project file since there were so many updates in it. 

I am attaching a screenshot to show how this folder is organized.  Note that it is in the KStars app bundle so it includes gphoto as well, but this is not a part of this pull request, it is a separate part of the KStars app bundle build.  But this screenshot should show the general layout of the folder.

screen shot 2018-01-22 at 11 08 58 pm

### DIFF
--- a/3rdparty/indi-apogee/apogee_ccd.cpp
+++ b/3rdparty/indi-apogee/apogee_ccd.cpp
@@ -40,12 +40,21 @@
 
 #include <memory>
 
+#ifdef OSX_EMBEDED_MODE
+ #include "Alta.h"
+ #include "AltaF.h"
+ #include "Ascent.h"
+ #include "Aspen.h"
+ #include "Quad.h"
+ #include "ApgLogger.h"
+#else
 #include <libapogee/Alta.h>
 #include <libapogee/AltaF.h>
 #include <libapogee/Ascent.h>
 #include <libapogee/Aspen.h>
 #include <libapogee/Quad.h>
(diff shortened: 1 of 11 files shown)


rlancaste:
Please let me know if there are any problems

knro:
Ok I'll merge this for now. If you and Peter arrive at a different solution, please submit a new PR.

PR #513, opened 2018-02-07
initial refactorings on imager agent

GuLinux:
- Improve encapsulation
 - Extract Group into separate file
 - Avoid raw strings
 - Extract utility methods like `currentGroup` and `nextGroup`


knro:
@polakovic any thoughts?

polakovic:
It looks correct.

GuLinux:
Just some heads up: this is a first step on various refactoring/improvements I'd like to make.
After this, I'd like to extract the sequence steps as well, and (my final goal) also add a few more controls, mainly:
 - Gain setting
 - Some kind of callback support pre/post shot and pre/post group (maybe a "launch script" field?)

knro:
You want to make more changes to the PR or should I merge it?

GuLinux:
I think it would be better to have multiple PRs, in order to have more incremental changes, it's probably even easier for you to review them (plus, I don't really have a lot of free time, so I'm not sure when the next changes will come though)

PR #521, opened 2018-02-16
help IDE with namespace, more wheel error checking, fix wheel disconnect bug

d33psky:


knro:
What's the reason for ASI_CCD_ISInit and ASI_EFW_ISInit? Each driver is compiled separately.

d33psky:
It helps my IDE 'explain' the namespace here, the ISInit of indi_aagcloudwatcher.h gets listed as definition for ISInit in asi_wheel and asi_ccd which is not helping of course.

3rdparty/indi-aagcloudwatcher/indi_aagcloudwatcher.h:void ISInit();

PR #527, opened 2018-02-26
Updated Maxdome II driver

juanmb:
I fixed some bugs in Maxdome II driver. I also refactored the code into a more object-oriented style.


knro:
Thanks Juan. A couple of comments:
1. I just added PollingPeriod property to DefaultDevice which all other drivers are using now, so please update your driver to use it accordingly.
2. The logging macros are useful to all INDI, can you make another PR to mass-change them?

juanmb:
OK. I removed the global POLLMS constant.
I'll make another PR adding the logging macros to indilogger.h.

PR #533, opened 2018-02-28
Add Skywatcher Alt-Az Simple mount driver

kecsap:
It is "forked" Skywatcher Alt-Az mount driver. Changes:
- Removed alignment subsystem codes
- Alt/Az <-> Ra/Dec conversions are based on pure libnova
- Supporting wedge and "EQ wedge" (placing the Alt/Az mount on the top of an EQ mount) modes
- Guiding support in wedge/EQ wedge modes.

### DIFF
--- a/3rdparty/indi-mgen/indi_mgenautoguider.xml
+++ b/3rdparty/indi-mgen/indi_mgenautoguider.xml
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <driversList>
-<devGroup group="Aux">
+<devGroup group="Auxiliary">
     <device label="Lacerta MGen Autoguider">
         <driver name="MGen Autoguider">indi_mgenautoguider</driver>
-        <version>0.2</version>
+        <version>0.1</version>
     </device>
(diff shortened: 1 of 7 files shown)


knro:
The name SW AltAz is confusing I think to end-users.

kecsap:
What name do you prefer? Can we just ditch the old driver? It will be anyway in git history and the alignment subsystem does not give any benefit in performance for these cheap Alt/Az mounts. I was able to do the same exposure duration with the old and the refactored driver.

knro:
Does your specific mount have a brand? Maybe use that.

kecsap:
This driver is generic Alt/Az driver like the old one. Now it was tested with Skywatcher Merlin and Skywatcher Virtuoso.

knro:
You never know maybe there are users out there who uses it. So think of a name but let's keep the existing one.

kecsap:
"Skywatcher Alt/Az Reborn"?

knro:
haha well I don't think we can use that. Maybe SkyWatcher Alt-Az Wedge?

PR #538, opened 2018-03-11
Fixes and better logging for scripting gateways

polakovic:


knro:
Thanks Peter! Can you please check the CI issues?

polakovic:
Hm, ignoring return value on tmpnam() with parameter is an error? :) Peter

knro:
All warnings are treated as errors. I guess we could add an exception for this specific warning but it would be better if you can solve it first.

PR #540, opened 2018-03-13
udev rule for /dev/integra_focusing_rotator0 and set the mandatory ba…

d33psky:
…udspeed. Also prevent the modem manager to probe the device as it breaks it.


knro:
Is this VID:PID unique to Integra or just some generic adapter? Just to be sure before merging this.

d33psky:
Good point. Turns out to be a regular Arduino :-/
I think we then should provide the udev rule as a hint for the user.
Shall we comment-out the udev line ? (So that a user can still enable it)

knro:
You can leave it commented if you include more information there. Also, this means we can only set default baud now, not port.

PR #543, opened 2018-03-22
Fix some casting issues on arm (Synscan Mount)

kecsap:


jpaana:
I think the root cause is that unlike int and short, "char" type is by default signed on x86 but unsigned on ARM, which is why it's always good to speficy the signedness if using it for anything else than ASCII text.

kecsap:
@TallFurryMan @jpaana What I don't understand that static_cast<int>(negative double) results zero on arm, but a correct, negative value on x86. I think it is a huge problem, if we can't trust these basic C++ constructs. static_cast has been told to be used because C-like cast or missing cast can result incorrect value, but it seems static_cast is also not reliable.

jpaana:
Hm, no, static_cast to int works fine on both, the issue is the final assignment to char str[20]; which is unsigned on ARM and assigning negative values to unsigned types is undefined for a reason. By casting to signed int first just happens to trigger a different undefined behavior as conversion from double to char is different than int to char. The blunder is that C defined all other "base types" (short, int, long, ...) as signed, but char is defined as "Smallest addressable unit of the machine that can contain basic character set. It is an integer type. Actual type can be either signed or unsigned. It contains CHAR_BIT bits."

A simple demonstrator:

int main(void)
{
   char c;
   signed char sc;
   unsigned char uc;
   int i;
   double d = -42.0;

   c = static_cast<char>(d);
   uc = static_cast<unsigned char>(d);
   sc = static_cast<signed char>(d);
   i = static_cast<int>(d);

   printf("char %d unsigned char %d signed char %d int %d\n", c, uc, sc, i);
   return 0;
}

outputs on x64:
char -42 unsigned char 214 signed char -42 int -42

on ARM by default:
char 0 unsigned char 0 signed char -42 int -42

but if compiled with -fsigned-char:
char -42 unsigned char 0 signed char -42 int -42

So assigning negative values to signed types always works fine, it's just when assigning to unsigned type all bets are off.

TallFurryMan:
Excellent demonstration.

TallFurryMan:
We'd be better off using uint8_t and int8_t to explicitly describe sign. But sometimes, for instance, it's convenient to refer to "the most effective loop counter on this architecture" as "unsigned int" or "signed int", which I'm not sure the explicit format provides.

TallFurryMan:
Ah it seems c++11 provides int_fast(8|16|32|64)_t for this exact purpose. I should update my reference....

PR #545, opened 2018-03-28
Compilation fix [3rdparty]

kecsap:


knro:
Can you suppress the other warnings until QHY provides a fix hopefully soon?

PR #546, opened 2018-03-29
Fix for Longitude not set correctly.

kbahey:
As detailed in the forums

http://www.indilib.org/forum/mounts/3231-ekos-sends-wrong-latitude-longitude-to-mount.html


kbahey:
Any idea when will this be in a release? Given the new stable scheme, are they timed releases, or whenever enough features/fixes are included?

knro:
It's the later case, I think hopefully within the next 2 weeks.

PR #548, opened 2018-03-29
Clear allocated IText structs for filter names to avoid crash

jpaana:
Recompiled INDI today after the memleak fixes and noticed that my SX wheel driver crashed with corrupted free. Noticed that it tried to free "text" member of an IText struct and it was garbage. Back trace showed it had been allocated with new() which doesn't initialize the struct in any way unless it has a constructor. So clear the tables to avoid this. More foolproof way would be to add default initializer to the struct though.

### DIFF
--- a/libindi/libs/indibase/indifilterinterface.cpp
+++ b/libindi/libs/indibase/indifilterinterface.cpp
@@ -124,6 +124,7 @@ bool FilterInterface::processText(const char *dev, const char *name, char *texts
                 delete [] FilterNameT;
 
             FilterNameT = new IText[n];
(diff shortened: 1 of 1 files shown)


knro:
yes a default constructor to the struct is better.. I also ended up adding {} everywhere to IText but this is not a good solution. I started actually to write the default constructor but then went into a rabbit-hole about initialization of the char array within the default constructor. I think I'll read more about it now and implement it since it might cause issues for other drivers too.

knro:
Ok looks like I'll adopt your solution in addition to other misc changes to free allocated memory before deleting FilterNameT

PR #549, opened 2018-03-30
Add binning support based on capture format (GPhoto)

kecsap:


knro:
So if I understand this correctly, binning is done by taking smaller jpeg images? Is the size as required from binning (1/4th for 2x2) or it's just smaller in size? Also, the formats containing "Medium", "Small", and "Smaller" are only available in Canon DSLRs IIRC.. need to check if Nikon, Sony, Pentax..etc support this method or not? If not, then binning should **only** be enabled for Canon.

kecsap:
1. It just smaller size. There is no way to manipulate the size.
2. I updated the patch now. The driver checks the transfer formats now during connection and it enables this binning emulation only if there is at least two formats containing "Medium" and "Small" in their names. "Smaller" is not always available even in Canon DSLRs.

PR #550, opened 2018-04-02
indi-eqmod: New property to avoid meridian flip

kkretzschmar:
This change introduces a new switch to avoid a meridian flip when triggering a goto, by exposing the existing forcwup feature. First test seems to be ok, but test on star is still to be done ....  

@geehalel: Jean-Luc, pease find my change for enabling the enforcing-counterweight-up feature for review. As already mentioned test on star is pending ...


knro:
Maybe also save this in the config? I think whenever this is turned out, there should be a warning to inform the user about this since it can lead to dangerous situations.

kkretzschmar:
This switch is by default set to "False" or "ENABLE=OFF"  and it should be used only temporarily. Hence IMHO it should  not be part of the config settings.  Yes writing an warning and informing the user is a good idea, I'll add that.
Btw I got some error ci/circleci build errors in "qhy_ccd_test.cpp", but my change didn't touch that file ... or maybe the problem is already fixed?

knro:
Make sure to rebase on master when you submit your next commit since the QHY error was fixed already.

knro:
Any update on the PR?

kkretzschmar:
Yes, I was able to test the new functionality yesterday night. It works. I found out that the "forcecw" flag is overwritten (disabled) if the hour angle becomes too close for potential collisions. So there is is already a "safety net", which triggers a meridian flip even if the forcecw flag is enabled.  

@geehalel: Nice work!

PR #564, opened 2018-04-10
Initialize ConnectionModeSP to avoid uninitialized access

jpaana:
Forgot this from previous patch, its name is compared against in DefaultDevice::ISNewSwitch() before being actually initialized in DefaultDevice::ISGetProperties(). Probably harmless, but gave a Valgrind warning anyway.


jpaana:
There's a better way to fix this, will do that.

PR #571, opened 2018-04-12
Updates to pyxis rotator to support 3 inch and 2 inch rotators

deisenlord:
Jasem, my changes.
- Added firmware and model parameters to an info tab
- Use the CVxxxx command to return the firmware version of the rotator and set
    - Firmware parameter
    - Model parameter (3 inch or 2 inch)
    - Correct default rotator rate (delay) of 6ms or 8ms (2 inch)
- Changed isMotionComplete to loop and consume all microsteps and update the PA parameter.
- Sized read in isMotionComplete to be approximately 1 degree so GUI update occurs each degree. Number of steps in 1 degree dependent on rotator model.
- Fixed setRotationRate to read response which was not being done and made setting the rate non-reliable.

I'm new to the indi driver parameter/property api so please take a good look at my changes. Also is my approach, the while loop calling isMotionComplete in TimeHit, ok?   Waiting for the POLLMS interval caused GUI to lag motion.
Best,
David

### DIFF
--- a/libindi/drivers/rotator/pyxis.cpp
+++ b/libindi/drivers/rotator/pyxis.cpp
@@ -33,6 +33,14 @@
 #define PYRIX_CMD 6
 #define SETTINGS_TAB    "Settings"
 
+// Recommended default rates for 3 inch and 2 inch rotators
+#define PYXIS_3INCH_RATE 6
+#define PYXIS_2INCH_RATE 8
+
+// Number of steps per degree for rotators
+#define PYXIS_3INCH_PER_DEG 128
+#define PYXIS_2INCH_PER_DEG 14
+
 std::unique_ptr<Pyxis> pyxis(new Pyxis());
 
 void ISGetProperties(const char *dev)
@@ -99,9 +107,18 @@ bool Pyxis::initProperties()
     IUFillSwitch(&PowerS[POWER_WAKEUP], "POWER_WAKEUP", "Wake Up", ISS_OFF);
     IUFillSwitchVector(&PowerSP, PowerS, 2, getDeviceName(), "POWER_STATE", "Power", SETTINGS_TAB, IP_RW, ISR_ATMOST1, 0, IPS_IDLE);
 
+    // Firmware version
+    IUFillText(&FirmwareT[0], "FIRMWARE_VERSION", "Version", "Unknown");
+    IUFillTextVector(&FirmwareTP, FirmwareT, 1, getDeviceName(), "FIRMWARE_VERSION", "Firmware", INFO_TAB, IP_RO, 0, IPS_IDLE);
+
+    // Firmware version
+    IUFillText(&ModelT[0], "HARDWARE_MODEL", "Model", "Unknown");
+    IUFillTextVector(&ModelTP, ModelT, 1, getDeviceName(), "HARDWARE_MODEL", "Model", INFO_TAB, IP_RO, 0, IPS_IDLE);
+
+
     serialConnection->setDefaultBaudRate(Connection::Serial::B_19200);
 
-    return true;
+    return true ;
 }
 
 bool Pyxis::Handshake()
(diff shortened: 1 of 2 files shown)


knro:
doing a while ( condition) loop would result in a blocking driver that cannot receive/send commands until the motion is complete. This approach is generally discouraged and this is while it is done in TimerHit() so that it won't be blocking, especially if the motion can take several seconds to complete.

Now I don't know the impact my changes made. Does the function usually end up with OVERFLOW or TIMEOUT responses before finally returning true?

deisenlord:
﻿Yeah, I figured this would be "bad" and understand the blocking now.  When I call isMotionComple just once from TimerHit the problem is that for long rotations the GUI ends up significantly lagging the actual rotation.   For a 30 degree rotation there will be 30 calls to isMotionComplete the way I have the buffer sized.  It's sized for reading 1 degree of microsteps.  That means 30 * 100ms delays = ~ 3 seconds.   So the rotator stops moving and there are approximately 3 reads left to be done so the GUI ticks off 28, 29, 30 after the rotation has stopped.   With the 6ms standard delay moving 1 deg (128 steps) takes 768ms.  Having a brain storm as I write this maybe I should just read ~100ms of steps at a time.  Is it ok for me to update the PA parameter every 100ms?    Or maybe I could update the PA at a multiple of 100ms,  maybe every second.   Do either of these sound better?

knro:
What if you increase the buffer size to say, 512 bytes for example? or even 1024? what would happen.

deisenlord:
So for 1024 the gui would update ~6 seconds and every 8 degrees and the read in isMotionComplete would take ~6 seconds to complete. 

Maybe I just need to read 1 degree of ticks + ~100ms of ticks to stay approximately synchronized.

deisenlord:
Jasem, I need to open my eyes wider...  The driver poll time was 1 second not 100ms which is what was causing the gui to fall behind.  I changed the poll time during the rotation to 100ms which resolves all problems with the rotator and gui staying up-to-date.

I also appear to have founds some bugs in the rotator firmware where it gets confused around 180 degrees.  I can ask it to go from 179 to 181 and it will go 540 degrees instead of just 2 which would wrap cables, yikes.  I'm almost positive its not an indi driver issue, as we are just requesting 179 then 181 degrees.  My firmware is old and will send out for a new chip to update it.

I will let you know what I hear from Optec regarding the behavior.  If its a firmware bug I don't think we need to delay release but you're the man.

knro:
This is good to go for now, let me know what Optec has to say about the rotation issue.

deisenlord:
Jasem, I discovered a problem with the original code's RotatorReverse functionality.  I couldn't effect the direction of the rotator from the control panel which was confusing me.   I believe the problem was a typo in one of the base classes indirotatorinterface.cpp so please take a look at my change.
I also figured out the behavior I asked Optec about, its a feature of the rotator where it will only rotate +-180 degrees from the home or zero position.  This prevents cable wrapping, so if your at 179 and ask to goto 270 it will go the long way around to prevent cables seeing more than 180 degrees of rotation.
Another thing I discovered is that in order to set the rotation direction via RotatorReverse the device has to be in the home position.   Nothing is enforcing this currently.  I will think about this.  Also when the rotation direction is set it is saved in EPROM.    Sorry I didn't get this stuff in the first go.
Best David

deisenlord:
Yes, thats what the pyxis is doing, the input is still 0-360, but it implements that as two ranges of 180 and moves between them such that cables never wrap.  As far as the code goes it only effects how I track PA while its moving to be consistent with how the device actually moves between angles.



In indirotatorinterface.cpp :



             bool rc = false;

-            bool enabled = (!strcmp(IUFindOnSwitchName(states, names, n), "ENABLED"));

+            bool enabled = (!strcmp(IUFindOnSwitchName(states, names, n), "REVERSE_ENABLED"));

             rc = ReverseRotator(enabled);



I've updated my branch.

David





-----Original message-----

knro:
Ok thanks for the note, I actually changed the name itself to be just ENABLED and DISABLED.

PR #574, opened 2018-04-14
Add NIR preflash for Moravian CCD if hardware supports it

StarryBird:
The latest version of Moravian CCD supports Near-IR Preflash to deal with the RBI effect. See here for further information: [What is Residual Bulk Image and how to deal with it](http://gxccd.com/art?id=418&lang=409)

Current INDI driver seems not have NIR preflash settings. I added related settings in driver Main Control page.

The setting contains to options:

1. Preflash duration in seconds. Value ranges from 0s to 30s with minimum step 0.01 s. For 0s duration it just disable preflash.

2. Number of clears. Value ranges from 1 to 10.

I have tested the code with my G2-3200. The CCD works OK with some delay before normal exposure (depends on both duration and number of clears settings).

### DIFF
--- a/3rdparty/indi-mi/mi_ccd.cpp
+++ b/3rdparty/indi-mi/mi_ccd.cpp
@@ -214,6 +214,8 @@ MICCD::MICCD(int camId, bool eth) : FilterInterface(this)
     hasGain    = false;
     useShutter = true;
 
+    canDoPreflash = false;
+
     setDeviceName(name);
     setVersion(INDI_MI_VERSION_MAJOR, INDI_MI_VERSION_MINOR);
 }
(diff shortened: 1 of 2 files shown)


knro:
@sWski Any thoughts on this?

sWski:
@knro Looks good, preflash is fully implemented in the gxccd library.
@StarryBird Maximum values supported are slightly bigger - 65s for preflash (65535ms to be precise) and maximum number of clears is 16. You can update that in your code, however one should not need bigger values than those mentioned in our article.

StarryBird:
Hi @sWski : 

     I have updated my code. Now maximum preflash duration can be 65.535 s and minimum step is 0.001 s. Maximum number of clears is set to 16. The default value is not changed.

PR #575, opened 2018-04-14
Fix problems with Pyxis Reverse functionality

deisenlord:
Jasem, my changes.
- Added firmware and model parameters to an info tab
- Use the CVxxxx command to return the firmware version of the rotator and set
    - Firmware parameter
    - Model parameter (3 inch or 2 inch)
    - Correct default rotator rate (delay) of 6ms or 8ms (2 inch)
    - Changed isMotionComplete to loop and consume all microsteps and update the PA parameter.
    - Sized read buffer in isMotionComplete to be approximately 1 degree so GUI update occurs each degree.  Number of steps in 1 degree dependent on rotator model.
     - Fixed setRotationRate to consume terminator which was not being done and made setting the rate non-reliable.

I'm new to the indi driver parameter/property api so please take a good look at my changes. Also is my approach, the while loop in isMotionComplete, ok?   Waiting for the POLLMS interval caused GUI to lag motion.
Best,
David

### DIFF
--- a/libindi/drivers/rotator/pyxis.cpp
+++ b/libindi/drivers/rotator/pyxis.cpp
@@ -170,7 +170,6 @@ void Pyxis::queryParams()
     // Reverse Parameter
     ////////////////////////////////////////////
     int dir = getReverseStatus();
-
     IUResetSwitch(&ReverseRotatorSP);
     ReverseRotatorSP.s = IPS_OK;
     if (dir == 0)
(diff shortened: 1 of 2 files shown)


deisenlord:
Sorry forgot to create a new PR for the changes

knro:
indirotatorinterface.cpp was already updated, update PR accordingly.

deisenlord:
Revered the change to indirotatorinterface.cpp, thanks for your patience.
David

PR #583, opened 2018-04-29
update weather watcher driver

fenriques:
Main reason for this change is to allow users to specify their own keywords for 4 weather related parameters: temperature, rain, wind, wind gust. 
So before connecting you have to set in option tab the file  (or URL) containing weather data and the 4 keywords matching the names in the file.

### DIFF
--- a/libindi/drivers/weather/weatherwatcher.cpp
+++ b/libindi/drivers/weather/weatherwatcher.cpp
@@ -115,26 +115,26 @@ bool WeatherWatcher::createPropertiesFromMap()
 
     for (auto const& x : weatherMap)
     {
-        if (x.first == tempT[0].text)
+        if (x.first == keywordT[1].text)
         {
             addParameter("WEATHER_TEMPERATURE", "Temperature (C)", -10, 30, -20, 40);
             setCriticalParameter("WEATHER_TEMPERATURE");
         }
-        else if (x.first == windT[0].text)
+        else if (x.first == keywordT[2].text)
         {
             addParameter("WEATHER_WIND_SPEED", "Wind (kph)", 0, 20, 0, 40);
             setCriticalParameter("WEATHER_WIND_SPEED");
         }
-        else if (x.first == gustT[0].text)
+        else if (x.first == keywordT[3].text)
 		{
             addParameter("WEATHER_WIND_GUST", "Gust (kph)", 0, 20, 0, 50);
 		}        
-		else if (x.first == rainT[0].text)
+		else if (x.first == keywordT[0].text)
         {
             addParameter("WEATHER_RAIN_HOUR", "Rain (mm)", 0, 0, 0, 0);
             setCriticalParameter("WEATHER_RAIN_HOUR");
         }
-        else if (x.first == forecastT[0].text)
+        else if (x.first == keywordT[4].text)
         {
             addParameter("WEATHER_FORECAST", "Weather", 0, 0, 0, 1);
             setCriticalParameter("WEATHER_FORECAST");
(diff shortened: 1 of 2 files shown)


knro:
Why not use default values for those keywords instead of leaving them to nullptr?

fenriques:
Simply because I had no idea that the 4th parameter of IUFillText was the default value...it was not used in the file / url Text and I worked only with copy&paste...

In this commit:
- Added default values for stored keywords
- Added Forecast keyword
- Tried to set a custom separator value with default value '='. Everything was fine but I got lost in C++ pointers, operators etc. My intention was to have in CreateMap something like:
`    while(std::getline(std::getline(iss, key, separator ) >> std::ws, val))`
but 
   ` char separator = separatorT[0].text; 
`
is a reference to the pointer I guess.

knro:
Is there a reason why there is a separate property for each keywords? You can have them all in one property to simplify things. The char is separatorT[0].text[0]

fenriques:
- Now used same property for all weather keywords but kept File and Separator in their own separate property. It made more sense to group just the weather related keywords in the same property.

- The separator can now be customized.

- A question regarding ranges: does make sense to have BOTH a 'max OK' AND 'min Warn'? Shouldn't be the same value?
For example if Wind is dangerous above 20km/h i would set Max OK = 20 AND Min Warn =20. What is the benefit of keeping them distinct?

knro:
Right max OK & min Warn are redundant. If you want to change that, please feel free to make another PR against indiweather.cpp ;-)

fenriques:
Hi Jasem,
took a little to make up my mind, I would like to share some thoughts on weather ranges (see previous post) before taking any action.
 
1.a - Best solution for me is to have just 2 parameters: Warning and Danger, no OK parameter. For example if you set Wind_Warning = 20 and Wind_Danger = 30, **anything** below 20 will be ok, no need to have OK limit (min nor max). 
1.b - Stretching this even further we could have just 1 parameter: Danger. Warning would be calculated as global percentage set on the property. E.g. if Warning_range = 20%, you would have a Wind Warning at 24 (30 - 20%), Temp Warning at 32  (if Temp Danger = 40) etc.

2- Other question. Does make sense to have a 'reverse direction' switch for each range? As of now every parameter assumes that the higher the value, the higher the danger; if Wind_Danger = 20, danger is higher than 20 not lower. But maybe I would like to set a warning for lower temperature, eg closing the dome if temp is lower than -20. Or Forecast in some weather station could have better weather at higher values. 

3 - Any change to present behaviour will break compatibility with previous versions. Worth doing that? is there a way to prevent it?

4- Not related to ranges, but if I were more skilled on Ekos front end, I would like to design a Weather Tab for Ekos window (mean not INDI panel) that shows recorded weather data in graphics for Rain, Temp, Wind / Gust and Forecast (like guiding graph for example). One could easyly monitor weather status, maybe export weather data as .csv for further inspection against captuered images etc.

Have any thoughts? or shall post this on the forum asking for users opinion

knro:
I think we could introduce a "Rule Set" for each parameter and have it parse the rule and apply it. Now I do not know what syntax should we use for this, but this would eliminate adding all kinds of extra properties to handle all the various cases. This would certainly some compatibility but the main point about shutting down the observatory still holds.

For Ekos, yeah something like that can be done of course, but would require quite a bit of work.

fenriques:
you mean the user writes an expression  like:
`x <  -20 OR X > +30`
where x is temp or rain etc.? I would say only OR (not AND) admitted. Could be quite easy to parse and validate.
Warnings should then be handled as percentages of danger ranges else would be too complicated to ask for another expression for warning ranges.

And maybe add also a default base behavior when the user just writes a single number eg 30 that would mean danger if x > 30

About the Ekos tab could you just help me setting up a blank page when a weather driver is selected?
And point me to a file where a graph is managed (eg the tracking graph in ?
I doubt I will every commit something stable enough for production but will use it to learn something more.

knro:
We could have 'and' as well for other rules. indi_eval does exactly that, it evaluates properties and returns true or false using many modifiers. Perhaps its logic can be used here as well. We can have 2 rules per property. the OK and WARN rules.

fenriques:
was expecting a indi_eval file somewhere. But i think you are referring to compiler.c, am I right?
/* module to compile and execute a c-style arithmetic expression made of INDI
 * operands. operand names must contain 2 dots and be surrounded by quotes.
 * The expression is compiled and the names of each operand are stored. The
 * values of an operand can be set later by name. Evaluation uses the last
 * known operand value.

fenriques:
indi_eval is the executable. Useful expression evaluation functions are in compiler.c am I right?
I'm getting lost on how use them inside the weather driver

knro:
yes, the ones in compiler.c. It would take a bit of work to see if the compiler can be adapted to evaluate expressions here... or maybe we're going too complicated here I don't know. What's a middle ground solution that does not bloat the driver, yet cover all cases?

PR #584, opened 2018-05-01
Fixed a protocol issue with ethernet connection to GTOCP4 using AP experimental 

ando--io:
When using the ethernet connection, the GTOCP4 is sending LF char at the end of its answer. this was not expected by the driver (it is even not documented by AP) and would eventually spoil the next request/reply with the mount as the LF char is left in the read buffer. 
I choose to apply a fix to indicom tty_read* methods, mimicing a previous correction done for gemini and ethernet connection. but this creates another layer of if/then to methods that should stay as generic and as simple as possible.


knro:
Right, the indicom is getting complex with special-cases. tcflush(fd, TCIFLUSH) does not clear it? All the LX200* commands call this after they read the response up to '#'. Or the above flush does not work for sockets?

@mikefulb Your thoughts?

mikefulb:
I don't use the Ethernet connection so I don't have experience with this 
issue.

I would recommend calling Howard @ Astro-Physics and bring up the issue 
- he has been helpful when I and Preston David have had protocol questions.

I think putting in a fix for undocumented behavior without feedback from 
Astro-Physics could be problematic going forwards if they change it.

Michael Fulbright

ando--io:
Better to get a confirmation from AP. Although the fix is not invasive, it would not do any arm if the LF char is not there.

knro:
Any response from AP?

mikefulb:
Yeah I'd be curious what AP said - again if its undocumented behavior we 
really should have clarification before committing to any change imho.

Michael

ando--io:
I've sent two mails to Howard at AP and no answer yet. if you know someone else able to answer...

knro:
I wonder if an alternative solution is to always set the byte after the stop char to '\0' explicitly?

ando--io:
I will check around your first suggestion, and understand why apparently tcflush does not clear the tcp reception buffer. this would be more elegant.

knro:
tcflush does not with sockets. The LF comes before the 0xA?

ando--io:
I have finally got a confirmation from Ray Gralak in the public mailing list that is char is intended. 

And I did not find any specific information about tcflush on socket.

 Now comes the question of the strategy to fix it. The solution I've proposed works but is not the best one, because it specialises this communication layer that should stay as neutral as possible.  
But the assumption that communication protocols are agnostics to their transport seems not always true. it is worth it to introduce a way for drivers to deal with this ?

mikefulb:
That Gemini stuff in indicom.c is not ideal at all but I guess the framework pushes you towards solving the AP problem in a similar way.  

I could see having a pre and post read/write hook you could register with indicom that could possibly handle these problems and maybe move this specific code from the general function.  I am not sure that is actually a better solution.

PR #586, opened 2018-05-04
Get camera temperature from EXIF data when using Canon DSLRs

k-ross:


k-ross:
Looks like libtiff5-dev isn't installed on the build server.

jochym:
But it seems that there is still other problem on travis. Please investigate it separately.

k-ross:
Looks like an unused parameter warning is being treated as an error. Will fix shortly.

jochym:
OK. Seems to be fixed.

k-ross:
Great!  Now you guys just need to decide if you want this change or not. :)

jochym:
That is a prerogative of our benevolent dictator. @knro what you think ? Looks interesting and fairly non-intrusive and would help with dark libraries for DSLR.

knro:
Great work, this is a long requested feature. Is this limited only to Canon? If yes, then TemperatureNP should only be defined for Canon as well.

k-ross:
Yes only Canon.  I will make that change to define in only for Canon

k-ross:
I made that change.  This change is limited to Canon, but other camera makers also supply the info, in their own proprietary tags.  Needs to be handled separately for each camera maker. Since I only have Canon to test with, this is what I have done.

PR #587, opened 2018-05-04
 V4L2 Driver: Add support for GRBG pixel layout for NexImage 5

james-lan:


knro:
Great work James! So do you think users would be able to 'plug n play' NexImage5 or are there still a lot of steps needed to be done to get this working? Can we write a udev rules to take care of these steps?

PR #594, opened 2018-05-13
Reviewed Weather Ranges Behaviour

fenriques:
Weather ranges are used to set the OK range of operation and trigger Warning and Danger signals on several weather parameters such as Rain, Temperature, Gust, Wind etc. System shuts down when in Danger range.

Behavior on master is:
- Set an OK range, eg Temp: -10, +30
- Set an Alert range
Issues that I saw were:
- Ok range had precedence over other ranges. I.E. if Temp Ok was -10, +30 and Alert was +25,+30 there was no alert!
- Maximum value of OK and Minumum of Alert has the same meaning. It is redundant. Or even could lead to error, like leaving a gap between Max OK and Min Alert,
- You didn't know (if not by trial and error) if Danger was outside of Ok range or Alert range.

To be fair this behavior has its own correct logic: Danger is always set from -infinite to + infinite, then you overlap a Alert range from, say  temp, from -20 to +40 and an OK range from -10 to +30. 
But it was not clear until I look at the source.

Proposed behavior: 
- Just set the Ok range, then
- Outside of Ok range is always Danger.
- Alert is automatically triggered at 85% of Ok range (both at low and high limit) without setting it explicitily. Eg if temp ok range is -10, +30, Alert is triggered at -4 and +24.

To me this is simpler and less prone to error.

Changes has been made to device classes that call AddParameter in order to comply to the new signature of the method. 
But the Ok range is kept from the previous version, the Alert range would be ignored. This should limit compatibility with previous version.

ferrante


knro:
The 85% is rather arbitrary, I think that should be another property. It could be different for each value and should be configurable. There are situations when "Alert" leads to taking action quite different from "Danger". This could be changed to:

void Weather::addParameter(std::string name, std::string label, double numMinOk, double numMaxOk, double alertWarn=0.85)

So it is set to 0.85 by default, but can be changed by either the driver itself, or by the user (as a property).

fenriques:
ok. it's in this commit. I set the warning as percentage of the range so 10% on a 0-20 range means a warning at 18.

I can't figure out how to load saved values for ranges. I mean when I set and save 20 as temp max OK value, then the new proposed value is default one, not the saved one.

knro:
So I added a simple weather simulator driver. But I found a big problem with your changes. Sometimes, you do not want warnings with the range, since you only need warnings for one end of the spectrum. Take wind for example, OK range 0 to 20 km/h. By default, 0 to 3 km/h would issue a warning, which is something clearly not desired.

For now I added a hack to check if the either the min or max value to be zero, then this warning is not displayed.

fenriques:
i was aware of that but thought it was a feature for e.g. temperature that
works in a range so needs upper and lower warnings.
moreover we are talking about ranges and not limits. ranges are upper and
lower bounded.

said that if you think it's best to have only the upper one I will fix it
immediately. it s just a matter of an if condition.

sorry for not having discussed before.

PR #596, opened 2018-05-16
Update ZWO ASI SDK to 0.7.0503

jpaana:
Changelog:
V0, 7, 0503 2018.5.3
Add support for ASI120MM-Mini
Fix issue with ASI1600: pattern in image,  no image under high USB traffic and 16bit 
Fix bug with ASI294:   hot pixels in image under ROI
Fix bug happens when being called from multiple thread
Add API ASIGetSDKVersion
Add BitDepth to ASI_CAMERA_INFO corresponding to sensor ADC bit depth


knro:
The filter wheel blob was not updated?

jpaana:
No, just camera.

PR #599, opened 2018-05-19
Adding OS X SBIG Support for older cameras to INDI

rlancaste:
This pull request should:

1.  Put the SBIG dynamic library binary file that SBIG installs into Frameworks into a dylib that can either be put into home-brew's usr/local/lib folder or into an application bundle's Frameworks folder to support applications such as OS X INDIServer and KStars.  This file was not modified, it's just treating the file in the same manner that it is treated on Linux and the way the other manufacturer's binary dynamic library files are placed for INDI.

2.  Put the SBIG camera firmware hex files into /usr/local/lib/indi/DriverSupport/sbig for homebrew or into an application bundle's /Contents/Resources/DriverSupport/sbig file for programs like INDIServer and KStars.  This is the same way Apogee, DSI, QHY, and GPhoto firmware files are currently placed for INDI on OS X.

3.  Add the ability for the INDI SBIG driver to load the firmware for numerous older SBIG cameras which were left behind by SBIG's changes to its OS X driver installer.  The firmware used to be loaded via a kext, but when apple made some changes, it broke the SBIG firmware loader.  This adds a firmware loader to the INDI SBIG Driver.  The firmware will only be loaded if the camera has not yet had its firmware loaded, and it will attempt to load firmware when the INDIServer is first started and when the camera is first attempted to be connected.

4.  Add minor changes that will allow the INDI SBIG driver to be built and to incorporate some fixes that Peter had made that allowed the INDI SBIG driver to work well on OS X.


knro:
Great work Rob!

1. #include <libusb-1.0/libusb.h> ... can this be made into <libusb.h> only?
2. Please enclose all the MacOS code with #ifdef __APPLE__ since it is not required on other systems

rlancaste:
No, I think the include has to be <libusb-1.0/libusb.h> otherwise it doesn't find the header file in the include directory since it is in a subfolder called libusb-1.0

rlancaste:
There, I enclosed the libusb stuff in an ifdef.  I don't think I can enclose the method  loadFirmwareOnOSXifNeeded() in an ifdef but I did enclose its contents.

polakovic:
What will happen, if SBig driver framework is already installed on Mac?

PR #600, opened 2018-05-20
fix pier side reporting for 10Micron mounts

d33psky:


d33psky:
fix is now tested and confirmed

PR #602, opened 2018-05-21
EQMod pulse limits and deferring status update during pulses

jpaana:
This change adds support for setting minimum pulse length and running short pulses synchronously without timer in EQMod.

As communication delay is quite big especially with Bluetooth serial connection, it's better to ignore very short pulses altogether rather than overcorrect with too long pulses. This delay seems to be around 50ms for Bluetooth connection and supposedly around 16ms for EQDIR cables and 32ms via handset in PC Direct mode (I have only tested BT version, other info from http://eq-mod.sourceforge.net/prerequisites.html)

Also added support for doing short pulses without timer by sleeping the remaining time after communication delay from starting the pulse. Timer is also now started before starting the pulse to remove that excess time from the delay. 

Last change is skipping reading scope status during a pulse as that causes major delay and shouldn't be that urgent. Ideally the status would be deferred until pulses are complete, but this simple version just skips it as it is retried after a second anyway.

The default values are on the conservative side (minimum pulse 10ms, minimum pulse timer 100ms) but previous behavior (except skipping status read) can be restored by setting both values to zero. I've personally run tests with values 50 and 150ms, but those are probably not suitable values for wired connections.

image

### DIFF
--- a/3rdparty/indi-eqmod/eqmod.cpp
+++ b/3rdparty/indi-eqmod/eqmod.cpp
@@ -162,6 +162,7 @@ EQMod::EQMod()
     gotoparams.completed = true;
     last_motion_ns       = -1;
     last_motion_ew       = -1;
+    pulseInProgress      = 0;
 
     DBG_SCOPE_STATUS = INDI::Logger::getInstance().addDebugLevel("Scope Status", "SCOPE");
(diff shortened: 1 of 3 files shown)


knro:
Beautiful work! I'll actually will give it a go now.

PR #607, opened 2018-05-29
Fix memory leak in indi_gphoto_ccd

k-ross:
The documenation for gp_file_get_data_and_size says the pointer returned is owned by libgphoto2, and will be destroyed when the file is destroyed.  This is only true if you used gp_file_new().  If instead you use gp_file_new_from_fd(), then the pointer returned is entirely owned by the caller, and must be freed. *sigh*


k-ross:
Looks like the Travis build failure is from knro's most recent commit, not from this change.

knro:
Oh... did you alert libgphoto2 folks about this?

PR #608, opened 2018-05-29
Fixing build error for indiduino and fishcamp

rlancaste:
This should fix a couple of build errors on OS X.


rlancaste:
The issue is that for some reason, the different architectures seem to disagree on whether uint64_t is an unsigned long or an unsigned long long

rlancaste:
The other issue is that my OS X computer did not error out when this OS X build did

rlancaste:
But finally got it!

Issue #556, opened 2018-04-06
Feature request: meridian limit setting for EQMOD


StarryBird:
Current version of the EQMOD only supports setting horizon limit and the meridian limit setting is not available. When I manually nudge the mount, the mount will not stop slewing even if it passes the meridian, which may damage the mount.  Is there any plan for implementing the feature in the future version? 

Thanks !

StarryBird:
@knro Are there any plans for this feature ?  I think it is important to have this feature to protect the mount.

knro:
I'm not exactly sure about the current behavior.

@geehalel What do you think?

geehalel:
The current behavior is to leave the motor run, be it while slewing or while tracking. I've added Horizon limits after a request some time ago, but I wonder if a driver is the right place to manage such things. At least it may be placed at the telescope level class as this is common to all telescope drivers. On my mind it would be better to leave those aspects to clients themselves.

knro:
I agree with Jean-Luc assessment, this belongs to the client.

StarryBird:
@knro How about implementing a meridian limit settings in mount control page in ekos for equatorial mount?

When the meridian limit is on and the mount has passed the meridian, stop the mount immediately, however, user can still slew the mount to the opposite direction at this time (user cannot slew the mount along with the tracking direction as the mount has passed the meridian, which is danger) .

How do you think of that ?

knro:
Thanks for the report. Please open this issue in the 3rdparty repo issues --> https://github.com/indilib/indi-3rdparty/issues

PR #612, opened 2018-06-03
Add meridian limits setting for EQMOD. Bugfix for slewing in EQMOD when scope is already in limit. 

StarryBird:
1. Add meridian limits setting for EQMOD.
2. Add limits checking before slewing scope in EQMOD to avoid the problem that the scope still can be moved for a little distance.

---------------------

This commit is mainly for issue https://github.com/indilib/indi/issues/556. Since no one implemented such feature so I added it. With some fixes for the problem that the scope would slew for a small amount of distance even if it was already in limits.

The meridian limit is implemented via the RA encoder. Since EQMOD for Windows uses that way so I just follow with it.

User can input the encoder step manually, or use the switch to get the current encoder values and set it automatically. A file saving and loading for encoder values are available.

Finally, the limit is enabled by default. If someone don't like to enable it by default, just turn it off in the config file.

selection_001

### DIFF
--- a/3rdparty/indi-eqmod/CMakeLists.txt
+++ b/3rdparty/indi-eqmod/CMakeLists.txt
@@ -59,7 +59,8 @@ if(WITH_ALIGN_GEEHALEL)
 endif(WITH_ALIGN_GEEHALEL)
 if(WITH_SCOPE_LIMITS)
   set(eqmod_CXX_SRCS ${eqmod_CXX_SRCS}
-    ${CMAKE_CURRENT_SOURCE_DIR}/scope-limits/scope-limits.cpp)
+    ${CMAKE_CURRENT_SOURCE_DIR}/scope-limits/scope-limits.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/scope-limits/meridian-limits.cpp)
 endif(WITH_SCOPE_LIMITS)
 
 IF (UNITY_BUILD)
@@ -89,4 +90,5 @@ if(WITH_ALIGN_GEEHALEL)
 endif(WITH_ALIGN_GEEHALEL)
 if(WITH_SCOPE_LIMITS)
   install( FILES  scope-limits/indi_eqmod_scope_limits_sk.xml DESTINATION ${INDI_DATA_DIR})
+  install( FILES  scope-limits/indi_eqmod_meridian_limits_sk.xml DESTINATION ${INDI_DATA_DIR})
 endif(WITH_SCOPE_LIMITS)
--- a/3rdparty/indi-eqmod/eqmod.cpp
+++ b/3rdparty/indi-eqmod/eqmod.cpp
@@ -3242,6 +3242,11 @@ double EQMod::GetDESlew()
 }
 
 #ifdef WITH_SCOPE_LIMITS
+bool EQMod::IsMountInNorthHemisphere()
+{
+    return (Hemisphere == NORTH);
+}
+
 bool EQMod::IsCurrentPosInLimit()     //return true if in limits.
 {
     if (meridian)
--- a/3rdparty/indi-eqmod/eqmod.cpp
+++ b/3rdparty/indi-eqmod/eqmod.cpp
@@ -3247,37 +3247,40 @@ bool EQMod::IsMountInNorthHemisphere()
     return (Hemisphere == NORTH);
 }
 
-bool EQMod::IsCurrentPosInLimit()     //return true if in limits.
+bool EQMod::IsCurrentPosInLimit()  //return true if in limits.
 {
-    if (meridian)
+    if (!meridian || !horizon)  //Error in limit checking module itself, return true to avoid dangerous situation.
+        return true;
+
+    if (meridian->IsLimitedSlewOn())
     {
         if (!(meridian->inLimits(mount->GetRAEncoder())))
             return true;
     }
 
-    double juliandate;
-    double lst;
-    struct ln_equ_posn steq;
-    struct ln_hrz_posn staz;
-    double current_az;
-    double current_alt;
-
-    juliandate = getJulianDate();
-    lst        = getLst(juliandate, getLongitude());
-
-    steq.ra  = (currentRA * 360.0) / 24.0;
-    steq.dec = currentDEC;
-
-    /* uses sidereal time, not local sidereal time */
-    /*ln_get_hrz_from_equ_sidereal_time(&lnradec, &lnobserver, lst, &lnaltaz);*/
-    ln_get_hrz_from_equ(&steq, &lnobserver, juliandate, &staz);
-
-    /* libnova measures azimuth from south towards west */
-    current_az  = range360(staz.az + 180);
-    current_alt = staz.alt;
-
-    if (horizon)
+    if (horizon->IsLimitedSlewOn())
     {
+        double juliandate;
+        double lst;
+        struct ln_equ_posn steq;
+        struct ln_hrz_posn staz;
+        double current_az;
+        double current_alt;
+
+        juliandate = getJulianDate();
+        lst        = getLst(juliandate, getLongitude());
+
+        steq.ra  = (currentRA * 360.0) / 24.0;
+        steq.dec = currentDEC;
+
+        /* uses sidereal time, not local sidereal time */
+        /*ln_get_hrz_from_equ_sidereal_time(&lnradec, &lnobserver, lst, &lnaltaz);*/
+        ln_get_hrz_from_equ(&steq, &lnobserver, juliandate, &staz);
+
+        /* libnova measures azimuth from south towards west */
+        current_az  = range360(staz.az + 180);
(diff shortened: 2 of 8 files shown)


StarryBird:
@TallFurryMan  Good question.  Currently user have to disable the limits temporarily. I am planning to add the ability to slew to the opposite direction if the scope is in limits.

knro:
I think such functionality should probably last with the client. Anyway, the last word is with @geehalel , what do you think?

TallFurryMan:
I tend to think this is a safeguard, so it may have its place in the driver at the same level as horizon limits or watchdog?Is there a way to manage trend, I mean, only prevent a move request that would worsen the situation, on both or single axis?

StarryBird:
@knro Though implementing the limits in client layer is OK, I haven't seen any meridian limit settings for equatorial mount in Ekos or any other clients. For Window version EQMOD, there's meridian limit setting for that. Since EQMOD driver for INDI has the ability to limit the horizon, I think there should be a meridian limit for that. Every time I slewed my mount, I must carefully check if it hit the tripod or in other dangerous position. I am planning to place my telescope remotely, I can't image that if the mount can slew without any limits since the chances to slew to a dangerous position by program (such as astrometry) is not so small.

TallFurryMan:
I second that, but only if the situation where slew requests are rejected can be rolled back (manual action in driver is a minimum to start with). About clients, there's an example in Ekos: when the scheduler sleeps waiting for the next job and does not park the mount, the mount is left to track (I will fix this issue soon, but that's a use case for your change). If for some reason the mount is already close to the tripod (it's possible to delay or disengage meridian flip), then it's possible that it will hit during that time. I think we can't trust the client for safety measures, especially when the client can override such safeties (not mentioning crashes without watchdog on the remote side). It's annoying when the telescope is remote, even more when the telescope is multi-user.

StarryBird:
@TallFurryMan  I will add the ability to slew the mount to the opposite direction if it was in limits so that the scope can restore from the limited state without turning off the limit manually.

StarryBird:
@knro Any idea with the failure of the continuous-integration ? I tested my code and there is no compile error. It looks like some package error occurred during the test.

selection_001

knro:
Hmmm I'm not sure, could be a transient error. @jochym any ideas?

StarryBird:
I just pushed a commit in which fix the slew limit switch cannot take effect in both horizon limit and meridian limit.

TallFurryMan:
I'll try to test this this week.

TallFurryMan:
I had a few relatively short tests with this PR:
- When meridian limits are not initialized, PR outputs a warning about mount being out of limits every second. Even after initializing, warning is output, which is confusing. When driver is disconnected, then reconnected, warning seems to stop.
- That warning is not entirely harmless: I had difficulties to achieve a slew in one step when the meridian limits were not set (or perhaps set, but both to 0). Strangely, I could still slew, but I had to issue the request multiple times. I was slewing from park position to Kocab/Phecab in UMi during that test.
- It's not clear which RA goes on west pier side and east pier side, but there's a message stating an error when choice is not compatible with hemisphere, so that's probably ok.
- RA is rounded when set, but I don't know the rule there. 6688791 got rounded to 6688790 and 10061565 got rounded to 10061600...
- I think "Abort tracking" should be on by default, but not "Abort slewing". It's quite difficult to make the mount slew on reversed pier side as it will automatically use the safest path (probably?). Blocking slew could lock the mount up I suppose, while tracking stopping is the feature avoiding the mount hitting the pier while capturing.

I'd like to continue testing tracking in additional situations, and to check the impact on capture.

TallFurryMan:
I did set meridian limits to [6688790,10061600], which is roughly +/-30 degrees east and west of meridian, with a different pier sides for RA. From a western goto location, I then slewed to 6160972, which is nearly at the meridian but on still on western side, and tracking stopped with "Aborting tracking". There seems to be something still wrong with pier side management, sorry.

TallFurryMan:
I think the non-initialized 0 value might be causing issues when adding RA values, or I might be doing things wrong.

To get the RA values in this test, I think I first disabled both aborts because they were originally set to zero and the mount would not slew (well, that's not entirely clear what happened by the way, I think it first agreed to move in DEC, but I'll retest). First I set the mount to slew at 30 degrees west of meridian, and set that value for east pier side. I then set the mount to slew to 30 degrees east of meridian, and set that value for west pier side. I then enabled tracking (which from memory had no effect when checked), then I set the mount to slew just west of the meridian, without flipping. 

I was expecting the feature to not block any goto request in that situation, as by default the setup would automatically choose the correct pier side. Opposite to this test, to engage the feature, I would need to set the mount to flip to the east side of the meridian, and leave it to track for two hours, then observe that tracking would stop at 30 degrees on the "wrong" pier side. Instead I observed that the feature disabled tracking immediately, but I'm not sure how immediate that was because of the looping warning message.

TallFurryMan:
There's something I don't understand: RA alone is not sufficient to determine pier side. You need to take DEC into account, don't you?

TallFurryMan:
I'm returning to this PR, as this night I went - well, my filter wheel went - hitting the tripod during a relatively long time because of the Ekos capture tab getting stuck waiting endlessly for a picture. I did not have this PR built in my INDI setup during that period.

We can't solve this issue with the built-in eqmod limits: the scope was looking at the zenith, but the mount did not process a meridian flip because the client was stuck. @knro, @geehalel, this is, indeed, a good example of a client causing damage to a remote installation.

So to me this PR is a good candidate for a useful safeguard. In the current state, as I understand the logic, the PR is causing the mount to stop tracking, and eventually slewing, when RA is outside a specific interval.

We know that what we do care about on an equatorial mount is to prevent the mount from tracking too long past the meridian when it started to track before the meridian. Thus, in the northern (southern) hemisphere, the useful safeguard is to enforce a limit to the RA encoder when the eqmod driver reports the scope is on the east (west) side of the pier.

@StarryBird could you simplify your PR to provide that only? Stop tracking when RA encoder exceeds the raMotor{East,West} limit depending on which hemisphere the scope is located in?

I don't think we need to worry for now about enforcing a limit when the eqmod driver reports the scope is on the "tracking-safe" side of the pier, as it won't be possible for a client to request an "inverted" goto and voluntarily cause the camera setup to hit the tripod.

StarryBird:
@TallFurryMan Sorry for my delay of the reply. I have been working on my own remote telescope control client development these days.

> There's something I don't understand: RA alone is not sufficient to determine pier side. You need to take DEC into account, don't you?

For meridian limit, the RA is sufficient to determine the pier side.  The idea for this PR is to limit the mount RA slewing/tracking to a range determined by 2 RA encoder values to avoid the mount hitting the tripod.  This is the same to the EQMOD on Windows.  Keep in mind that the meridian limit is only related to RA, so there's no need to take DEC into account. For DEC limit, it is already done by horizon limit.

> @StarryBird could you simplify your PR to provide that only? Stop tracking when RA encoder exceeds the raMotor{East,West} limit depending on which hemisphere the scope is located in?

I don't think we need to worry for now about enforcing a limit when the eqmod driver reports the scope is on the "tracking-safe" side of the pier, as it won't be possible for a client to request an "inverted" goto and voluntarily cause the camera setup to hit the tripod.

So you mean that only to stop the tracking when mount is outside of the limit ?

There are 2 ways to disable the slewing limit:

1. Edit config file `indi_eqmod_meridian_limits_sk.xml`, locate the `MERIDIAN_LIMITS_ON_LIMIT_SLEW` property and set the switch to OFF.  I am not recommend to to this as when you move your scope manually, you still have the chances to slew outside the limit and cause danger.

2. Slew your mount to the pier side EAST and stop when the hour angle is about MINUS 5 minutes, in meridian limit setting, hit the switch pier side EAST to record current RA encoder value.  Then, slew your mount to the pier side WEST and stop when the hour angle is about PLUS 5 minutes, hit the switch pier side WEST and save these 2 values.
You can get the hour angle value from Ekos mount control tab.

The second method allows your mount to slew slightly over meridian on both pier side east and west, so when you engaged the meridian flip in Ekos, the mount should flip normally.  I will check the meridian flip this weekend and give you the feedback.

Currently the mount can stop slewing or tracking according to the hemisphere where the scope is located.
Check this commit here: https://github.com/indilib/indi/pull/612/commits/26f13e0d2fb49116e9e54256f862ebc1cfda99d6

I am planning to test this PR further this weekend based on your feedback, some problems will be fixed soon.

TallFurryMan:
I think you have to reconsider your limits: if you don't take DEC into account when you limit RA on one side of the pier, you prevent the mount from tracking on the opposite side of the pier in the quadrant that is higher than the equator, that is good, but you also prevent the mount from tracking in the quadrant that is lower than the equator on that side of the pier, and that is not what you want. Divide the planisphere into four quadrants using the meridian and the equator to see what I mean.

knro:
Ok this PR does not belong in the INDI Core library repository. I think it needs to be rebased on the latest indi-eqmod in the 3rd party repo.

PR #613, opened 2018-06-04
Adding FFMV

rlancaste:


knro:
@jochym Any ideas why this keeps failing? The dependencies look about right, what are we missing?

knro:
You have to revert the padding changes, this breaks the driver.

rlancaste:
Yep I changed that back in the next commit when I realized the real problem

PR #614, opened 2018-06-06
Issue Sync without Position Reset for EQ mounts in SynScan driver

k-ross:
Frank Liu, R&D Engineer for Skywatcher, says to only issue a Sync command, and not to use the Position Reset command, when syncing.  I removed the position reset code for EQ mounts, but left it in for Alt/Az mounts, since it seems to be working, at least for the person (@kecsap) who put it in there in the first place. :)

The code prior to kecsap's recent fix would always send a position reset command, but it would send Alt/Az coordinates, even to an EQ mount.  This would really screw up EQ mount alignment.

The reason a lone Sync command appeared to not work before, is because it will only accept a Sync command if the offset is relatively small, within 6-7 degrees or so.  So you must already have done an alignment through the handset (a 1-star alignment would suffice), and only use the Sync command to "touch-up" the alignment.  You can't take a scope, power it on, point it to a random place in the sky, do a plate-solve, and sync.  That won't work.


kecsap:
Hi,

I think it is fine, but I think your explanation in the PR is very important. I would like to ask to copy paste the whole text as it is:
- to the commit message as additional lines
- and to the code as a comment inside the Sync function.

We would forget about these invaluable information over the years.

Csaba

k-ross:
I need to close and open a new PR.  I need to start using feature branches for PRs.

PR #615, opened 2018-06-06
Issue Sync without Position Reset for EQ mounts in SynScan driver

k-ross:


k-ross:
@kecsap I'm new to git and github.  This is now on a feature branch in my local repository, which is how github recommends I do things.  My local commits on master were messing with the PRs.

kecsap:
Yeap!

PR #616, opened 2018-06-06
Fix build with Qt 5.11

aleixpol:
qt5_use_mpdules cmake macro has been deprecated since Qt 5.0, just use
the targets as offered instead.


knro:
Great, thanks!!

PR #618, opened 2018-06-16
Fix Celestron pulse guiding commands

juanmb:


knro:
Thanks Juan! How about making more improvements like sleeping on very short pulses? or stopping other serial operations while the pulse is active? That is, anything to minimize interference with the timers.

knro:
I'll merge this for now, and the other improvements can come in another PR.

juanmb:
This was only a code fix. I'll work on a better implementation of pulse guiding based on your suggestions, but it won't be as quick as this one.

PR #620, opened 2018-06-20
Enable/disable EOS camera mirror lockup with custom function

tstibor:
Employ PTP property called CustomFuncEx to enable/disable mirror
lockup for EOS cameras. Before this commit mirror lockup works only,
when mirror lockup is enabled by hand via the EOS camera menu.
Details of the EOS customfuncex are provided at
https://github.com/gphoto/libgphoto2/blob/master/camlibs/ptp2/canon-eos-customfunc.txt

The CustomFuncEx can be tested with gphoto2 as follows:

enable mirror lockup:
gphoto2 --set-config customfuncex=20,1,3,14,1,60f,1,1

disable mirror lockup:
gphoto2 --set-config customfuncex=20,1,3,14,1,60f,1,0

### DIFF
--- a/3rdparty/indi-gphoto/gphoto_driver.cpp
+++ b/3rdparty/indi-gphoto/gphoto_driver.cpp
@@ -36,6 +36,10 @@
 #include <tiffio.h>
 #include <tiffio.hxx>
 
+#define EOS_CUSTOMFUNCEX                "customfuncex"
+#define EOS_MIRROR_LOCKUP_ENABLE        "20,1,3,14,1,60f,1,1"
+#define EOS_MIRROR_LOCKUP_DISABLE       "20,1,3,14,1,60f,1,0"
+
 static GPPortInfoList *portinfolist   = nullptr;
 static CameraAbilitiesList *abilities = nullptr;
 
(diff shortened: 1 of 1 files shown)


knro:
Great idea, looks good to me!

knro:
Can you please take a look at this report? http://indilib.org/forum/general/3620-indi-gphoto-ccd-with-dslr-t7i.html#27564

From the logs, I see that setting the custom function happen at 83.13 seconds mark, and then all subsequent exposures fail.

tstibor:
Yes indeed it looks like setting customfuncex: 20,1,3,14,1,60f,1,0 (disable mirror lockup)
causes the problem. A quick fix would be probably to #if 0 the two enable/disable settings:

```
#if 0
gphoto_set_widget_text(gphoto, gphoto->customfuncex_widget,                                                         
                               EOS_MIRROR_LOCKUP_ENABLE); 
gphoto_set_widget_text(gphoto, gphoto->customfuncex_widget,                                                     
                                   EOS_MIRROR_LOCKUP_DISABLE);
#endif 
```
Do you also have the problem with your DSLR @knro. I have a Canon 700D and it works without any problems. If this problem occurs only on some EOS cameras one could strcmp for the model and disable for them the mirror lockup feature.

PR #623, opened 2018-06-21
Fix GCC 8.X error: catching polymorphic type

tstibor:


knro:
Thanks, we've been getting lots of errors recently with gcc 8.x. Now the same error in Apogee :(

tstibor:
I send another PR for fixing these issues in apogee_ccd.cpp and qsi_ccd.cpp

PR #639, opened 2018-07-05
Adding the INDI Webcam Driver

rlancaste:


knro:
The cmake files need to go to the root cmake_modules directory so only FindFFMpeg.cmake needs to be added there.

jochym:
This is quite substantial patch. It is nice it builds on OSX. I'll try to make it build on linux without too large changes in the build system. I have some issues with extending dependencies - it will make building harder for users (and us!) on multiple platforms. I'll try to report/make progress here.

rlancaste:
Thank you very much for your help Pawel.  This was a challenging driver to make, it took about a week to get it all working.  Sorry about the dependencies, I don't want to add to them either, but  It was the only way I found to make webcams work in OS X since it supports avfoundation.

rlancaste:
One issue with deleting those register-all calls.  That is fine on OS X, but on Linux, when I removed those calls, it made the driver not find the codecs.  So I had to add them back.  It is possible that something was out of date on the system I was testing on, but those register calls might still be necessary.  It didn't seem to hurt to leave them in, except for complaining that they were depreciated.

rlancaste:
another bonus to using ffmpeg for Linux was that the actual code is shorter to provide most of the same functions as the v4l2 driver and it supports other cameras that v4l2 does not support

jochym:
Note that if your branch is named drv_webcam the ubuntu builds will build only the basic library and your driver, also on CI - much faster.

rlancaste:
Thank you for the info about building.  As I said, on my machine, those calls were not needed at all, it just worked.  It was when I was testing on a linux machine with xenial that I had trouble.  It probably means that the system was using an out of date ffmpeg.  I will try again on a system that is more up to date without those calls.

jochym:
Xenial is kind of oldish (2 years) - we are developing for the next release so it is better to go with the newer versions. Maybe you can call these functions conditionally or something? If it does not work in xenial we should better find a way to include this code. Maybe switch off the the compiler option for your module?

jochym:
We have a problem with driver for centos. The packages for centos are 2.6.x and the driver does not compile for them. Please wait for the image to build and look at the build - maybe it is possible to fix that. The xenial has 2.8.x and it compiles. Furthermore - I have checked and only fedora has 4.0 ffmpeg so I think you should disable that warning and add the calls back.

jochym:
@rlancaste, @knro : I did a little investigation and it seems there are multiple versions (2, 3, 4) and variants (libav, ffmpeg). I think that before merging we need to decide which variant (preferably both) and version is supported. If it is not possible to support all options we need to put some conditions into build infrastructure and just skip the driver builds on not supported platforms. I think we will encounter more cases like this with expanding set of 3rd party drivers. Indeed, I think we will eventually need to split the repo into core library and  core drivers and 3rd party drivers build against the library. I have opened an issue about it (https://github.com/indilib/indi/issues/646) - but this will not solve the problems here.

So, please @rlancaste :
- [ ]  review the requirements of the driver against the above list (2,3,4/libav,ffmpeg) and decide which you want to support
- [ ]  implement switches in the build to skip the driver if it cannot be supported on particular platform
- [ ]  I will adapt the CI images to include the dependencies and/or accommodate other required changes. I have already split the images build into separate repository so I can update them independently from the main repo.
- [ ]  We can merge !

knro:
Looks like it needs patching for CentOS?

PR #644, opened 2018-07-09
Add macOS support

sWski:
- add gxccd library v0.2.4 compiled for macOS 10.7+
- remove fitsio requirement as it isn't used


knro:
Mac users are going to be thrilled about this. Thanks Jakub!

PR #647, opened 2018-07-13
Nighscape 8300 Driver/Minor Optec IFW changes

dirkenstein:
This is the Nightscape 8300 driver that I commented on in the forum.
Needs libftdi1/libusb and D2XX for full support. New CMake module to find D2XX added.


dirkenstein:
Need to provide new pull request to make D2XX (3rd party binary-only lib) optional

PR #648, opened 2018-07-14
Nighscape 8300 Driver/Optec IFW supports TCP/IP

dirkenstein:
new pull request to replace yesterday's as this failed a number of compile options due to D2XX being mandatory.

New version will now build as long as at least libftdi1 and libusb1 are available.

Also removed serial support by default as it doesn't work (kernel ftdi_sio driver can't handle bulk data well).

### DIFF
--- a/3rdparty/CMakeLists.txt
+++ b/3rdparty/CMakeLists.txt
@@ -87,6 +87,7 @@ option(WITH_RADIOSIM "Install RadioSim Receiver" On)
 option(WITH_GPSNMEA "Install GPS NMEA Driver" On)
 option(WITH_ARMADILLO "Install Armadillo & Platypus Driver" On)
 option(WITH_FXLOAD "Install FX3 compatible fxload tool" Off)
+option(WITH_NIGHTSCAPE "Install Nightscape 8300 Driver" On)
 
 
 find_package(FFmpeg)
@@ -306,6 +307,10 @@ endif()
 if (WITH_WEBCAM)
 add_subdirectory(indi-webcam)
 endif()
+
+if (WITH_NIGHTSCAPE)
+add_subdirectory(indi-nightscape)
+endif(WITH_NIGHTSCAPE)
 
 # Check if libraries are found. If not, we must build them, install them, THEN run CMake again to build and instal the drivers. If all the libraraies are installed, then we build and install the drivers only now.
 if (LIBRARIES_FOUND)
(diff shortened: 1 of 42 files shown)


knro:
@jochym What's going on with the CI?

knro:
Can you please rebase to fix the conflict above? Also, I think we need to provide documentation for this over at indilib.org (e.g. http://indilib.org/devices/ccds/apogee-ccd.html)

knro:
Great work! The documentation + screenshots would be greatly appreciated!!

dirkenstein:
Is there a standard format for the docs at indilib.org?
I guess you want ekos screenshots. Do you have the source for the apogee doc  page so i can edit it appropriately.

I've added the standard text documents in the source dir and pushed them.

knro:
Do you have a user account on INDI Library website? I can give you permissions to create a new listing.

dirkenstein:
My forum account is dirkn.

knro:
Ok you have permission now. You can now edit this --> http://indilib.org/devices/ccds/celestron-nightscape-8300.html

dirkenstein:
Edits completed, hopefully sufficient for the purpose.

knro:
Wow that was fast! Thanks a lot! I just added an image. I'll try to prepare the 3rd party release today.

knro:
Ok I just pushed few improvements to the code. But I'm not sure what drop_root_privileges() is for? Is the driver required to start as root to set it as high priority (renice to -20) which would require root privileges?

dirkenstein:
Hi Jasem,

The drop_root_privileges is only needed if you run the driver setuid root. It's optional, and has no effect on the driver if run as a normal user. 

I don't provide instructions for doing it because it's inherently a bad idea, but it may be needed in two scenarios:

a) setting up your driver to have access to the USB devices without root needs some udev scripts on Linux. I can provide these if required so the usb device  for the nightscape is added to group 'usb'. You add your indi user to that group and it has access to the device.

b) Setting priority to -20 helps with the USB stability on my mac. It may not have as much affect for everyone. You can't renice to -20 without root privileges or some really complicated /etc/limits settings.

knro:
Excellent, do you have the VID:PID for the camera so I can add a udev rule at least for Linux users?

dirkenstein:
This rule works

SUBSYSTEMS=="usb", ATTRS{product}=="Nightscape CCD Camera", GROUP="usb"

dirkenstein:
VID and PID 👍 
  int vid = 0x19b4;
    int pid = 0x0065;

PR #667, opened 2018-08-01
bug fix for new serial buffer length

Blueshawk:
I hope I'm doing this right as I've never done it before. This looks like it should offer the most recent changes including the new 64bit buffers in lx200_onStep.cpp to be merged to indilib/master. Azwing would like to wait till we get docs and all but it's borked at the moment in stable. there are quite a few changes here. be SURE i'm doing the right thing before accepting it.


Blueshawk:
As usual my attempts lead to action by others. #668  is probably a better idea.


Trying is one sure way to fail. -Ray Wells 4.0

PR #695, opened 2018-09-01
Optec TCFS driver enhancements for your consideration

ken-self:
Primarily to implement the Mode A and B temperature compensation slopes and delay to complete full set of functions.
In the process I removed the dependency on the skeleton XML since I couldn't work out how it works.  Hope this is a good thing.
Fixed an issue with the Auto modes where the focuser does not accept any commands so the polling commands continually time out. Now when in Auto mode TimerHit reads the messages sent by the focuser
Altered FFMODE to retry 5 times whenever used (via SetManualMode())
Altered timeout to 2 seconds from 5 seconds. 5 seconds seems like an eternity

### DIFF
--- a/libindi/drivers/focuser/tcfs.cpp
+++ b/libindi/drivers/focuser/tcfs.cpp
@@ -80,6 +80,7 @@ void ISSnoopDevice(XMLEle *root)
 *****************************************************************/
 TCFS::TCFS()
 {
+    currentMode = MANUAL;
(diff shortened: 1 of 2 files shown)


knro:
This looks good to go. Thanks for the great work!

ken-self:
Unfortunately not as Focus Boss runs off a different driver - indi_lynx_focus rather than the indi_tcfs_focus driver that I modifed

AstroMuck:
Bummer, but thanks for letting me know!

PR #701, opened 2018-09-03
Rtlsdr

iliaplatone:


iliaplatone:
Jasem, I'm sorry but I must update both indi_rtlsdr and libdspau in these commits.

iliaplatone:
There is some problem when using FITS format: Spectrum FIT header is corrupted

PR #709, opened 2018-09-08
[indi-rtlsdr] fixed remaining time lock and sample format set to 16bi…

iliaplatone:
…ts unsigned


iliaplatone:
Here I removed unused blobs from the blob vector so no empty file are sent

PR #716, opened 2018-09-29
One more chunk size change in SX driver

polakovic:


knro:
So the 32MB on x86 didn't have make any difference?

PR #733, opened 2018-10-15
[dspau] Updated version to 2.0.0

iliaplatone:


knro:
@jochym Do we need to build on Trusty? Seems to be missing one package added by @iliaplatone

iliaplatone:
This problem is because within the dependencies was added a missing package: liblimesuite-dev wich does not exists in trusty, for this reason I build it from scratch inside the deps script

PR #735, opened 2018-10-16
[indi-sdr-hw] removed spectrum streams from limesdr and rtlsdr

iliaplatone:


knro:
What's the cause for this?

knro:
@iliaplatone can you please give more details on why spectrum streams are removed?

iliaplatone:
I removed them because of some issue in libDSPAU. I'm trying to find where is the bug, but I couldn't find it yet.

PR #747, opened 2018-10-29
changed UI button order to a more natural/MGEN way

chkettu:
MGEN-UI-buttons now ordered as on the original device:

[  ESC ].   [  SET ]
[  UP  ]
[ LEFT ]   [ RIGHT ]
[ DOWN ]


knro:
@TallFurryMan Please have a look.

TallFurryMan:
I agree this looks more like the MGen pad. Originally, I coded a 3 column per 4 row table with some inactive buttons. But the result had no added value, because the INDI panel has no defined UI, so I simplified it to basic. It's only a view on properties after all.

PR #750, opened 2018-10-31
fix some bug of telescope

yangzwoptical:
the modify of ieqprodriver.cpp is questionable, maybe split "guide rate" to "RA guide rate" and "Dec guide rate" is better.

### DIFF
--- a/libindi/drivers/telescope/ieqprodriver.cpp
+++ b/libindi/drivers/telescope/ieqprodriver.cpp
@@ -910,8 +910,8 @@ bool set_ieqpro_guide_rate(int fd, double rate)
     int nbytes_read    = 0;
     int nbytes_written = 0;
 
-    int num = rate * 1000;
-    snprintf(cmd, 16, ":RG%03d#", num);
+    int num = rate * 100;//zy2018 set RA and Dec with the same value
+    snprintf(cmd, 16, ":RG%02d%02d#", num, num);//zy2018
 
     DEBUGFDEVICE(ieqpro_device, INDI::Logger::DBG_DEBUG, "CMD <%s>", cmd);
 
@@ -997,7 +997,7 @@ bool get_ieqpro_guide_rate(int fd, double *rate)
 
         if (sscanf(response, "%d", &rate_num) > 0)
         {
-            *rate = rate_num / 1000.0;
+            *rate = (rate_num%100 + rate_num / 100)/2/100.0;//zy2018 
             tcflush(fd, TCIFLUSH);
             return true;
         }
--- a/libindi/drivers/telescope/lx200driver.cpp
+++ b/libindi/drivers/telescope/lx200driver.cpp
@@ -861,7 +861,7 @@ int setCalenderDate(int fd, int dd, int mm, int yy)
     DEBUGFDEVICE(lx200Name, DBG_SCOPE, "<%s>", __FUNCTION__);
     const struct timespec timeout = {0, 10000000L};
     char read_buffer[RB_MAX_LEN];
-    char dummy_buffer[RB_MAX_LEN];
+ //   char dummy_buffer[64];//zy2018
     int error_type;
     int nbytes_write = 0, nbytes_read = 0;
     yy = yy % 100;
@@ -875,10 +875,11 @@ int setCalenderDate(int fd, int dd, int mm, int yy)
     if ((error_type = tty_write_string(fd, read_buffer, &nbytes_write)) != TTY_OK)
         return error_type;
 
-    error_type = tty_nread_section(fd, read_buffer, RB_MAX_LEN, '#', LX200_TIMEOUT, &nbytes_read);
+ //   error_type = tty_read_section(fd, read_buffer, '#', LX200_TIMEOUT, &nbytes_read);//zy2018
     // Read the next section whih has 24 blanks and then a #
     // Can't just use the tcflush to clear the stream because it doesn't seem to work correctly on sockets
-    tty_nread_section(fd, dummy_buffer, RB_MAX_LEN, '#', LX200_TIMEOUT, &nbytes_read);
+//    tty_read_section(fd, dummy_buffer, '#', LX200_TIMEOUT, &nbytes_read);//zy2018
+	 error_type = tty_read(fd, read_buffer, 1, 3, &nbytes_read);//zy2018
 
     tcflush(fd, TCIFLUSH);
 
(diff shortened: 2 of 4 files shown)


libindi/drivers/telescope/lx200zeq25.cpp
```diff
@@ -1321,7 +1321,7 @@ int LX200ZEQ25::setZEQ25GuideRate(double rate)
         LOGF_DEBUG("RES (%s)", response);
 
         tcflush(PortFD, TCIFLUSH);
-        return true;
+        return 0;//zy2018
```
knro: should be return TTY_OK

libindi/drivers/telescope/lx200driver.cpp
```diff
@@ -875,10 +875,11 @@ int setCalenderDate(int fd, int dd, int mm, int yy)
     if ((error_type = tty_write_string(fd, read_buffer, &nbytes_write)) != TTY_OK)
         return error_type;
 
-    error_type = tty_nread_section(fd, read_buffer, RB_MAX_LEN, '#', LX200_TIMEOUT, &nbytes_read);
+ //   error_type = tty_read_section(fd, read_buffer, '#', LX200_TIMEOUT, &nbytes_read);//zy2018
     // Read the next section whih has 24 blanks and then a #
     // Can't just use the tcflush to clear the stream because it doesn't seem to work correctly on sockets
-    tty_nread_section(fd, dummy_buffer, RB_MAX_LEN, '#', LX200_TIMEOUT, &nbytes_read);
+//    tty_read_section(fd, dummy_buffer, '#', LX200_TIMEOUT, &nbytes_read);//zy2018
+	 error_type = tty_read(fd, read_buffer, 1, 3, &nbytes_read);//zy2018
```
knro: Please explain the reason for this.

libindi/drivers/telescope/ieqprodriver.cpp
```diff
@@ -910,8 +910,8 @@ bool set_ieqpro_guide_rate(int fd, double rate)
     int nbytes_read    = 0;
     int nbytes_written = 0;
 
-    int num = rate * 1000;
-    snprintf(cmd, 16, ":RG%03d#", num);
+    int num = rate * 100;//zy2018 set RA and Dec with the same value
```
knro: The documentation here: http://www.ioptron.com/v/ASCOM/RS-232_Command_Language2014_V2.0.pdf

specifies this format:

Command: “:RGnnn#”
Response: “1”
Sets the guiding rate n.nn * sidereal rate. nnn can only be entered in the range of [10, 90].

libindi/drivers/telescope/ieqprodriver.cpp
```diff
@@ -910,8 +910,8 @@ bool set_ieqpro_guide_rate(int fd, double rate)
...
+    snprintf(cmd, 16, ":RG%02d%02d#", num, num);//zy2018
```
knro: he documentation here: http://www.ioptron.com/v/ASCOM/RS-232_Command_Language2014_V2.0.pdf

specifies this format:

Command: “:RGnnn#”
Response: “1”
Sets the guiding rate n.nn * sidereal rate. nnn can only be entered in the range of [10, 90].

libindi/drivers/telescope/lx200driver.cpp
```diff
@@ -1283,7 +1284,8 @@ int Sync(int fd, char *matchedObject)
     if ((error_type = tty_write_string(fd, ":CM#", &nbytes_write)) != TTY_OK)
         return error_type;
 
-    error_type = tty_nread_section(fd, matchedObject, RB_MAX_LEN, '#', LX200_TIMEOUT, &nbytes_read);
+ //   error_type = tty_read_section(fd, matchedObject, '#', LX200_TIMEOUT, &nbytes_read);//zy2018
+ 	error_type = tty_read(fd,  matchedObject, 1, LX200_TIMEOUT, &nbytes_read);
```
knro: Explain the reason for this change as well.

libindi/drivers/telescope/lx200driver.cpp
yangzwoptical: because this command return "1", not "#":

 Command: “:SCYYMMDD#”
 Response: “1”
 Sets the current Local Date.

libindi/drivers/telescope/ieqprodriver.cpp
yangzwoptical: We have a CEM25 and it is tested, the true command is  ":RGaabb#", guide rate of RA is 0.aa, guide rate of Dec is 0.bb

libindi/drivers/telescope/lx200driver.cpp
yangzwoptical: Command: “:CM#”
Response: “1

knro: Where did you get this from? The Meade documentation here says otherwise:

https://www.meade.com/support/LX200CommandSet.pdf

I can test it today as well with Meade at home.

libindi/drivers/telescope/ieqprodriver.cpp
knro: Ok, there are two commands set. The older one is :RGAAA# and the new one is :RGAABB# I think it depends on the firmware.. can you post the firmware version for the CEM25? Is it 20171001 or earlier?

libindi/drivers/telescope/lx200driver.cpp
yangzwoptical: I refer to this: http://www.ioptron.com/v/ASCOM/RS-232_Command_Language2014_V2.0.pdf

knro: lx200driver.cpp is for Meade based scopes only. Where does iOptron make a call here?

libindi/drivers/telescope/ieqprodriver.cpp
yangzwoptical: I think no, the firmware version here:
HC:170518
RA:170106
Dec:170106
is it the reason that command is different?

libindi/drivers/telescope/lx200driver.cpp
yangzwoptical: I use indi_lx200zeq25 to connect iOptron zeq25

knro: Ok then lx200zeq25 needs to implement its Sync function.. that's the solution.

knro:
Ok I pushed the fixes myself to GIT. Please git pull and let me know if that resolves the issues you have seen.

PR #756, opened 2018-11-05
added _GNU_SOURCE, so it compiles on Centos 7.3

pkubanek:


pkubanek:
I made a change to supply the original CMAKE hardcoded version with -tgz suffix. Hope this will resolve the confusions.

libindi/CMakeLists.txt
```diff
@@ -83,7 +97,7 @@ OPTION (INDI_CALCULATE_MINMAX "Calculate and store image minimum and maximum val
 ###################################################################################################
 IF (INDI_FAST_BLOB)
 # Append ENCLEN attribute to outgoing BLOB elements to enable fast parsing by clients
-add_definitions(-DWITH_ENCLEN)
+add_definitions(-DWITH_ENCLE)
```
jpaana: This looks like an unwanted change as the code has #ifdef WITH_ENCLEN

PR #757, opened 2018-11-07
PegasusAstro Ultimate Powerbox driver added + updated to support also old devices.

polakovic:
The only change is 99-indigo_aux_upb.rules and driver version.


polakovic:
Jasem decided he doesn't want UPB driver here :(

Issue #669, opened 2018-08-02
Build error - need to link with libc6


gnthibault:
Dear all,

Thank you for developing such a nice piece of software, it has been incredibly useful for me during the past year.

I recently got a build error, and I am not sure wether it comes from my side or not:

CMakeFiles/indi_hid_test.dir/libs/indibase/hid_libusb.c.o: In function `get_usb_string':
/home/gnthibault/projects/indi/libindi/libs/indibase/hid_libusb.c:366: undefined reference to `libiconv_open'
/home/gnthibault/projects/indi/libindi/libs/indibase/hid_libusb.c:379: undefined reference to `libiconv'
/home/gnthibault/projects/indi/libindi/libs/indibase/hid_libusb.c:395: undefined reference to `libiconv_close'

It looks like indibase should be linked with a shared object that contains the symbols libiconv_open, libiconv, libiconv_close.
I do have iconv.h in /usr/include, this is why the build does not fails at compile time when processing libindi/libs/indibase/hid_libusb.c, but only at link time

On my machine, it looks like I cannot find those symbols in any libiconv shared object.
It looks like, on ubuntu, content of libiconv is now part of the libc: https://askubuntu.com/questions/898826/why-is-there-no-libiconv-package-in-ubuntu

If I am correct, I should change something in /libindi/CMakeLists.txt, in order to set ICONV_LIBRARIES to basic libc, but I don't know exactly what is the correct way to do that

Thank you in advance for your help

knro:
Which Ubuntu are you on?

gnthibault:
Ubuntu 16.04
I saw that there is a cmake_modules/FindIconv.cmake inside indi, but it does not look like it is used inside of a find_library call ?

knro:
It's only used for Cygwin. It's not required for regular Ubuntu. Are you trying to develop for INDI? If not, it might be easier just to use the INDI PPA: http://indilib.org/download/ubuntu.html

gnthibault:
I am not actively developing something inside indi right now, but I do some experiment from time to time with the source code. I am not in a hurry to get this problem sorted out.

gnthibault:
Thank you very much for accepting the PR !
I can now close the issue.

gnthibault:
I have no idea why I am the only one to experience this issue

```
make
[  0%] Linking CXX executable indi_hid_test
CMakeFiles/indi_hid_test.dir/libs/indibase/hid_libusb.c.o: In function `get_usb_string':
[...]

PR #761, opened 2018-11-12
Now Iconv is automagically found thanks to module finder. No need to …

gnthibault:
This PR follows a previous small PR here: https://github.com/indilib/indi/issues/669
I found that the most automatic and straightforward way to get indilib to build on my machine (ubuntu 16.04) was to run the FindModule Iconv with quiet option.

Now indilib project build perfectly on my system.


knro:
What problem was this trying exactly to solve? Please reopen against master if needed.

PR #766, opened 2018-11-20
Added fix for QHY5-M- CancelQHYCCDExposing* crash

downrange75:
Tried both CancelQHYCCDExposingAndReadout and CancelQHYCCDExposing from the SDK with the QHY5-M- (SSAG). In either case, this results in a crash of the indi driver. Added a condition to bypass this call for the QHY5-M-. The device now works using Ekos as a guide camera.

### DIFF
--- a/3rdparty/indi-qhy/qhy_ccd.cpp
+++ b/3rdparty/indi-qhy/qhy_ccd.cpp
@@ -929,17 +929,26 @@ bool QHYCCD::AbortExposure()
     }
     pthread_mutex_unlock(&condMutex);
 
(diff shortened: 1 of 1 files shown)


knro:
Does this apply to other cameras?

downrange75:
This change applies only to the QHY5-M-.

knro:
I know. I meant is the _problem_ applicable to other cameras as well that we should include in this?

downrange75:
That I do not know. I've only found these symptoms related to the QHY5-M- while looking for a solution to my issue.

PR #772, opened 2018-11-26
Fix setting limits to weather values

jpaana:
Refactoring of the weather class missed passing ISNewNumber data to the new weather interface so limits couldn't be changed. Also removes a couple of stray white space again (sorry my editor has OCD for white space :)

### DIFF
--- a/libindi/libs/indibase/indiweather.cpp
+++ b/libindi/libs/indibase/indiweather.cpp
@@ -35,7 +35,7 @@ namespace INDI
 {
 
 Weather::Weather() : WI(this)
-{    
+{
 }
 
 Weather::~Weather()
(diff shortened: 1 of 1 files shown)


knro:
Thanks a lot. My editor (Qt Creator) is quite the opposite and keep inserting these white spaces. It's quite annoying and not sure how to fix it. I have to check from git diff what changed and go back and fix it.

jpaana:
There seems to be an option in QtCreator's Options -> Text Editor -> Behavior -> Clean whitespace which might help, I don't use it that much myself though.

knro:
Already have this checked, now I checked also "Whole document" let us see if it makes a difference.

PR #777, opened 2018-12-10
Switch Travis linux distro to xenial

glowmouse:
The indilib/indi travis ci is currently failing when compiling code with std::regex. The cause appears to be the gcc version on trusty.  It uses 4.8.4 - std::regex isn't properly supported in gcc until 4.9.   There are discussions in a few places.  i.e.,  

https://stackoverflow.com/questions/8561850/compile-stdregex-iterator-with-gcc
https://github.com/RobotLocomotion/drake/issues/1649

I'm proposing switching the ubuntu distro that Travis uses to Xenial, which uses gcc 4.9+.  i.e.,

$ gcc --version
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.10) 5.4.0 20160609

Test results from my travis branch:  https://github.com/glowmouse/indi/commits/travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ compiler:
 
 language: cpp
 sudo: required
-dist: trusty
+dist: xenial
 osx_image: xcode8.3
 
 matrix:


knro:
Thanks! Seems same issue for CentOS.

jochym:
But I think we should support CentOS. This is a huge user base (Centos + RHEL) and it is also base for scientific linux and others. I vote against dropping centos support.

knro:
I'm not suggesting we drop centos, but is there a more current version with gcc 4.9+?

jochym:
I'll check. Maybe we can switch gcc on centos itself.
P.

glowmouse:
Fix Travis and spiral on circleci/build_centos?

PS - I was impressed by how easy the ci/circleci docker containers were to use.  Kudos.

jochym:
@knro : Do we still support trusty? If not We should indeed switch off trusty build on travis otherwise I will try to switch to 4.9 to trusty. I will work on this tomorrow.

knro:
@jochym As far as I'm concerned, we support 16.04+. But if you'd like to keep Trusty support, then sure as long as it can build.

jochym:
There seems to be a problem with trusty. I will work on this a little bit more but for now I have dropped trusty from travis, set it to xenial and moved trusty to circleci - where we have bigger control over the build env (docker). @glowmouse is there anything more in this PR to be merged?

glowmouse:
Nothing more from me.

jochym:
So this is incorporated in master - closing

PR #781, opened 2018-12-15
Try to switch trusty build to newer gcc

jochym:


jochym:
@knro Would you take a look at this? Something is fishy with the travis build on trusty. I can switch gcc on trusty to 4.9/5/6/7 but it does not build with the same error in SoapySDR. When I switch to xenial with gcc 5.4.0 it builds fine, as well as on centos with 4.9. I do not get it. Maybe you will find some clue.
Mor now the build on centos is fixed (updated docker) and I will try to add trusty and xenial to our docker library and check if it builds.

knro:
@iliaplatone Can you please check the issue with SoapySDR build failure?

jochym:
I have found a potential reason - interaction of clang (installed on travis), gcc-4.9 and our build system.
Maybe we need to look over the precise toolchain selection logic. But for now I am moving trusty to circle and setting travis to build xenial. We lose nothing (we have still two slots in second stage build on CCI and gain xenial build.

PR #785, opened 2018-12-18
Use long arg with printf %ld

nadvornik:
Without this patch eqmod on aarch64 prints confusing debug messages like this:
```
DEBUG<->5267645.183416 sec<---->: Init() : Motors not initialized -- read Init steps RAInit=8388608 DEInit = 281470690131968
```


nadvornik:
OK, changed.

jpaana:
I would have rather used %u as format as the values printed are uint32_t, not long.

PR #796, opened 2019-01-02
SSAG abort exposure fix

IvanBenak:
Fix for the Orion SSAG abort exposure issue.  Improves guiding reliability while switching targets or stopping guiding.

### DIFF
--- a/3rdparty/indi-ssag/ssagccd.cpp
+++ b/3rdparty/indi-ssag/ssagccd.cpp
@@ -41,6 +41,9 @@
 #define HEIGHT 1024
 
 static std::unique_ptr<SSAGCCD> camera(new SSAGCCD());
+/* Ivan Benak Start:  Fix SetTimer to abort */
+int T_Handle;
+/* Ivan Benak End:  Fix SetTimer to abort */
 
 void ISGetProperties(const char *dev)
 {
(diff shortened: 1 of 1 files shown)


knro:
We don't have many users using SSAG.. can you report on what works and what doesn't? what camera are you using exactly? i.e. exact model?

IvanBenak:
Sure, the camera has no model number printed on it, but the manual that came with it states it is a 52064 (Orion StarShoot AutoGuider)

The issue presented itself when cancelling a running exposure.  Ekos would still feel that an exposure was in progress after the cancel and would not permit another one to begin. As the SetTimer was still active, the TimerHit routine would be called and the image from the SSAG would pop up in the FITS viewer. With normal image acquisition, it was possible around then to get a new snapshot.  As for guiding or field solving, those two seemed to be confused as you couldn't start up a new guide/field solve without restarting the indi server and Ekos/KStars as the message would indicate that a capture was already in progress.  I was not able to use the automated processing as when a new target processing began, the guiding function would be terminated and could not be restarted on the next target. 

After making the changes, cancelling a image acquisition proceeds smoothly, with no image popping up after a set time frame and that all of the photo / field-solving / guiding functions are able to try again without issues/restarts.

knro:
The SSAG works otherwise fine?

IvanBenak:
Hello, yes, I haven't encountered any other issues with it so far.  It would appear to work just fine.

PR #797, opened 2019-01-02
Cleanup of encoder calculation in eqmod

nadvornik:
- EncoderFromDegree and EncoderFromHour made exactly inverse to EncoderToDegrees and EncoderToHours, these methods work with angle adjusted by pier side
- EncoderFromRA and EncoderFromDEC made inverse to EncodersToRADec, these methods work with plain eq. coords, pier side is passed as extra arg
- fixed inverted pier side in EncoderTarget
- do not modify currentPierSide in EncodersToRADec, return it in output arg instead
- EncodersTo... methods kept unchanged to make sure the overall behavior stays the same
- Added unit test


knro:
@geehalel Can you please check this change?

knro:
Thanks! I will need to test this in my observatory as well. Any changes to EQMod have to be thoroughly field tested. Did you check the pier side in different orientations in the sky?

geehalel:
Hi,
Unfortunately I have no time and mostly weather conditions to perform tests. It's been months since I made any observations. I've looked at the code but that does not really help. 
I would suggest to the INDI team to write a new driver from scratch, as Skywatcher has now published the [motor controller protocol](https://inter-static.skywatcher.com/downloads/skywatcher_motor_controller_command_set.pdf). I think that alignment may be omiitted as it is only really useful for visual/nomad obs and when you don't have astrometry solving. In all other cases you could just use standard syncs. Hence you would have a much simpler driver, and surely safer as it is a big concern for many people here.

knro:
Tested it tonight and all appears to be in working order, great work!

PR #804, opened 2019-01-07
Dspau

iliaplatone:


knro:
Looks like few merge errors? <<<<< is present in the CMake file

PR #807, opened 2019-01-09
Advertise client that the driver implement the FocuserInterface

pchev:
The driver LX200_OnStep now implement the FocuserInterface. 
FOCUSER_INTERFACE must be added to DRIVER_INTERFACE so the client know it can use this driver as a focuser.

See:
https://www.ap-i.net/mantis/view.php?id=2051


knro:
@azwing Please check since you are maintaining the driver.

azwing:
Thank you for the improvement

Issue #798, opened 2019-01-04
Enforce counter-weight up in eqmod


nadvornik:
Hi,

Is anybody using the "Enforce counter-weight up" property as implemented in #550 ?
I have found it is rather unstable and hard to use.
For example:
- the object is 1 sec east of meridian
- the client wants to avoid extra meridian flip so it requests goto with "Enforce counter-weight up" enabled
- the message is delayed by network latency, the object passes meridian meanwhile
- eqmod performs goto to the object which is now west of meridian, with "Enforce counter-weight up" enabled
- it results in the extra meridian flip that should have been avoided

Also, I have another use case, not mentioned in #550: I have clear view only to the west, so I can keep things simple and avoid meridian flips at all.

I'd suggest this:

Add new property "Target Pier Side" with values 
- "Auto" - current behavior, with counterweight down
- "West", "East". - enforce given pier side, abort goto if the target is outside of safe range

maybe also rename  "Enforce counter-weight up" to  "Allow counter-weight up" and use it to extend the safe range 1 hour across meridian. Alternatively this can be given as numeric property.

I can implement it if this is an acceptable change.

nadvornik:
ok, see
https://indilib.org/forum/mounts/4383-enforce-counter-weight-up-in-eqmod.html

TallFurryMan:
Also check #612, which attempts to implement the safe range part.

PR #814, opened 2019-01-09
WIP: Replace FORCECWUP with TARGETPIERSIDE in eqmod

nadvornik:
This is what I have so far for
https://github.com/indilib/indi/issues/798

I have found that pier side and limits checking on south hemisphere was broken - pier side should not depend on hemisphere.

I have not tested it much yet, I just made it pass the unit tests.

### DIFF
--- a/3rdparty/indi-eqmod/eqmod.cpp
+++ b/3rdparty/indi-eqmod/eqmod.cpp
@@ -38,6 +38,7 @@
 #include <memory>
 #include <cstring>
 #include <unistd.h>
+#include <assert.h>
 #include <indicom.h>
 
 #include <connectionplugins/connectiontcp.h>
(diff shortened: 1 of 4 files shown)


knro:
Great, thanks! I will try to give this a go tonight if the weather clears up. Anything specific to look for beside checking the pier side?

knro:
@geehalel I know you're busy, but if you have some comment on this, please let us know :-)

knro:
@nadvornik Can you please email me at mutlaqja@ikarustech.com ? Thanks.

geehalel:
Hi,
I remember a discussion some time ago with Patrick resulting in a patch to fix eqmod Side of Pier (see [commit](https://github.com/indilib/indi/commit/ce360adfa417574f6b609100216a6951bd75f0f5)). That was based on the following [paper](https://www.indilib.org/media/kunena/attachments/495/ASCOMSideOfPierBehaviour.pdf) from ASCOM. I also just looked into the Eqmod ASCOM releases notes and see that they have added an option:
`ASCOM setup - option to select pierside method (Pointing, Physical or None)`
I don't know what is currently used in clients (Ekos, PHD2, Linguider, ...) nor how, so you're surely more advised to propose changes than I am. All this seems to be a question of using the same definition for drivers, clients and... users. 
Your above idea to replace FORCECWUP with ALLOWCWUP/TARGETPIERSIDE is indeed more natural, you're right. Furthermore when using alignment you may not know the resulting pierside.
Thanks for correcting pier side and limits checkings in south hemisphere, even not defined, things may be wrong ;-).

jpaana:
Tested this last night, worked nicely.

PR #822, opened 2019-01-14
Dspau

iliaplatone:


knro:
submodules break launchpad builds for some reason last I checked. We have to check now if that's fixed otherwise it will break the whole INDI build.

iliaplatone:
Ok, I'll wait for this. I felt free to do this because of updating issues with dspau, maybe it's good to ensure, if this is possible, to not keep track of the latest commits of submodules, but only stable versions or tags.

knro:
Ok no confirmation so far. Can we do this as is without submodules? Please check #832

iliaplatone:
Yes I'll prepare a new PR asap

knro:
Ok, please open a new PR. Currently, 3rd party is broken until this is fixed :(

PR #825, opened 2019-01-15
Version 1.5 Alignment & cosmetic changes

azwing:


libindi/drivers/telescope/lx200_OnStep.cpp
```diff
@@ -38,7 +38,7 @@ LX200_OnStep::LX200_OnStep() : LX200Generic(), FI(this)
     currentCatalog    = LX200_STAR_C;
     currentSubCatalog = 0;
 
-    setVersion(1, 4);
+    setVersion(1, 5);
```
knro: Don't forget to change the version in drivers.xml

azwing:
Jasem,

Thank you for the tip. I changed it

PR #832, opened 2019-01-22
Fix compilation due to libdspau changes from dspau_t to double

james-lan:


james-lan:
Note the that the check that failed failed on is in the ASI camera focuser.

knro:
#822 should fix this.

PR #833, opened 2019-01-23
Optec Tcfs focuser

ken-self:
Two changes bundled together.
The main one allows changes to the focuser position whilst it is in Auto mode. Normally disallowed by TCFS this allows the client scheduler to adjust the offset on filter change without manual intervention. The driver puts the focuser in manual mode to change the offset then returns to auto mode.
The second change implements a feature of TCFS Commander which allows not just a temperature gradient but also a zero degree intercept so that the focuser position can be determined for a given ambient temperature (assuming complete linearity). This allows fully automatic operation which can be enabled/disabled by StartMode switch. Arguably this should be a client feature but should help adoption of INDI by TCFS Commander users.
Would you like me to move the driver to Third Party?


knro:
Hi Ken, thanks for the PR! Since this driver does not have external dependencies, there is no need to move it to 3rd party.

PR #835, opened 2019-01-26
Driver for ScopeDome dome controller

jpaana:
I've now used this driver for over a year actively so I guess it's starting to be ready for wider testing too :)  I wrote some documentation for it in http://s2.org/~jpaana/scopedome/ which also documents the prerequisites and other caveats. I didn't want to squash the commits so that there is some history left. Phew!


knro:
Thanks Jarno! Can we get the documentation uploaded to INDI? like here: https://indilib.org/devices/domes.html

jpaana:
There seems to be an error on MacOS, have to fix that.

PR #838, opened 2019-01-28
Fixing Libdspau build issue in craft

rlancaste:


rlancaste:
I will wait to see what updates libdspau will get in the next couple of days, and then add another pull request for craft and libdspau if needed.

PR #843, opened 2019-02-01
Proposal to add libdsp into indidriver

iliaplatone:


knro:
So what's the purpose of integrating dsp into indidriver?

iliaplatone:
This library can be useful for convolution in ccds, matrix transformations for mounts and filtering and spectrums in detectors.
I'd propose this library to keep inside this one some functions dedicated in this role.

knro:
Ok great, I can see how useful it can be for some applications. Thanks!

PR #855, opened 2019-02-10
[Avalon StarGO] Enabling/disabling focuser

sterne-jaeger:
Querying the focuser status without a focuser connected leads to problems with StarGO when running several hours. It might lead that the StarGO device no longer responds and requires a reset.

With this version, the focuser can be activated and deactivated avoiding unnecessary (and potentially troublesome) querying to the AUX device.

### DIFF
--- a/debian/indi-avalon/changelog
+++ b/debian/indi-avalon/changelog
@@ -1,46 +1,11 @@
-indi-avalon (1.3) bionic; urgency=medium
+indi-avalon (1.3) stretch; urgency=medium
 
-Version 1.3 - 2019-02-08 (bug fix release)
-        + control added to activate/deactivate focuser on AUX1
-	  focuser status read only if focuser is activated
-	+ Doubled parking status light removed
+  * Enabling/disabling focuser.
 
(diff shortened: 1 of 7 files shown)


knro:
Great! Maybe also change debian/indi-avalon/changelog  as well to sync the package version?

debian/indi-avalon/changelog
```diff
@@ -1,5 +1,46 @@
-indi-avalon (1.0) bionic; urgency=medium
+indi-avalon (1.3) bionic; urgency=medium
```
knro: This is invalid format... just duplicate my first entry tha

knro: This is an invalid changelog .. the debian one is used to build packages and have very specific format. Please duplicate my first entry and modify it. or use **dch** to append a new changelog. We have a problem just yesterday with indi-starbook failing to build for a couple of weeks and it was traced to a _single space_ that wasn't added!

sterne-jaeger:
Hope it's correct now. How can I test it?

knro:
This is fine now, we'll know soon if there are issues with the debian builds!

PR #858, opened 2019-02-17
Dsp documentation

iliaplatone:


knro:
Great! Do you have more commits pending or should I merge this?

iliaplatone:
I guess you can merge as is right now.
Please take a look into defaultdevice: I added two methods that use the DSP API, I put them there to get them usable by any driver, I hope these are not too slow on embedded devices. Also the base type is double so in vfp machines there should be some problem.
An altrnative should be putting them into indiclient, let me know what do you think.

iliaplatone:
... ops. sorry, wait, there's an issue on mac

knro:
Ok, this is a lot more than documentation! I just checked all the changes:

1. DefaultDevice should not have any DSP related functions. Why can't these functions move to indidetector.cpp ?
2. astro.c and align.c can be merged in indicom

iliaplatone:
The DSP functions can be used on CCD, for example to do some kind of sharpen or wavelets. Since is bothering reproduce these methods each time I put them there.
Is it possible to add a specific class for DSP functions to be inherited by single drivers?

knro:
isn't the dsp lib suppose to provide such functions then?

iliaplatone:
Yes but sometimes is tedious rewrite those functions.
I'll move them into Detector.

libindi/libs/indibase/basedevice.h
```diff
@@ -20,6 +20,7 @@
 
 #include "indibase.h"
 #include "indiproperty.h"
+#include "dsp.h"
```
knro: This shouldn't be in the header as it breaks all 3rd party drivers not using DSP.

libindi/libs/indicom.c
```diff
@@ -112,7 +113,7 @@ int extractISOTime(const char *timestr, struct ln_date *iso_date)
 
 /* sprint the variable a in sexagesimal format into out[].
  * w is the number of spaces for the whole part.
- * fracbase is the number of pieces a whole is to broken into; valid options:
+ * fracbase is the number of M_PIeces a whole is to broken into; valid options:
```
knro: pi --> M_PI mass replace?

libindi/libs/indicom.c
```diff
@@ -379,7 +380,7 @@ int tty_timeout(int fd, int timeout)
     /* Return -1 due to an error */
     else if (retval == -1)
         return TTY_SELECT_ERROR;
-    /* Return -2 if time expires before anything interesting happens */
+    /* Return -2 if time exM_PIres before anything interesting happens */
```
knro: Same here

libindi/libs/indicom.c
```diff
@@ -721,7 +722,7 @@ int tty_connect(const char *device, int bit_rate, int word_size, int parity, int
     }
 
     // Set raw input (non-canonical) mode, with reads blocking until either a single character
-    // has been received or a one second timeout expires.
+    // has been received or a one second timeout exM_PIres.
```
knro: same

iliaplatone:
-Werror,-Wsign-compare-Werror=maybe-uninitialized causes other parts not to build, is it an error of mine?

knro:
No there are my errors, I'll fix them. I'll merge you PR now, thanks!!

PR #859, opened 2019-02-17
fix for focuser to accept negative values

azwing:


knro:
Negative values in what sense? Focuser position value in INDI are always positive with 0 being the fully retracted position.

azwing:
OnStep firmware allows configuration where Focuser travel is between -xx to +xx and 0 (zero) is half travel.
Personnally I use a traditional coarse 0 .. +xx but many user have the above, this is the reason for the change

knro:
I recommend against this as any negative values would break the INDI focuser interface. I just updated the documentation to explicitly highlight this. If the focuser supports SYNC then it would be easy to reset the position and not rely on negative value.

For example, support the users wants to go inward 5000 steps, but the current position is 2000 steps. If the minimum is 0, then then they cannot go back 5000 steps. So if you SYNC the current position to 5000, then set the GOTO target to 0, you end up with the same result.

azwing:
I agree.
On the other hand it is really easy to change the focuser set-up in the firmware so that focuser always is positive, and consequently I will close the pull request leaving the situation as is.

Thank you

james-lan:
It actually doesn't break the Indi focuser interface, actually my own checks were causing issues, this simply removes them. Ekos also supports this just fine. I've been using this code for a few weeks without issue. 

I'm confused as to why you think it will break it?

knro:
Not Ekos, INDI Focuser Interface as it defines the minimum absolute position to be zero. Check here: https://www.indilib.org/api/classINDI_1_1FocuserInterface.html

james-lan:
While that may be the stated method, and indeed it does use unint, it works, because everything already treats it as int32, not uint32.  From a quick scan of the driver with focuser interface, OnStep is not the only one to treat it that way. 

I would ask that this actually be put through: If you look here, the only thing removed are extra checks. So there is no harm in this going in. It never actually explicitly deals with negative values. 

There is harm in that OnStep does not support sync to an arbitrary location. It only supports 2 focuser input values commands: 0 and Mid. Which if you use +/- X, then they will initialize to the same (0) point. In which case, you can only go one way. Position is also stored in firmware. 

I'll open an issue on the FocuserInterface, because I believe it is actually not in line with the INDI specification, and should be changed from uint32_t to int32_t, but I'll make that argument there rather than here though, after gathering data on the drivers and their use of int32 and uint32, to make sure that nothing breaks. 

__In any case:__ The checks removed here do not work for their intended job, and negative values aside (which it happens to fix) shouldn't be there.

PR #860, opened 2019-02-17
indi-duino fixes

nadvornik:
Here are some indi-duino fixes.


knro:
Thanks for the PR!

A side question, do you think we need to have a proper documentation for how can new people get started with this? The documentation on the website could perhaps use some update?

nadvornik:
IMHO the existing documentation in README is OK to get started.

Biggest problems I had was with certain pins not working for no obvious reason, but that is hopefully fixed now.

PR #868, opened 2019-02-24
Indiserver fixes

nadvornik:
1. Do not use file descriptors from inactive DvrInfo. It crashes when the same fd is re-used. There already is the same handling for clients.

2. Indiserver allocates 50KB for each message, but for the -m option it considers real message size - typically under 0.5KB. When the queue size reaches the default 128MB limit, it allocates more than 12GB of memory, so it is usually killed by OOM killer much earlier.
The patch reduces the buffer to 2KB per message and changes size calculation to report all the allocated memory.

### DIFF
--- a/libindi/indiserver.c
+++ b/libindi/indiserver.c
@@ -77,6 +77,7 @@
 #define MAXSBUF       512
 #define MAXRBUF       49152 /* max read buffering here */
 #define MAXWSIZ       49152 /* max bytes/write */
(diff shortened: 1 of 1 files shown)


knro:
Thanks! Looks like the 50kb issue was a carry over from the change to base64 enhancements and the contributor forgot about it. Thanks for catching this and the other issues, great work!

PR #869, opened 2019-02-26
Celestron work

Chris-Rowland:
These changes implement a working version of the Celestron focuser driver.
Changes are to CMake, celestron.cpp.h and new files celestronauxpacket.cpp.h.


Chris-Rowland:
celestron fcuser implementation in celestron_work.

PR #875, opened 2019-03-03
Minor fixes to Celestron SCT Focus

Chris-Rowland:
The celestron packet data is now shown in the logs as a series of hex values instead of as a string.
The LOG_INFO message in initProperties has been removed.  Not sure why but it came up as a message box, not in the logs.


knro:
Great! Do you think this can be included in the regular celestron mount driver as well?

PR #877, opened 2019-03-06
Add focuser to Celestrob GPS

Chris-Rowland:
These changes add the Celestron focuser to the Celestron GPS driver.
I'm not that solid on the detail of INDI so they are worth a careful look.

### DIFF
--- a/libindi/drivers/telescope/celestrongps.cpp
+++ b/libindi/drivers/telescope/celestrongps.cpp
@@ -174,6 +174,24 @@ bool CelestronGPS::initProperties()
     FocusAbsPosN[0].value = 0;
     FocusAbsPosN[0].step  = 1000;
 
+    // Maximum Position Settings, will be read from the hardware
+    FocusMaxPosN[0].max   = 60000;
+    FocusMaxPosN[0].min   = 1000;
+    FocusMaxPosN[0].value = 60000;
+    FocusMaxPosNP.p = IP_RO;
+
+    // Focuser backlash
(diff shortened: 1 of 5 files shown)


libindi/drivers/telescope/celestrongps.cpp
```diff
@@ -351,11 +386,32 @@ bool CelestronGPS::updateProperties()
         // Sometimes users start their mount when it is NOT yet aligned and then try to proceed to use it
         // So we check issue and issue error if not aligned.
         checkAlignment();
+
+        //  handle the focuser
+        if (fwInfo.hasFocuser)
+        {
+            LOG_INFO("update focuser properties");
```
knro: This is probably not required.

knro:
Great job Chris! Looks like your issues with GIT are a thing of the past now. There are a few issues reported by CI, please take a look at this. You can still make changes to this PR and it will be reflected after each push.

libindi/drivers/telescope/celestrongps.cpp
Chris-Rowland: That code is intended to only show the focuser tab if a focuser exists, it seems to work but is there a better way?

knro: I meant just the part about LOG_INFO(...), probably the user does not need to see this message. You can make it LOG_DEBUG(...) for example or just comment it.

Chris-Rowland: I tend to be a bit too generous with log messages, I'll make it debug for the next build but won't commit specially.

PR #880, opened 2019-03-08
Fix problem with limits and position of focuser

Chris-Rowland:
The data is read from the hardware as a char array and some compilers
treat this as signed, giving bad values. Cast to uint8_t to prevent this
- hopefully.
We may not be able to tell until users try it.

### DIFF
--- a/libindi/drivers/telescope/celestrondriver.cpp
+++ b/libindi/drivers/telescope/celestrondriver.cpp
@@ -324,7 +324,7 @@ bool CelestronDriver::get_version(char *version, int size)
     if (!send_command("V", 1, response, 3, true, false))
         return false;
 
-    snprintf(version, size, "%d.%02d", (uint8_t)response[0], (uint8_t)response[1]);
+    snprintf(version, size, "%d.%02d", static_cast<uint8_t>(response[0]), static_cast<uint8_t>(response[1]));
 
     LOGF_INFO("Controller version: %s", version);
     return true;
@@ -338,7 +338,7 @@ bool CelestronDriver::get_variant(char *variant)
     if (!send_command("v", 1, response, 2, true, false))
         return false;
 
-    *variant = (uint8_t)response[0];
+    *variant = static_cast<uint8_t>(response[0]);
     return true;
 }
 
@@ -378,7 +378,7 @@ bool CelestronDriver::get_model(char *model, int size, bool *isGem)
     if (!send_command("m", 1, response, 2, true, false))
         return false;
 
-    int m = (uint8_t)response[0];
+    int m = static_cast<uint8_t>(response[0]);
 
     if (models.find(m) != models.end())
     {
@@ -419,10 +419,10 @@ bool CelestronDriver::get_dev_firmware(int dev, char *version, int size)
 
     switch (rlen) {
     case 2:
-        snprintf(version, size, "%01d.0", (uint8_t)response[0]);
+        snprintf(version, size, "%01d.0", static_cast<uint8_t>(response[0]));
         break;
     case 3:
-        snprintf(version, size, "%d.%02d", (uint8_t)response[0], (uint8_t)response[1]);
+        snprintf(version, size, "%d.%02d", static_cast<uint8_t>(response[0]), static_cast<uint8_t>(response[1]));
         break;
     default:
         return false;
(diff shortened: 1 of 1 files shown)


libindi/drivers/telescope/celestrondriver.cpp
```diff
@@ -185,15 +185,19 @@ int CelestronDriver::send_command(const char *cmd, int cmd_len, char *resp,
...
     {
-        LOGF_ERROR("Received %d bytes, expected %d.", nbytes, resp_len);
+        hex_dump(hexbuf, resp, nbytes > resp_len ? nbytes : resp_len);
+        LOGF_ERROR("Received %d bytes, expected %d <%s>", nbytes, resp_len, hexbuf);
         return 0;
     }
 
+    if (resp_len == 0)
+    {
+        LOG_DEBUG("resp_len 0, no response expected");
```
knro: wouldn't this spam the debug log? many commands have resp_len = 0

libindi/drivers/telescope/celestrondriver.cpp
```diff
@@ -319,7 +324,7 @@ bool CelestronDriver::get_version(char *version, int size)
     if (!send_command("V", 1, response, 3, true, false))
         return false;
 
-    snprintf(version, size, "%d.%02d", response[0], response[1]);
+    snprintf(version, size, "%d.%02d", (uint8_t)response[0], (uint8_t)response[1]);
```
knro: Please use newer cast: static_cast<uint8_t>(response[0])...etc

libindi/drivers/telescope/celestrondriver.cpp
```diff
@@ -761,7 +766,7 @@ bool CelestronDriver::set_track_mode(CELESTRON_TRACK_MODE mode)
 
 bool CelestronDriver::hibernate()
 {
-    return send_command("x#", 2, response, 0, true, true);
+    return send_command("x#", 2, response, 1, true, true);
```
knro: Can you elaborate on this? hibernate has a response? this didn't work before? or was the response ignored?

libindi/drivers/telescope/celestrondriver.cpp
```diff
@@ -857,8 +862,16 @@ bool CelestronDriver::foc_limits(int * low, int * high)
     if (rlen < 8)
         return false;
 
-    *low = (response[0] << 24) + (response[1] << 16) + (response[2] << 8) + response[3];
-    *high = (response[4] << 24) + (response[5] << 16) + (response[6] << 8) + response[7];
+    *low = ((uint8_t)response[0] << 24) + ((uint8_t)response[1] << 16) + ((uint8_t)response[2] << 8) + (uint8_t)response[3];
```
knro: how does the cast here affect the results? it is shifting bits, does it matter what the type is? bits are bits. Am I missing something?

Chris-Rowland:
I don't think so, commands with no reply data still respond with a # character.
Chris
sent from mobile so terse.
On 8 Mar 2019 10:12, Jasem Mutlaq <notifications@github.com> wrote:@knro requested changes on this pull request.



In libindi/drivers/telescope/celestrondriver.cpp:
>          return 0;
     }
 
+    if (resp_len == 0)
+    {
+        LOG_DEBUG("resp_len 0, no response expected");

wouldn't this spam the debug log? many commands have resp_len = 0



In libindi/drivers/telescope/celestrondriver.cpp:
> @@ -319,7 +324,7 @@ bool CelestronDriver::get_version(char *version, int size)
     if (!send_command("V", 1, response, 3, true, false))
         return false;
 
-    snprintf(version, size, "%d.%02d", response[0], response[1]);
+    snprintf(version, size, "%d.%02d", (uint8_t)response[0], (uint8_t)response[1]);

Please use newer cast: static_cast<uint8_t>(response[0])...etc



In libindi/drivers/telescope/celestrondriver.cpp:
> @@ -761,7 +766,7 @@ bool CelestronDriver::set_track_mode(CELESTRON_TRACK_MODE mode)
 
 bool CelestronDriver::hibernate()
 {
-    return send_command("x#", 2, response, 0, true, true);
+    return send_command("x#", 2, response, 1, true, true);

Can you elaborate on this? hibernate has a response? this didn't work before? or was the response ignored?



In libindi/drivers/telescope/celestrondriver.cpp:
> @@ -857,8 +862,16 @@ bool CelestronDriver::foc_limits(int * low, int * high)
     if (rlen < 8)
         return false;
 
-    *low = (response[0] << 24) + (response[1] << 16) + (response[2] << 8) + response[3];
-    *high = (response[4] << 24) + (response[5] << 16) + (response[6] << 8) + response[7];
+    *low = ((uint8_t)response[0] << 24) + ((uint8_t)response[1] << 16) + ((uint8_t)response[2] << 8) + (uint8_t)response[3];

how does the cast here affect the results? it is shifting bits, does it matter what the type is? bits are bits. Am I missing something?

libindi/drivers/telescope/celestrondriver.cpp
Chris-Rowland: I don't think so, the commands with no reply still return a # character.

libindi/drivers/telescope/celestrondriver.cpp
Chris-Rowland: Yes, hibernate returns a #.  I have a log showing this somewhere.

libindi/drivers/telescope/celestrondriver.cpp
Chris-Rowland: This is where the problem is, some compilers convert the signed char to a signed integer with the same value, then do the shift.
For example 0x80 is converted to 0xFFFFFF80, then an 8 bit shift gives 0xFFFF8000.

What's wanted is that 0x80 is shifted to 0x8000, this is what the RasbPi compiler does, which is why I didn't see it.
penguin's log shows this, it's OK with the Celestron SCT because the buffer uses unsigned char aka uint8_t but Celestron GPS uses a char array and it goes wrong for him on that one only. He's connected twice, using both celestron SCT and Celestron GPS.

I'm going to look at changing the type cast as you suggested.

libindi/drivers/telescope/celestrondriver.cpp
Chris-Rowland: done

PR #884, opened 2019-03-11
Experimental code to do a last align

Chris-Rowland:
This is really for the CGX and CGX-L mounts only because they have
switches that define the index position but it will work with other GEM
mounts if the user comments out the check in updateProperties, line 398
of celestrongps.cpp.

Seems OK on my AVX.


Chris-Rowland:
Don't know why the test has failed, from what I can see the response of "#" in set_sim_response in hibernate() should be correct.  This pattern seems to be used in a number of other places.

Chris-Rowland:
I simply follow what git requires by rote and it seemed to require that this was removed. I've never done so, neither remotely or locally.

knro:
it's restored now. It's my bad I merged this without noticing that.. thanks for finding about this @jpaana !

@Chris-Rowland when you run git status, you can see if a file is reported as deleted, and you can do a git checkout to restore it if it was deleted by mistake.

PR #888, opened 2019-03-16
Synscan: invert slew direction for AltAz mounts

artix75:
Slew directions are inverted on my Synscan GOTO Dobson mount (Orion SkyQuest XT8g) using the new driver (it doesn't happen with the Synscan legacy driver), so that the "up" ("north") button moves the tube downwards and, viceversa, the "down" button moves the tube upwards.
The same happens with the "left" button which moves the tube to the right and viceversa.

So, in my mount (and I think also in other AltAz mounts), the negative direction (37) means "North" for the vertical (altitude) axis and "West" for the horizontal (azimuth) axis.

Anyway, I don't know if this is a general issue for all other AltAz mounts, as I can only test it with my own mount.

So, I think it's worth to better investigate this behaviour in order to understand if it's a general behaviour for all AltAz mounts (I guess it is).


knro:
Thanks, looks fine for Alt-Az mounts.

PR #889, opened 2019-03-16
Extended telescope joystick motion control

tcrass:
Dear maintainers,

when I tried to use an XBox 360 wireless controller for slew control, I realized that

- the only joystick which got mapped correctly (as JOYSTICK_1) was the left analog joystick;
- the other joysticks (JOYSTICK_2 and JOYSTICK_3) where partly mapped to the right analog joystick's axes, partly to the two triggers;
- the directional pad wasn't mapped to any JOYSTICK_* at all.

To this end, I implemented a mechanism which allows for explicitly defining which joystick axes are to be used for controlling N/S and W/E motion, respectively.

During the course of this work, I also fixed two general bugs regarding joystick axis handling.

Finally, I modified the code responsible for snapping a joystick's angle to x/y direction. I have the impression that this greatly facilitates controlling the RA and DEC axes separately.

Hope you like the changes I made... 

Cheers --
    Torsten

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ libindi/make_tcfs_symlink.cmake
 *.cbp
 install_manifest.txt
 .idea
+*.kdev4
 libindi/examples/tutorial_eight/tutorial_eight
 libindi/examples/tutorial_five/tutorial_dome
 libindi/examples/tutorial_five/tutorial_rain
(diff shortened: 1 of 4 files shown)


knro:
How would this affect users who use the current system with simpler pads?

tcrass:
Jasem,

> How would this affect users who use the current system with simpler pads?

not at all, I reckon -- I haven't removed any settings, just added 
things, and the default behaviour for slew control is still to use the 
existing 4-way joystick mode.

Cheers --

	Torsten

PR #893, opened 2019-03-22
Starbook 0.5

not7cd:
This update should cover all RPC commands in the command interface class.
There will be increased usage of exceptions to handle unknown states.


knro:
Can you please check the CI errors?

knro:
Would this work also with Vixen Starbook Ten?

not7cd:
I'm getting reports that it works to some extent with Starbook TEN. I can't support this, so maybe someone with access to one could contribute.

knro:
Ok great thanks!

PR #895, opened 2019-03-23
use external weather station safety logic via script or curl and support hysteresis on soft errors

d33psky:


knro:
Great work Hans! Can we get some documentation/guide using the driver as well with screenshots?

d33psky:
Thanks. Yes definitely. I'll make an entry in https://indilib.org/devices/weather-stations.html with screenshots. I'm also thinking about providing two example scripts. A bit like libindi/drivers/dome/dome_script.txt but unsure where to put those. Then there's the 'weather safety service' itself which this driver talks to. I want to help people build one of their own, and what the driver expects the service provides.

libindi/drivers/weather/weather_safety_proxy.cpp
```diff
@@ -0,0 +1,474 @@
...
+    setVersion(1, 0);
+
+    setWeatherConnection(CONNECTION_NONE);
+}
+
+WeatherSafetyProxy::~WeatherSafetyProxy() {}
+
+const char *WeatherSafetyProxy::getDefaultName()
+{
+    return (const char *)"Weather_Safety_Proxy";
```
knro: Do not use _ in the name. Also no need for (const char *)

knro:
Thanks, can you attach the screenshots? I couldn't find them in the submission.

d33psky:
i'm still working on the weather-stations/weather-safety-proxy.html

d33psky:
> i'm still working on the weather-stations/weather-safety-proxy.html
This is now done. https://www.indilib.org/devices/weather-stations/weather-safety-proxy.html is ready.

knro:
Is this PR ready or you need to include more changes?

d33psky:
it's ready to be merged into master

PR #898, opened 2019-03-30
Create issue template for bug reports

not7cd:
Looking at issues I can see an emerging pattern and lack of proper data to tackle some of the issues.


knro:
Thanks!! We really need this!!

knro:
Oppps.. looks I merged prematurely. Github expects this file where exactly? Because when I create a new issue, it's not loaded.

not7cd:
It might be a case for legacy workflow. 
I think I have put this in the right place. New workflow should be available here:
https://help.github.com/en/articles/about-issue-and-pull-request-templates

not7cd:
Because normal users can't add labels to issues, this template might not work. I need to look through examples again.

knro:
I already added a couple of templates via the Github interface.

PR #904, opened 2019-04-04
Starbook 0.5~something

not7cd:
Somehow my PR slipped through OSX CI


not7cd:
Also forgot to enable and test location handling

PR #909, opened 2019-04-09
Change INDI Track Mode

Chris-Rowland:
the Indi track mode has been changed to Sidereal, Solar, or Lunar for Equatorial mounts only

### DIFF
--- a/libindi/drivers/telescope/celestrondriver.cpp
+++ b/libindi/drivers/telescope/celestrondriver.cpp
@@ -578,6 +578,7 @@ bool CelestronDriver::sync(double ra, double dec, bool precise)
     return send_command(cmd, strlen(cmd), response, 1, true, true);
 }
 
+// NS+ 5.28 and more only, not StarSense
 bool CelestronDriver::unsync()
 {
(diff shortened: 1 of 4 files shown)


Chris-Rowland:
I'm uneasy about adding custom track rates. There are pass through commands that allow this, used for Alt Azm tracking, but I'm not at all sure how they will behave when integrated with other commands such as turning tracking on and off, guiding, PEC, and the various mount jog commands.

All this could be a support nightmare, it could also work with some hardware versions and not others and the chance of getting good data from users is slim.

On that theme I've been discussing the various rate commands with Derik Devecchio who is the Celestron developer and he points out that the passthrough command rate will be lost when any HC command is sent, such as starting tracking or a goto command, maybe guiding.  He did suggest adding RS232 HC commands to set/get the track rate, he didn't see it as a big problem, but no promises.  It will be new NS+ HCs only of course.

PHD2 has a comet tracking mode where the guide position is changed at the appropriate rate and the normal guiding used to compensate.  I'd use that.

Chris

Chris-Rowland:
Puzzled, the errors seem to be in third party drivers.

PR #918, opened 2019-04-16
Detect another Celestron Nexstar EVO broadcast

luzik:
My brand new EVO sends such packets in AP mode

IP 192.168.123.102.4097 > 255.255.255.255.55555: UDP, length 108

..but listen on TCP 2000


knro:
@jochym Can you please check?

luzik:
Uhg, this source UDP port is quite dynamic. Today it was 4098. Nevertheless source UDP port should not be in relation with listening TCP port.

jochym:
Yes. You are right. Using source port in the detection is wrong. It may change with theimplementation and system activity.

jochym:
Are you shure @luzik that the package length is different?

luzik:
Positive. First comment actually contains tcpdump line. Beside that "recvlen" printf was tested.

PR #919, opened 2019-04-17
Improve detection of Quantum filter wheel

Chris-Rowland:
This adds code to Handshake() to identify a quantum wheel; previously any open serial port would be accepted.
I've also added documentation for the quantum protocol and implemented  timeouts on serial reads because the wheel seems not to respond to commands that are not recognised and this would leave the serial port occupied.
The wheel has it's own USB ID, could these be used to identify the wheel?


knro:
For serial devices, it's not easy to figure out the VID:PID. The handshake is sufficient.. though if there a way to know from the VID:PID, I'd imagine it would be faster (assuming it is truly unique).

Chris-Rowland:
Thanks Jasem, the read seems fine.  I could read the filter names and offsets from the wheel but that's something for another day.
Don't know what to do about the Travis CI build error, it doesn't seem to be anything to do with me.

knro:
Yeah, the error is elsewhere. Thanks a lot!

PR #929, opened 2019-04-27
Incompatible frame sizes are reduced internally to ensure compatibility

sterne-jaeger:
ASI requires, that a ROI has a width mod 8 = 0 and height mod 2 = 0. To ensure this, the driver reduces the frame size internally by the respective mod values, if necessary. That means there might be a difference between the set frame size and the size of the captured image.

This is done intentionally to avoid very technical error message that might annoy users. For details about this decision see the discussion here: https://indilib.org/forum/ccds-dslrs/4956-indi-asi-driver-bug-causes-false-binned-images.html

### DIFF
--- a/3rdparty/indi-asi/asi_ccd.cpp
+++ b/3rdparty/indi-asi/asi_ccd.cpp
@@ -44,6 +44,9 @@ static int iAvailableCamerasCount;
 static ASI_CAMERA_INFO *pASICameraInfo;
 static ASICCD *cameras[MAX_DEVICES];
 
+static bool warn_roi_height = true;
+static bool warn_roi_width = true;
+
 //pthread_cond_t cv         = PTHREAD_COND_INITIALIZER;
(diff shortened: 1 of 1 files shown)


knro:
How often is this warning triggered? it might cause further confusion to users.

sterne-jaeger:
Always when the ROI or binning is set. I checked it with EKOS. Here the message comes always twice, since when the binning is changed, the ROI is updated. But for a setup with 
- Red 1x1
- Green 4x4
- Red 4x4

the message is generated twice by the Green capture, but not repeated for the following 4x4 Red frame.
The question is, what creates more confusion: a warning that appears from time to time or recognizing that the captured frame is smaller than set.

knro:
I think it should be set only **once** when the condition is first met and suppressed thereafter. IMO, if the user keeps seeing that warning, it will make them believe things are broken rather than just a reminder of what's going on.

sterne-jaeger:
Ok, you are right. And I will reduce it to info-level.

PR #939, opened 2019-05-03
Implement CW/CCW motion controls in ScopeDome driver

jpaana:
In current version pressing CW or CCW motion buttons crashes due to unimplemented Move method in the driver class when dome base class tries to call it causing it to call itself instead ending up in infinite recursion.


knro:
More changes planned or ready to land?

jpaana:
I think this one is ready now, I'll do another one with calibration at some point.

PR #942, opened 2019-05-06
Bugfix  pulse and movement

TallFurryMan:
Fixes North/South guiding/movement orientation.
Fixes property switch "Use Pulse Cmd" so that it rejects the pulse command feature (only timed moves).
Markdown documentation was missing.

### DIFF
--- a/libindi/drivers/telescope/eq500x.cpp
+++ b/libindi/drivers/telescope/eq500x.cpp
@@ -674,6 +674,39 @@ void EQ500X::setPierSide(TelescopePierSide side)
     IDSetSwitch(&PierSideSP, "Not supported");
 }
 
+bool EQ500X::MoveNS(INDI_DIR_NS dir, TelescopeMotionCommand command)
+{
+    // EQ500X has North/South directions inverted
+    int current_move = (dir == DIRECTION_NORTH) ? LX200_SOUTH : LX200_NORTH;
+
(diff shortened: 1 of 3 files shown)


TallFurryMan:
Aaah let me change the version number.

TallFurryMan:
@knro b5ed8ce fixes the issue experienced by other pull requests on the OSX build. However two hosts were not resolved while building. Can the build be retriggered as it seems a transitory DNS issue, or could you merge this to fix the issue?

TallFurryMan:
Oh I see that none of my unit tests in the CircleCI do run, apparently because nothing is built? That seems to be a recent regression as I did verify those on my own CircleCI account at some point. I'll double check.

TallFurryMan:
OK: my CircleCI config only really build-test my own branch `master`, where I reduced the build to Ubuntu Bionic. When I push a pull request branch, all variations are tested (because I kept the original CircleCI config in the PR), but actually nothing is built at all (because the PR branch is filtered out) and thus nothing is tested. So the CircleCI reports are misleading.

Do we need to fix the CircleCI config in the master branch at some point, so that contributors do trig their own CircleCI runs before the PR is accepted?

TallFurryMan:
Umm apparently I should push my master branch as pull request actually for the CircleCI tests to run. That's an interesting problem.

PR #947, opened 2019-05-08
Added support for multiple read modes in QHY sCMOS cameras

irydeo:
Added support for multiple read modes in QHY sCMOS cameras, code has been tested in one QHY42Pro, that allows two read modes:

* HDR: output is a single 4096*2048 resolution image, it contains two images. Left is low gain channel image and right side is high gain channel image.
* STD: output is a single 2048*2048 resolution image. 

Implementation uses INumber types, as future improvement I plan to use ISwitch that will provide a more user friendly implementation, showing read modes names instead its index number.


knro:
Great! Why aren't you using ISwitch now though since it makes more sense as there are only two read modes?

irydeo:
> Great! Why aren't you using ISwitch now though since it makes more sense as there are only two read modes?

Hi Jasem,

sure, that's my idea for next commits, current implementation could be considered as _beta_ (it is stable but focused only on functionality and far to be _user friendly_ :) ), my concern is that number of read modes is unknown at the beginning... currently these kind of cameras have two, but probably more can be added.

knro:
Sure, I'll merge this for testing now.

PR #948, opened 2019-05-15
Eqmod: Fix RA drift with HEQ5 firmware 106

pchev:
The HEQ5 with the old motor board with firmware version 106 as a bug that make the RA to drift at a rate about 4 arcsec per minute when driven at sidereal rate. Unfortunately this board cannot be upgraded to a new firmware. This issue was the reason for the "drift compensation" slider in EQASCOM.

The solution is to override the Timer Interrupt Frequency returned by the board when this board version is detected, as this is already done for the 80GT and 114GT.

I find the new frequency after some trial and now my mount as no more drift.


knro:
Can you bump the version of INDI EQMod driver and also the associated debian changelog? Thanks!

pchev:
Is this version change OK?

Issue #951, opened 2019-05-22
EQ500-X: different slew rate after go-to in SkySafari


TallFurryMan:
**Describe the bug**
Omegon EQ500-X driver has a different slew rate before and after issuing a go-to request from SkySafari.

**To Reproduce**
Exact steps to reproduce the behavior.
1. Run driver, connect from SkySafari
2. Select slew rate other than guide rate, execute a manual move
3. Execute a go-to to an object
4. Check the current slew rate displayed in SkySafari
5. Execute a manual move

**Expected behavior**
5. Effective slew rate on the mount and slew rate displayed in SkySafari are identical.

**Actual behavior**
5. Slew rate at which the manual move is executed is "guide", slew rate displayed in SkySafari retains the setting before the go-to request.

**Screenshots**
n/a

**Desktop (please complete the following information):**
 - OS: undisclosed, probably StellarMate 1.2.x or early 1.3.x
 - Version: undisclosed, probably 1.7.7 nightly or 1.7.8

**Log Files**
n/a

TallFurryMan:
The driver actually reports the slew rate change properly, but the SkySafari client has no return path for the slew rate. Should `indi_eq500x_telescope` simply restore the previous slew rate on successful execution or abort, or should we modify the SkySafari middleware to notify slew rate change to SkySafari?

TallFurryMan:
Working as intended. Closing.

PR #952, opened 2019-05-23
Bugfix  slew rate after goto

TallFurryMan:
When a Go-To is requested to the driver, the final slew rate was always "guide". Because SkySafari is not updated with slew rate changes, there was in general a discrepancy between the mount control on the mount settings.

This PR restores the slew rate after a Go-To request and an Abort request. It also makes sure executing a second Go-To while one is running properly restores the original slew rate.

This PR adds tests reproducing the issue, then code fixing that issue. CircleCI validated all tests in the branch this PR originates from, but this PR does not run any actual tests.

### DIFF
--- a/libindi/drivers/telescope/eq500x.cpp
+++ b/libindi/drivers/telescope/eq500x.cpp
@@ -491,10 +491,8 @@ bool EQ500X::ReadScopeStatus()
         else
         {
             LOG_INFO("Slew is complete. Tracking...");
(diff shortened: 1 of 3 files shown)


TallFurryMan:
Fixes #951.

TallFurryMan:
Sorry, while I can retrig builds to work around package installations, I can't fix those Travis OSX builds.

PR #960, opened 2019-06-08
[limesuite] updated dependency script

iliaplatone:


knro:
Thanks! It's still failing with soapy.

iliaplatone:
Jasem, it'n now installing limesuite, but it's failing osx installation in a different point.

knro:
Thanks, that fixes the issue. The remainder issue is related to SBIG driver on OSX.

PR #961, opened 2019-06-10
Added calibration and fixed temperature update

rishi-garrod:
Added a Calibration routine.
Changed the focuser in motion timer function to cease when movement has ceased.
Reset temperature counter to allow updates of the temperature.

### DIFF
--- a/libindi/drivers/focuser/sestosenso.cpp
+++ b/libindi/drivers/focuser/sestosenso.cpp
@@ -16,6 +16,50 @@
     License along with this library; if not, write to the Free Software
     Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 
+    Commands and responses:
+
+    Only use the SM/Sm commands during calibration. Will cause direction reversal!
+    #Sm;xxxxxxx! Set xxxxxxx as min value
+    #SM!	Set current position as max
+    #SM;xxxxxxx!	Set xxxxxxx as max value (xxxxxxx between 0 to 2097152)
+
+    #SPxxxx! Set_current_position as xxxx
+    #SC;HOLD;RUN;ACC;DEC! Shell_set_current_supply in HOLD, RUN, ACC, DEC situations (Value must be from 0 to 24, maximum hold value 10)
+    #QM! Query max value
+    #Qm! Query min value
+    #QT! Qeury temperature
+    #QF! Query firmware version
+    #QN! Read the device name	-> reply	QN;SESTOSENSO!
+    #QP! Query_position
+    #FI! Fast_inward
+    #FO! Fast_outward
+    #SI! Slow_inward
+    #SO! Slow_outward
+    #GTxxxx! Go_to absolute position xxxx
+    #MA! Motion_abort and hold position
+    #MF!	Motor free
+    #PS! param_save save current position for next power ON and currents supply
+    #PD! param_to_default , and position to zero
+
+    Response examples:
+
+    #QF! 14.06\r
+    #QT! -10.34\r
+    #FI! FIok!\r
+    #FO! FOok!\r
+    #SI! SIok!\r
+    #SO! SOok!\r
+    #GTxxxx! 100\r 200\r 300\r xxxx\r GTok!\r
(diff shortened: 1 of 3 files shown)


libindi/drivers/focuser/sestosenso.cpp
```diff
@@ -88,10 +132,22 @@ bool SestoSenso::initProperties()
...
-//    IUFillNumber(&LimitsN[SS_MIN_LIMIT], "FOCUS_MIN_STEP", "Min", "%.f", 0, 10000., 0., 0.);
-//    IUFillNumber(&LimitsN[SS_MAX_LIMIT], "FOCUS_MAX_STEP", "Max", "%.f", 10000,  2097152., 0., 200000.);
-//    IUFillNumberVector(&LimitsNP, LimitsN, 2, getDeviceName(), "FOCUS_LIMITS", "Limits", MAIN_CONTROL_TAB, IP_RW, 0, IPS_IDLE);
+    // Focuser calibration
+    IUFillText(&CalibrationMessageT[0], "CALIBRATION", "Calibration stage", "");
+    IUFillTextVector(&CalibrationMessageTP, CalibrationMessageT, 1, getDeviceName(), "CALIBRATION_MESSAGE", "Calibration", MAIN_CONTROL_TAB, IP_RO, 0, IPS_IDLE);
+
+    IUFillSwitch(&CalibrationS[0],"CALIBRATION_START","Start", ISS_ON);
+    IUFillSwitch(&CalibrationS[1],"CALIBRATION_NEXT","Next", ISS_OFF);
+    IUFillSwitchVector(&CalibrationSP, CalibrationS, 2, getDeviceName(), "FOCUS_CALIBRATION", "Calibration", MAIN_CONTROL_TAB, IP_RW, ISR_1OFMANY, 0, IPS_IDLE);
```
knro: Maybe it would make more sense to have ISR_ATMOST1 switch with both switches in the OFF position.

Then we CALIBRATION_START is ON, you begin the calibration process. If it is pressed again, the calibration is cancelled/stopping. 

When CALIBRATION_NEXT is ON, you go to the next step (but it is immediately switched to OFF in the driver so that it can be pressed again in the next step by the user).

Sorry if it is not clear!

rishi-garrod: Understanding exactly how the switches work has been a struggle
I don't actually want to be able to cancel the calibration process. It leaves the focuser in an undefined state. You really have to complete the process once you start.
So if I understand correctly all I need to do is change line 140 to ISS_ON?

rishi-garrod:
Where do I add the design? I can't find out where to update the docs.

libindi/drivers/focuser/sestosenso.cpp
knro: You can commit to your branch and it will show up here. No need to cancel the PR.

So making them both ISS_ON will not work. They need both to be ISS_OFF and the rule is ISR_ATMOST1 which means that all switches can be off, but only one can be ON at a time.

After user click START, start the process and then set the switch to ON. START=ON, NEXT=OFF

When user clicks NEXT, go to next stage, but keep START=ON and NEXT=OFF

when process is completed, both switches are OFF.

knro:
oh forget about the "design" part. I just had to add a comment to cover my previous comments. Ignore it. Please just read my last comment on this topic.

libindi/drivers/focuser/sestosenso.cpp
rishi-garrod: Got it. Makes perfect sense. I have set them to OFF, when I start the Calibration process I set START ON and at the end I switch NEXT OFF. I also changed the FastMove to be all OFF.

knro:
I just submitted a few changes to fix a couple of warnings and minor issues, can you please double check?

rishi-garrod:
Is there anywhere I can see which changes you made?

knro:
It's in master, the changes are in this commit https://github.com/indilib/indi/commit/e05e89fa096a746ee5503638b48b45f3b0b00e1f

rishi-garrod:
All looks good. Working as expected. Thanks for your support.

Issue #955, opened 2019-05-29
Configurable polling interval for SQM


skolarianer:
Please consider to make the polling interval of SQMs configurable. AFAIK the driver queries the SQM every second. So the connection to the unit is never really closed. There is no chance for other processes to access the SQM between readouts.

Our unit is connected to an SQM-network and reports measurements every 5 minutes. But when the indi driver is active no other requests are possible.

It'd be great if there was an option to change the polling interval and a closing of the socket after each readout.

knro:
It's already configuration in Options --> Polling. But you can set it to a max of 60000 ms now = 1 minute. I just increased the maximum limit to 600,000 or 10 minutes.

skolarianer:
Sorry, can't find it. Or do you mean other options?
In the options tab of the Pegasus PPB for example I can see polling settings.

![image](https://user-images.githubusercontent.com/6286711/58600108-782c3300-8283-11e9-83ad-9ab413c52bd7.png)

knro:
Right it wasn't added. Please use the next nightly (v1.1 of the driver) and see if it has the control.

skolarianer:
The control element is now available.

It worked with a polling period of 1000ms. When I changed the period to 5000ms, it stopped working and I can’t set it back to 1000ms.

Can’t reconnect either. But when restarting the INDI services I can connect again, but do not get any data anymore. Sometimes I get this error in that case:

` [ERROR] Failed to parse input r, 00.00m,0000573909Hz,0000000000c,0000`

<details>
    <summary>Here is the full log:</summary>

```
[2019-06-07T15:12:02.524 CEST DEBG ][           org.kde.kstars.indi] - INDI: Stopping local drivers...
[2019-06-07T15:12:02.527 CEST DEBG ][           org.kde.kstars.indi] - Removing managed driver "CCD Simulator"
[2019-06-07T15:12:02.528 CEST DEBG ][           org.kde.kstars.indi] - INDIListener: Removing device CCD Simulator with unique label  ""
[2019-06-07T15:12:02.530 CEST DEBG ][           org.kde.kstars.indi] - Removing managed driver "SQM"
[2019-06-07T15:12:02.530 CEST DEBG ][           org.kde.kstars.indi] - INDIListener: Removing device SQM with unique label  "SQM"
[2019-06-07T15:12:02.533 CEST DEBG ][           org.kde.kstars.indi] - INDIListener: Removing client manager for server localhost @ 7624
[...]

PR #972, opened 2019-06-26
Modified SQM::getReadings() to support a serial number in parsed answer

Noodlesalat:
Sometimes the SQM answers with its serial number at the end of its data stream. Now it should always send the serial number and indi ignores it.
It should be a hotfix for this undocumented behaviour of the sqm.

This is related to #955


knro:
Thanks! Have to test this with my SQM.

PR #976, opened 2019-06-28
Fix error: ‘prop’ was not declared in this scope

tstibor:


knro:
Sorry I pushed the fix a few minutes ago before I see this. Thanks for the heads up.

Issue #982, opened 2019-07-06
Downscale of 16bit stream


nadvornik:
**Describe the bug**

ASI1600mm camera has 12bit ADC so the 16bit output has lowest 4 bits set to zero.
The downscale  code in StreamManager::asyncStream
https://github.com/indilib/indi/blob/master/libindi/libs/stream/streammanager.cpp#L326
takes lowest 8 bits so the result is effectively only 4 bit.

I think the downscale code should use ADC precision.

knro:
As you can see, the downscale code is just cutting off anything largest than 255 and is not performing proper 16bit to 8bit downscaling since this can be pretty slow. We need a multithreaded and smart downscaling that utilizes some of the modern processors capabilities on x86 and arm processors so that it won't causes a significant delay in streaming.

knro:
Ok, I'll test it with the zwo cameras I have and see if I can get better streaming.

nadvornik:
Right shift by 4 bits before cutting values over 255 should not be that slow.
The question is how to get the ADC precision of the camera.

knro:
I haven't tested ZWO yet.. does it report bit depth as 12bit in the main control panel?

nadvornik:
yes

ebertin:
That would be great! Note that in principle the raw data should be gamma-compensated before being streamed and displayed: checkout [this link](https://blog.johnnovak.net/2016/09/21/what-every-coder-should-know-about-gamma/) or (sorry for the self-citation) [this paper](http://adsabs.harvard.edu/abs/2012ASPC..461..263B).
Although realtime encoding at fast video rates may require a reasonably fast system with SIMD instructions (or a GPU).
Note that the KStars FITS viewer also lacks proper gamma compensation.

knro:
@ebertin Thanks for the reference. Many image related functions in both INDI & KStars could make use of SSE/SIMD instructions as now I'm just relying on the compiler to be smart enough about loops which isn't enough.

Do you know what's the best way to proceed for this downscaling to 8bit? You know about this a lot more than my superficial knowledge of the subject.

We have uint16_t input buffer, native bit depth (16, 14, 12, or 10) and we want uint8_t proper live stream of the data.

ebertin:
Well formally you should use equation (1):
output = round(255 * p)
where:
 p = 0 if I < 0
 p = a * I if 0 <= I <= Ii
 p = (1+b) * powf(I, 1.0/gamma) - b if Ii <= I <= 1
with:
[...]

PR #984, opened 2019-07-11
Apply gamma when downscaling 16bit to 8bit stream

nadvornik:
First attempt to solve #982

### DIFF
--- a/libindi/libs/stream/streammanager.cpp
+++ b/libindi/libs/stream/streammanager.cpp
@@ -44,6 +44,8 @@ StreamManager::StreamManager(DefaultDevice *mainDevice)
     m_isStreaming = false;
     m_isRecording = false;
 
+    prepareGammaLUT();
+
     // Timer
     // now use BSD setimer to avoi librt dependency
     //sevp.sigev_notify=SIGEV_NONE;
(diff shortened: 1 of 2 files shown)


knro:
How does it work with the ZWO camera now? can you post pre and post screenshots? 

Under simulators, it seems to apply a high-pass filter and all the lower intensities stars are cut off.

@ebertin is this expected?

post_gamma
pre_gamma

nadvornik:
Near zero it depends on the parameter a = 12.92
With downscaling it is
output = input / 65536 * 256 * 12.92
So it cuts off the lowest ~4 bits of 16bit.
With 12bit ADC it does exactly what I want.

knro:
What about 10 and 14 bits? we have to accommodate these as well.

nadvornik:
I can see these options:
-  use non-standard gamma curve parameters, maybe make it it configurable via property,
- leave it as it is - 10bits is OK and for 14bits there will be 2bits loss of precision - I don't think it would have significant effect - the simulator images above have unrealistic S/N ratio, IMHO.
- implement 16bit streaming

ebertin:
Well the sky background is not supposed to lie at 0 anyway. There should normally be an offset. Unless the images are dark-subtracted, but without applying an offset it would not make any sense because negative noise peaks would be clipped?

knro:
Ok, since this is for streaming, let us check the user feedkback.

PR #992, opened 2019-07-29
Various FocusLynx fixes / enhancements

DanielVanNoord:
This fixes an issue with the new Optec ThirdLynx FocusLynx controller (which allows a value in a response to be null be default). It also changes the isResponseOK() to match the behavior of the Optec Gemini (by reading to the END for commands) and adds the new Optec Sagitta focuser device type.


knro:
Great, thanks! Is there a reason to increase the timeout?

DanielVanNoord:
Just to match Gemini, I saw similar behavior with the driver failing to read all the way to the end response.

PR #994, opened 2019-08-03
Celestron Focuser position not update with Evolution mount

docbobo:
Celestron NexStar Evolution mounts do not seem to support retrieval of pier side information. At least with my mount (latest FW), the Celestron GPS driver even warns about that fact upon startup.

The ReadScopeStatus() operation exits if HasPierSide() is FALSE, directly before it would read the position of the motorized focus motor. This PR reverses the order: first read the position of the focus motor, then perform the pier side handling. 

At least here, this fixes the problem: the motor position is reported correctly, thus enabling autofocus workflows.


docbobo:
@Chris-Rowland

Chris-Rowland:
Seems OK.

PR #996, opened 2019-08-04
Update OnStep to new driver version

james-lan:
Version 1.7:
    - Added support for Reporting Guide rate (to PHD2 among others)
    - Updated Error codes to match up with Android/SHC (Unknown reserved for unknown, so Unspecified = Unknown on other platforms)
    - Added descriptions to SlewRate to match, slider kept which matches OnStep values
    - Support for up to 9 stars for alignment
    - Changed align so the last step isn't the (Optional) Write to EEPROM
    - Added support for polar adjustments, without having to redo the entire model. (:MP# command)
    - Support for Full Compensation/Refraction only, and single or dual Axis tracking
    - Cleanups

(Apologies for the cleanups, It shouldn't have mixes of tabs/spaces now (all spaces), but it makes the diffs a annoying now, but should be better in the future.)


james-lan:
Somehow got an old name in the header when merging things, which is fixed now. Sorry for the close/reopen.

PR #997, opened 2019-08-11
Add missing fclose calls

jpaana:
As per thread at https://www.indilib.org/forum/development/5528-inditelescope-cpp-left-open-files.html I also checked a few other places which had left files open in my system and added the missing fclose calls.


knro:
Great thanks for catching these!

PR #1004, opened 2019-08-29
changed dsp_t into double in libDSP

iliaplatone:


knro:
Ok, so CI error about under param. You can use INDI_UNUSED to suppress such errors or find why they are not getting used.

knro:
There is another unused variable error in the CI.

PR #1013, opened 2019-09-15
Add methods to set polling rate limits in DefaultDevice

jpaana:
As discussed in thread at https://www.indilib.org/forum/development/5694-override-focuser-polling-period-control.html it would be useful to have methods to set device specific minimum and maximum polling rates. Some devices don't work with too frequent polling and some need to be polled often enough (at least ScopeDome is such, it resets the serial connection if there isn't traffic for a certain time). I haven't actually tested these functions yet, but they seemed straight forward enough to implement and at least it compiles :)

### DIFF
--- a/libindi/libs/indibase/defaultdevice.h
+++ b/libindi/libs/indibase/defaultdevice.h
@@ -463,8 +463,7 @@ class INDI::DefaultDevice : public INDI::BaseDevice
         }
 
         void setDefaultPollingPeriod(uint32_t period);
-        void setMinimumPollingPeriod(uint32_t period);
-        void setMaximumPollingPeriod(uint32_t period);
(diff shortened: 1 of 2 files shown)


libindi/libs/indibase/defaultdevice.h
```diff
@@ -463,6 +463,8 @@ class INDI::DefaultDevice : public INDI::BaseDevice
         }
 
         void setDefaultPollingPeriod(uint32_t period);
+        void setMinimumPollingPeriod(uint32_t period);
```
knro: Wouldn't be better to have one function?

setPollingPeriodRange(min, max) ?

jpaana: Yes, that's better, done.

PR #1023, opened 2019-10-03
Add sync capability for all Optec focuser types

sterne-jaeger:
The sync capability has disappeared for relative focusers. At least my PDMS focuser accepts the sync command and I assume that all other relative focusers do this as well.

### DIFF
--- a/drivers/focuser/focuslynxbase.cpp
+++ b/drivers/focuser/focuslynxbase.cpp
@@ -67,10 +67,11 @@ FocusLynxBase::FocusLynxBase()
     focusMoveRequest = 0;
     simPosition      = 0;
 
-    // Can move in Absolute & Relative motions, can AbortFocuser motion, and has variable speed.
+    // Can move in Absolute & Relative motions, can AbortFocuser motion, can sync, and has variable speed.
     FI::SetCapability(FOCUSER_CAN_ABORT    |
                       FOCUSER_CAN_ABS_MOVE |
                       FOCUSER_CAN_REL_MOVE |
+                      FOCUSER_CAN_SYNC     |
                       FOCUSER_CAN_REVERSE  |
                       FOCUSER_HAS_BACKLASH);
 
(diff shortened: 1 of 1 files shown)


knro:
Ok, I figured related focusers do not support syncing to an absolute position? will it not be forgotten when the power is cycled?

sterne-jaeger:
At least mine, a PDSM Focuser, remembers the position - and that's exactly the problem when you changed the position manually.

knro:
So if it remembers absolute position, how come it's a relative focuser??

sterne-jaeger:
Hm, good point. I thought an absolute focuser always knows its position, no matter what happens. As far as I understood is this exactly what absolute encoders do.

But if we declare the PDMS as absolute focuser, I'm equally fine. But using it without the SYNC button is not a good idea. Maybe we should contact Jeff from Optec. He should have the best overview of the behaviour of their focusers.

sterne-jaeger:
Here the answer from Optec:

Interesting question Wolfgang.  

Under the ASCOM definitions any focuser capable of discrete steps is considered an "absolute" focuser.   A plain old DC motor used to drive a focuser can be used by varying the input voltage - this was the preferred way when the original LX200 was introduced for instance.  A DC motor with varying voltage input was classified as a "relative" focuser by the original ASCOM team.

We've extended these definitions.  We define a "homing absolute" focuser as any focuser that performs a home and knows it's absolute position.  The original TCF-S, TCF-Lynx, Fast-Focus, TCF-Leo, and Gemini are all absolute homing focusers.

Focusers that do not home but still have discrete steps would be considered "non-homing absolute" focusers.  Examples include your Posi-Drive motor, HSM, QuickSync, and DirectSync motors.  We include a clutch mechanism that allows manual focus but you must synchronize (Sync) the position again after the manual move to restore the absolute position.

Fortunately, most auto-focus routines work in a relative fashion.  That is, the auto-focus algorithm moves from the current position (regardless of the value) to the best focus relative to the current position.

I hope this makes sense.

    Very best regards,

	Jeff Dickerman
	Optec Sales.

PR #1028, opened 2019-10-06
Fix out of bounds access in myfocuserpro2.cpp

jpaana:
There seems to have been a small gaffe that TemperatureSettingN array is being accessed by index 1 even though there is only one value in the array. This causes compiler warning and is obviously bad otherwise too, so I wonder why no one noticed before :)

`
/home/pi/src/indi/drivers/focuser/myfocuserpro2.cpp: In member function ‘bool MyFocuserPro2::readTempeartureCoefficient()’:
/home/pi/src/indi/drivers/focuser/myfocuserpro2.cpp:458:30: warning: array subscript 1 is above array bounds of ‘INumber [1]’ [-Warray-bounds]
         TemperatureSettingN[1].value = val;
         ~~~~~~~~~~~~~~~~~~~~~^
/home/pi/src/indi/drivers/focuser/myfocuserpro2.cpp: In member function ‘virtual bool MyFocuserPro2::ISNewNumber(const char*, const char*, double*, char**, int)’:
/home/pi/src/indi/drivers/focuser/myfocuserpro2.cpp:811:65: warning: array subscript 1 is above array bounds of ‘INumber [1]’ [-Warray-bounds]
             if (!setTemperatureCoefficient(TemperatureSettingN[1].value))
                                            ~~~~~~~~~~~~~~~~~~~~~^
`

### DIFF
--- a/drivers/focuser/myfocuserpro2.cpp
+++ b/drivers/focuser/myfocuserpro2.cpp
@@ -455,7 +455,7 @@ bool MyFocuserPro2::readTempeartureCoefficient()
     int rc = sscanf(res, "B%d#", &val);
 
     if (rc > 0)
-        TemperatureSettingN[1].value = val;
+        TemperatureSettingN[0].value = val;
     else
     {
         LOGF_ERROR("Unknown error: Temperature Coefficient value (%s)", res);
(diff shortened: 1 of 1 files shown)


knro:
@ajt68 FYI

PR #1031, opened 2019-10-13
Deep Sky Dad AF3 driver

DeepSkyDad:
Driver for our new autofocuser

### DIFF
--- a/drivers/focuser/deepskydad_af3.h
+++ b/drivers/focuser/deepskydad_af3.h
@@ -1,0 +1,147 @@
+/*
+    Deep Sky Dad AF3
+
+    Copyright (C) 2019 Pavle Gartner
+
+    Based on Moonline driver.
+    Copyright (C) 2013-2019 Jasem Mutlaq (mutlaqja@ikarustech.com)
+
+    This library is free software; you can redistribute it and/or
+    modify it under the terms of the GNU Lesser General Public
+    License as published by the Free Software Foundation; either
+    version 2.1 of the License, or (at your option) any later version.
+
+    This library is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


drivers/focuser/deepskydad_af3.h
```diff
@@ -0,0 +1,147 @@
...
+        INumber MoveCurrentMultiplierN[1];
+        INumberVectorProperty MoveCurrentMultiplierNP;
+
+        //Current hold
+        INumber HoldCurrentMultiplierN[1];
+        INumberVectorProperty HoldCurrentMultiplierNP;
+
+        // Max movement
+        INumber FocusMaxMoveN[1];
+        INumberVectorProperty FocusMaxMoveNP;
```
knro: There is no need to define this property since it is defined by the INDI::FocuserInterface already (FocusMaxPosNP). So instead just override the function bool SetFocuserMaxPosition(uint32_t ticks) and implement the functionality there.

DeepSkyDad: Hi Jasem, thank you!

I am not sure I understand.

_"There is no need to define this property since it is defined by the INDI::FocuserInterface already (FocusMaxPosNP)"_

I do not have FocusMaxPosNP defined in the header. I have only defined FocusMaxMoveN, as it is not defined in FocuserInterface?

DeepSkyDad: Just to clarify that I understand this correctly:

Maximum position (FocusMaxPosNP) is the biggest value that motor can be at. e.g. 1000000?

Maximum move is maximum change of position in single movement and I did not find this in the focuserinterface. So I added FocusMaxMoveNP.

Is there already a property for maximum number of steps in a single move?

knro: Ah okay, I got the two confused then. Yes, you understanding is correct.

PR #1035, opened 2019-10-20
Fix minor typos

yurchor:
Many thanks in advance for fixing these typos.


knro:
wow thanks for all these changes. What program do you recommend to check for such typos in strings? I'm using QtCreator so I'll see if they have aspell support.

yurchor:
> wow thanks for all these changes. What program do you recommend to check for such typos in strings? I'm using QtCreator so I'll see if they have aspell support.

A simple script then manual filtering:

#!/bin/bash 

for i in `find . -name \*.cpp -o -name \*.c`
do
	grep -oP '(?<=").*(?=")' <$i|hunspell -d en_US -l>>err
done
cat err|sort|uniq>errors
rm -r err

PR #1037, opened 2019-10-21
A few more typo fixes

yurchor:
Overseen by the previous inspection.

Many thanks for reviewing this and your patience.


knro:
Good catch! Will try to run these checks regularly. It would be nice to have spell checking to be part of the CI checks as well in the future.

PR #1039, opened 2019-10-24
Backlash and Go Home support added

ajt68:


knro:
Thanks! Please update the version number in drivers.xml file as well. They are still set there manually.

ajt68:
Ermm OK, where is the drivers.xml? Should I remove the setversion from the focuser code if it has to be set manually in the driver.xml, or do both files set the version in different places?

ajt68:
I am not happy with NOT using the built in base backlash support, a single value and a single enable. 

The  focuser has independent steps and enables for both in and out, so I coded to support the permutations, but the more I think about it the more I convince myself that one value applied for both in and out or neither is less likely to cause problems.
The way I see it is if you have different in and out values then the absolute position will creep towards one end or the other (by the difference in the backlash steps) if you keep switching from in and out by varying amounts. 
I can see that for very specific cases that there may be a small difference, but again on average for the full length of the rack they must be the same, otherwise this would cause the gears to bind!
I'm thinking in circles about this one, but the fact INDI only supports a single value makes me feel I should also do the same and implement using the INDI supported control.

knro:
That's fine. Not all drivers/hardware would match up with the generic focuser interface in INDI. IIRC, most drivers use a single value regardless of direction. Thanks for the PR, will merge it soon.

PR #1040, opened 2019-10-25
PMC8 Driver update

karlrees:
PMC8 driver update to better support Ethernet/WiFi connections, simplify mount communication, and more intelligently determine mount type and associated parameters

- Wrapped all tty_reads in a get_pmc8_response function that performs various associated actions such as error checking, debug output, and compensation for certain errors/events that can occur when communicating over WiFi
- Auto detect mount type from firmware
- Added variable for detected mountType to firmwareInfo so that we can access it without relying on the UI switches
- Changed set_myMount function name to set_mountParameters to be more consistent with its purpose
- Set mount parameters from driver if we can auto-detect mount type, instead of waiting for UI to do it
- Moved mount type guessing code to else loop after we check firmware (to continue to support older firmware that does not specify mount type)
- Set default Ethernet address/port for mounts to OEM default
- Moved default serial baud rate assignment to initProperties and made it applicable to all mounts
- Initialized axis scales to a default value to avoid crash in KStars Lite
- Updated version number to 0.3
- Updated drivers.xml to reflect changes and improve labels

### DIFF
--- a/drivers.xml
+++ b/drivers.xml
@@ -209,17 +209,17 @@
             <driver name="Digital Setting Circle">indi_dsc_telescope</driver>
             <version>1.0</version>
         </device>
-        <device label="Explore Scientific PMC-Eight" manufacturer="Explore Scientific">
+        <device label="ES G-11 PMC-Eight" manufacturer="Explore Scientific">
             <driver name="PMC8">indi_pmc8_telescope</driver>
-            <version>0.2</version>
+            <version>0.3</version>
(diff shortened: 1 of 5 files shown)


knro:
Thank you for the PR, this look pretty good to me. What do you think @mikefulb ?

knro:
@karlrees Can you please add your name to the file credits as well in the header?

karlrees:
Done

PR #1043, opened 2019-11-15
Added support for iOptron iPolar.

wvreeven:
A simple change to add the iOptron iPolar scope with the correct pixel sizes.


knro:
drivers.xml needs to include an entry for iOptron iPolar

PR #1047, opened 2019-12-01
Simulator pier side

sterne-jaeger:
Some scopes do not have a native pier side information. This creates problems with guiding on different targets, where it is necessary to receive pier side information so that the guider knows when to swap calibration data.

This change shifts the capability of pier side calculation from the telescope simulator down to the base class of all INDI telescopes. Telescope drivers that want to use this need to turn on the new TELESCOPE_HAS_PIER_SIDE_SIMULATION.

Additionally, a switch is added so that the pier side simulation can be switched on or off at any time by the user.

As a first step, the telescope simulator is using this capability.


libs/indibase/inditelescope.cpp
```diff
@@ -1432,6 +1456,22 @@ bool Telescope::ISNewSwitch(const char *dev, const char *name, ISState *states,
...
         }
 
+        ///////////////////////////////////
+        // Simulate Pier Side
+        ///////////////////////////////////
+        if (!strcmp(name, SimulatePierSideSP.name))
+        {
+            IUUpdateSwitch(&SimulatePierSideSP, states, names, n);
+            int index = IUFindOnSwitchIndex(&SimulatePierSideSP);
+            if (index == -1)
```
knro: Switch state must be set follow by IDSetSwitch(...) before returning.

sterne-jaeger: I'm not so sure what IUUpdateSwitch() does. Do I need it at all, since I set the switch values in setSimulatePierSide()?

libs/indibase/inditelescope.cpp
```diff
@@ -1709,9 +1749,16 @@ bool Telescope::updateTime(ln_date *utc, double utc_offset)
 
 bool Telescope::updateLocation(double latitude, double longitude, double elevation)
 {
-    INDI_UNUSED(latitude);
-    INDI_UNUSED(longitude);
+    // FIXME (sterne-jaeger): needs to be called from all classes deriving from inditelescope.
```
knro: perhaps a better approach is to define updateObservorLocation() which does this and then call it from ISNewSwitch just after calling updateLocation? This way the derived classes don't need to call this.

sterne-jaeger: Ooops, that's a leftover from a first attempt. I first thought I would need it for calculating the pier side - which turned out to be unnecessary. I'll remove it.

sterne-jaeger:
> I'm think this is not really a capability. So Maybe INDI::Telescope should see if TELESCOPE_HAS_PIER_SIDE is defined or not, if NOT define, then it always defines simulate pier side but as DISABLED (by default). Now either the driver or the user can switch it on.

That's what I first tried, but I want to avoid that the pier side simulation appears automatically. Having this capability each driver which is using it needs to turn it on manually.

libs/indibase/inditelescope.cpp
sterne-jaeger: Aaah, now I understand what you mean. In the case that index == -1, IDSetSwitch(...) is not called. All right, I'll fix it.

libs/indibase/inditelescope.cpp
sterne-jaeger: OK, have to check it. This functionality is pushed down from telescope_simulator.cpp

sterne-jaeger: I've implemented it now as proposed. Now we keep the observer's location in the memory for all drivers.

Chris-Rowland:
I've also been looking at this but you beat me to it, no matter.
My approach was to add a void fakePierSide(double rightAscension, double latitude)  method that could be in inditelescope and called by the driver at a time of it's choosing, usually after a slew had been started.  The driver would also need to add the TELESCOPE_PIER_SIDE capability.

Your method reduces the amount of change required in the driver but will it interferes with a driver that already implements setPierSide()?

Chris

sterne-jaeger:
> Your method reduces the amount of change required in the driver but will it interferes with a driver that already implements setPierSide()?

No, it doesn't, take a look into telescope_simulator.cpp. In `ReadScopeStatus()`, the telescope simulator sets the pier side like this:
`setPierSide(expectedPierSide(currentRA, currentDEC))`

If a driver wants to implement the simulated pier side, it needs

1. set the pier side in its own `ReadScopeStatus()` implementation as in the example above
2. set the capability `TELESCOPE_HAS_PIER_SIDE_SIMULATION`

With 2. I avoid that the switch for turning on the simulation appears in drivers that did not implement the usage of `expectedPierSide(...)`

Here a screenshot with the option turned off:

Bildschirmfoto vom 2019-12-01 18-22-04

And here with the option turned on:

Bildschirmfoto vom 2019-12-01 18-22-43

knro:
@Chris-Rowland Any further concerns?

Chris-Rowland:
No, the TELESCOPE_HAS_PIER_SIDE_SIMULATION does what is needed.
Chris

Chris-Rowland:
I think this needs reopening.

Unfortunately the pier side simulation does not work properly.  Using the azimuth gives the wrong pier side for places a long way from the primary meridian.  It needs to be a function of hour angle, this is not the same as azimuth.  It also should not change slewing Northor south of East and West, only when slewing through South or North.

The slew when pier side changes needs a fuller implementation so slews that cross the meridian go the long way when crossing the meridian both ways. At present only slews from East to West do this.

And I don't think that the meridian flip behaviour should be implemented when simulate pier flip is off.  At present it seems to be.

I'll have a go at working through this,  I won't change the fundamental structure, just the implementation.

Chris

I found this while looking at a way of checking if a flip had happened by checking the time for the flip slew.  Was surprised that my new code worked until I saw that the flip slew was taking a long time.  What I'm thinking of is that if the flip doesn't work the flip delay is increased and the flip state reset.
C

PR #1049, opened 2019-12-01
Update Celestron Guide Commands

Chris-Rowland:
This is principally to update the Celestron Guide commands because the existing Timed guide commands calls the pulse guide (aka ACK Guide) and this generates a timeout error if it's not supported.
It uses ACK Guide or time guide according to the HC and motor versions ranther than giving the user an option and will send multiple ACK guide commands if the duration if greater then the max AUX guide length of 2.55 seconds.
The code has been tidied up a little, principally using static_cast<T>() because my compiler complained.
Chris


knro:
Thank you, looks good to me. Great work on fixing the warnings.

PR #1052, opened 2019-12-10
Added recording the absolute position of the focuser to the FITS head…

joxda:
…er. Not sure whether the rotator position was actually properly recorded in the FITS header before these changes, but hopefully they are now.


knro:
You're right, it was set to ROTATOR but was listening to FOCUSER. It worked because I had a Focuser/Rotator but this is the correct way to implement it now, great work!

PR #1053, opened 2019-12-11
Fix select motortype, reverse status and enable focuser backslash

tstibor:
According to DMFC serial command table:
http://pegasusastro.com/wp-content/uploads/2015/01/DMFC-Serial-Command-Table.pdf
the command to select the motortype is: "R:X", where X := 0 is DC and X := 1 is stepper.

In addition, the reverse motor direction is read out as N:0 (normal) or N:1 (reverse),
however, the enum is defined as follows:
enum { REVERSED_ENABLED = 0, REVERSED_DISABLED = 1 }.
So it is vice versa to the enum.

The focuser backslash cannot be activated because the command to
enable it, is issued as "C:0" (Disables compensation).


knro:
Thank you Thomas, looks good to me!

PR #1057, opened 2019-12-18
Update Simulated Pier Side 

Chris-Rowland:
The base pier side simulation has been updated to use hour angle rather than azimuth because this makes it work correctly throughout the sky. It also manages enabling and disabling pier side simulation better.

The behaviour of the telescope simulator for doing a long slew when a meridian flip is required has been updated so the flip will be done for all slews that cross the meridian.  This is only done if pier flip simulation is enabled. If it isn't all slews are direct.  This simulates a fork mount on a wedge.


libs/indibase/inditelescope.cpp
```diff
@@ -3027,6 +3042,20 @@ bool Telescope::getSimulatePierSide() const
...
 
+const char * Telescope::getPierSideStr(TelescopePierSide ps)
+{
+    switch (ps)
+    {
+    case PIER_WEST:
+        return "PIER_WEST";
+    case PIER_EAST:
+        return "PIER_EAST";
+    case PIER_UNKNOWN:
```
knro: Would the warning go away if you add default along with PIER_UNKNOWN?

libs/indibase/inditelescope.cpp
```diff
@@ -410,8 +410,9 @@ bool Telescope::updateProperties()
         if (HasPierSideSimulation())
         {
             defineSwitch(&SimulatePierSideSP);
-            // ensure that all properties are set
-            setSimulatePierSide(m_simulatePierSide);
+            loadConfig(true, "SIMULATE_PIER_SIDE");
+            int index = IUFindOnSwitchIndex(&SimulatePierSideSP);
```
knro: Perhaps a better method is to use IUGetConfigSwitch to immediately read back the switch value from the config file? loadConfig does not immediately trigger update of the property in question.

knro:
@sterne-jaeger @rlancaste @TallFurryMan Please share your feedback

Chris-Rowland:
case PIER_UNKNOWN:
     default:
         return "PIER_UNKNOWN";
This gives me a warning about not needing the default statement when all 
options are covered.

Maybe just the default statement will work but I don't like to have to 
write bad code to satisfy the compiler.

I'm not familiar with the detail of the indi support functions, if 
IUGetConfigSwitch is better then so be it.

I've committed a new version...

Chris

libs/indibase/inditelescope.cpp
Chris-Rowland: Replaced PIER_UNKNOWN with default and use IUGetConfigSwitch.

libs/indibase/inditelescope.cpp
```diff
@@ -410,8 +410,10 @@ bool Telescope::updateProperties()
         if (HasPierSideSimulation())
         {
             defineSwitch(&SimulatePierSideSP);
-            // ensure that all properties are set
-            setSimulatePierSide(m_simulatePierSide);
+            loadConfig(true, "SIMULATE_PIER_SIDE");
```
knro: loadConfig is no longer needed now.

sterne-jaeger:
I will check it the next days, I was „slightly“ busy the last two weeks. @Chris-Rowland: Have you checked the MF with changed polling period? We had problems recently with either long or short polling intervals.

Chris-Rowland:
On 19/12/2019 11:01, sterne-jaeger wrote:
>
> I will check it the next days, I was „slightly“ busy the last two 
> weeks. @Chris-Rowland <https://github.com/Chris-Rowland>: Have you 
> checked the MF with changed polling period? We had problems recently 
> with either long or short polling intervals.
>
What sort of polling periods? And what sort of problems?

Chris

Chris-Rowland:
On 19/12/2019 11:01, sterne-jaeger wrote:
>
> I will check it the next days, I was „slightly“ busy the last two 
> weeks. @Chris-Rowland <https://github.com/Chris-Rowland>: Have you 
> checked the MF with changed polling period? We had problems recently 
> with either long or short polling intervals.
>
Found #1046, It seems, rather to my surprise, to be OK even with the 
maximum polling period of 60 seconds, which is about the time it takes 
for a meridian flip.  It's also OK at 10ms although the log file gets 
big rather rapidly.

Would have liked to reduce the maximum polling period to something 
sensible, like 10 seconds but this seems to be private to DefaultDevice.

Chris

Chris-Rowland:
Another few questions...

Should the capability flag bits follow the SimulatePierSideSP switch?  
if Simulate PierSide is off then should TELESCOPE_PIER_SIDE_SIMULATION 
be zero?
Similarly for TELESCOPE_PIER_SIDE, should that be zero in the simulator 
if simulation is off?

As it is we have three ways to show the pier side simulation state, the 
capability flags, the switch state and m_simulatePierSide.  It's 
confusing which of these apply and if they should be different.

And how do clients know if pier side is available? I can't see an 
obvious link to the capability flags. Should the driver report 
PIER_UNKNOWN if pier side is not available?

Chris

drivers/telescope/telescope_simulator.cpp
```diff
@@ -366,49 +366,65 @@ bool ScopeSim::ReadScopeStatus()
...
-                dx = fabs(dx);
-                if (dx == 0)
+                // set deltaRa according to the target pier side so that the
+                // slew is away from the meridian until the direction to go is towards
+                // the target.
+                switch (currentPierSide)
                 {
-                    // first step eastward if we start the meridian flip
-                    // ensure that the first step is bigger than the standard step to ensure flipping
-                    dx    = dt+0.5;
```
sterne-jaeger: This must not get lost with the change, since we need to ensure that with the first step fabs(deltaRa) * 15. > da_ra. If that's not the case, the simulator will not send the simulator in a 360° loop westward but will immediately jump to the target location without showing the behaviour of a meridian flip.

Test with 250ms polling period in the INDI Options tab to observe this behaviour.

knro:
Clients would know by the state of the property and which switches are on/off. We should probably document this to make it clear for the clients taking advantage of such property.

Chris-Rowland:
If you look at the code you will see that the pier flip slew has been 
totally changed.  The difference between the current and target Ra - 
deltaRa is set to a value that forces a move away from the meridian and 
this remains the case until deltaRa is large and towards the target:

             // In meridian flip, move to the position by doing a full 
rotation
             if (forceMeridianFlip)
             {
                 // set deltaRa according to the target pier side so 
that the
                 // slew is away from the meridian until the direction 
to go is towards
                 // the target.
                 switch (currentPierSide)
                 {
                 case PIER_EAST:
                     // force Ra move direction to be positive, i.e. to 
the West,
                     // until it is large and positive
                     if (deltaRa < 9)
                         deltaRa = 1;
                     else
                         forceMeridianFlip = false;
                     break;
                 case PIER_WEST:
                     // force Ra move direction to be negative, i.e. East,
                     // until it is large and negative
                     if (deltaRa > -9)
                         deltaRa = -1;
                     else
                         forceMeridianFlip = false;
                     break;
                 case PIER_UNKNOWN:
                     break;
                 }

             }

Without this for PierEast deltaRa would be small and negative, giving a 
move directly to the targetRa, with this deltaRa is forced to be 
positive so moving away from the meridian.  It will decrease until it 
reaches -12hrs where the shortest direction will be in the required 
direction and deltaRa becomes +12hrs so commanding s move in the correct 
direction. I continue forcing the slew until deltaRa is less than +9 
hrs, then turn off the forcing and letting the slew finish normally.  
Similarly for the flip in the other direction.  If the pier side is 
unknown then the slew will always be normal, this is how a fork mount 
will behave.  This is OK for poll intervals down to 10ms and up to 60s 
although in this case it is difficult to tell what direction the move is 
because it happens in one 60 second interval.

If there are still issues with timing the best way will be to use an 
internal control of the update time instead of using the poll interval.

Chris

Chris-Rowland:
On 20/12/2019 07:06, Jasem Mutlaq wrote:
>
> Clients would know by the state of the property and which switches are 
> on/off. We should probably document this to make it clear for the 
> clients taking advantage of such property.
>
By 'Property' and 'switches' you mean INDI properties and switches I 
suppose.  I can see that these can be sent to a client but the telescope 
capability is a series of bits, not INDI switches, and there doesn't 
seem to be an obvious way for these to get to a client.  I'm looking at 
kstars/kstars/indi/inditelescope.cpp.

All I can see to do is use the TELESCOPE_PIER_SIDE switch (NOT the 
capability) and if this returns PIER_UNKNOWN then a meridian flip is 
never required.

It this OK? Is it what you - and clients will expect?

I'll make sure that the simulator implements that behaviour, if simulate 
pier side is off then PierSide will always be PIER_UNKNOWN.  I'll also 
check the Celestron driver to ensure that it respects that behaviour for 
fork mounts.

Chris

PR #1062, opened 2020-01-09
Fix Version read in MoonLite focuser

Chris-Rowland:
There's a problem in at least some of the Moonlite focuser firmware versions where the ':GV#' command reply does not send the '#' terminator. This fix resolves this by asking for two characters rather than looking for the terminator. Other commands are not changed.
Tested by running on my Moonlite DRO box.


knro:
Thanks! Looks like other files were included in the PR by mistake as well.

drivers/focuser/moonlite.cpp
```diff
@@ -616,7 +614,16 @@ bool MoonLite::sendCommand(const char * cmd, char * res, bool silent)
     if (res == nullptr)
         return true;
 
-    if ((rc = tty_nread_section(PortFD, res, ML_RES, ML_DEL, ML_TIMEOUT, &nbytes_read)) != TTY_OK)
```
TallFurryMan: So you are considering the GV command returns a fixed-length block of data of size 2, then the remainder is flushed. OK for me as far as I know about the Moonlite protocol. Rest of the file is OK too.

Chris-Rowland:
I think they are files which have been updated since the last time I 
updated my repository.  They don't appear in the merge to your master.

Chris-Rowland:
On 09/01/2020 13:13, TallFurryMan wrote:
>
> *@TallFurryMan* commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In drivers/focuser/moonlite.cpp 
> <https://github.com/indilib/indi/pull/1062#discussion_r364730449>:
>
> > @@ -616,7 +614,16 @@ bool MoonLite::sendCommand(const char * cmd, char * res, bool silent)
>       if (res == nullptr)
>           return true;
>   
> -    if ((rc = tty_nread_section(PortFD, res, ML_RES, ML_DEL, ML_TIMEOUT, &nbytes_read)) != TTY_OK)
>
> So you are considering the GV command returns a fixed-length block of 
> data of size 2, then the remainder is flushed. OK for me as far as I 
> know about the Moonlite protocol. Rest of the file is OK too.
>
The MoonLite ASCOM driver has this fix, it seems that the terminator is 
missed from the version reply, at least in some firmware versions.  I 
could probably find the email conversation with Ron Newman but it may 
take some finding as this was done abut 2013.  The ASCOM sources are 
perfectly clear about it.

Chris

knro:
4 files appear in the merge.. I tried to "revert" 2 of them that were included by mistake but that didn't work so you need to remove them somehow.

knro:
@Chris-Rowland Can you please follow the example here to remove inditelescope.cpp and telescope_simulator.cpp from the PR?

https://stackoverflow.com/questions/38743912/remove-a-file-from-a-git-pull-request

Chris-Rowland:
Thanks, I've tried that.  If it doesn't work I'll try this:
https://xkcd.com/1597/

knro:
hehe it's alright, I just pushed your fix manually: da7b39786dfe11592887c1789e1c58112cd5b73b

PR #1063, opened 2020-01-12
TeenAstro INDI Device driver

fdesvallees:
This is the initial release of the TeenAstro INDI device driver. Documentation is attached.
[TeenAstro INDI device driver.pdf](https://github.com/indilib/indi/files/4050487/TeenAstro.INDI.device.driver.pdf)


knro:
Two files were added in the wrong place (CMakelists.txt and drivers.xml). Please correct those. Also, can you attach the documentation in plain text as well?

fdesvallees:
removed because of errors in file locations. New pull request issued

PR #1066, opened 2020-01-14
[INDI common]

iliaplatone:
Added functions in common routine source.
* Added photon flux calculation
* Added Planck constant and Lumen value #defines
* Updated absolute magnitude and parallax estimations

### DIFF
--- a/libs/indicom.c
+++ b/libs/indicom.c
@@ -1641,9 +1641,21 @@ double estimate_field_rotation(double HA, double rate)
     return HA;
 }
 
-double parsec2m(double parsec)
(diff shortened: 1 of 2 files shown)


TallFurryMan:
👍 For some reason I always associate indicom with communication utilities, which made your changes weird 😉

knro:
@TallFurryMan indicom stands for INDI Common, so not necessairly communications. I actually moved communication out to https://github.com/indilib/indi/tree/master/libs/indibase/connectionplugins

but never went through with migrating drivers away from indicom when it comes to serial communication.

PR #1067, opened 2020-01-14
Sensors

iliaplatone:
Created SensorInterface and Spectrograph, Detector subclasses.
TODO: work in indi-3rdparty to conform with this new standard.
The sensor interface more represents the mean of single-dimensional blob
devices.
The first implemented drivers will be spectrographs and radio receivers
as spectrograph since sampling time can represent spectra in both cases
and continuum can be done both in 1element samples and chunks or even
spectra.
The BLOB management is done in the base interface class. Subclasses
should implement properties or further BLOBs by adding methods or
functionalities into themselves.
Helper methods also should be present into child classes only.


knro:
This is a very *welcome* change, thanks Ilia!

I feel we ought to perhaps provide some documentation for prospective developers about this so they can use it as a starting point. You think Radio sim is sufficient? maybe update the README there?

iliaplatone:
I have to check out how to manage detectors. These kind of devices output pulses so I guess to use BLOBs which use the single samples as timestamps. Kind of detectors are cosmic ray revealers or photon counting devices. Also correlators may use this class. Once a solution will be find for this I can write some documentation about these classes.

jpaana:
I now submitted a PR for RTLSDR (https://github.com/indilib/indi-3rdparty/pull/50), couldn't add iliaplatone as reviewer for some reason, probably because it's a different repo.

PR #1071, opened 2020-01-17
TeenAstro INDI device driver - version 0.9 

fdesvallees:


knro:
Looks like there is some CI issue to fix. For the write method, simply use int rc = write ..etc
followed by INDI_UNUSED(rc);

fdesvallees:
Fixed compile warning

PR #1074, opened 2020-01-20
    [DSP] Added new plugin for Signal processing

iliaplatone:
This is applicable on BLOBs of INDI::SensorInterface children or
    INDI::CCD classes.
    Must implement DSP directly into SensorInterface (and CCD if agreed) or
    drivers should have this implemented when developers decide?
    
    Features:
    - Buffer transformations like DFT or Laplace, Wavelets.
    - Convolutions (Lowpass, Highpass filters), downloaded BLOBs from clients
    can be used as convolution matrices.
    - Primarily written for SensorInterface, can be applied into CCD, for
    star or object recognition for example.
    - Creates new BLOBs and uploads them to the client separately from the
    main BLOB.
    
    TODO:
    - Is download possible from KStars (BLOB write-only and FITS/Picture
    parsers)? - Also useful for Lime transmitter.
    - Cross and auto correlation.


knro:
Regarding BLOBs download... can this be used in some simulator to check out?

iliaplatone:
I didn't add it into any driver.. quickly into CCD simulator should be a good candidate. I have to write some documentation.

PR #1075, opened 2020-01-22
Dsp

iliaplatone:


iliaplatone:
@knro I can't upload the matrix data, a BLOB for the convolution plugin. Can you help me please?
The file is libs/indibase/dsp/convolution.cpp
I put a draft for a fits parser into ISNewBLOB there.

knro:
@iliaplatone matrix data is in that file? What error do you git when uploading?

iliaplatone:
ISNewBLOB simply gets not called

iliaplatone:
Also wavelets values are not updated

knro:
Ok, can you tell me what steps do I need to replicate/test this issue with ISNewBlob and the wavelet values?

iliaplatone:
BLOB upload:
You have to start CCD Simulator and enable, under signal processing tab, convolution.
This enables a RW BLOB, a small image to convolute with the whole image received from camera.
Since file parsing is not tested neither complete I couldn't test about crashes etc..
Wavelets:
When enabled wavelets in Signal Processing I can't change values. This also uses convolution from libdsp so It will be slow maybe

libs/indibase/indiccd.cpp
```diff
@@ -1895,6 +1928,12 @@ bool CCD::ExposureComplete(CCDChip * targetChip)
     // Reset POLLMS to default value
     POLLMS = getPollingPeriod();
 
+    if(HasDSP()) {
+        uint8_t* buf = (uint8_t*)malloc(PrimaryCCD.getFrameBufferSize());
```
knro: Use static_cast<uint8_t*>(...)

libs/indibase/dsp/manager.h
```diff
@@ -0,0 +1,70 @@
+﻿/*******************************************************************************
+  Copyright(c) 2017 Jasem Mutlaq. All rights reserved.
+
+ Connection Plugin Interface
```
knro: Correct header.

libs/indibase/dsp/manager.cpp
```diff
@@ -0,0 +1,149 @@
...
+    convolution = new Convolution(dev);
+    transforms = new Transforms(dev);
+    spectrum = new Spectrum(dev);
+    histogram = new Histogram(dev);
+    wavelets = new Wavelets(dev);
+}
+
+Manager::~Manager()
+{
+    convolution->~Convolution();
```
knro: Why is calling the destructor explicitly required?

libs/indibase/dsp/dspinterface.h
```diff
@@ -78,29 +85,24 @@ class Interface
...
     private:
-                double primaryFocalLength, primaryAperture;
-        double RA, Dec;
-        double Lat, Lon, El;
-        double MPSAS;
-        char processedFileName[MAXINDINAME];
-        char processedFileExtension[MAXINDIFORMAT];
-        int BufferSizesQty;
-        int *BufferSizes;
+        long BufferSizesQty;
```
knro: If you can use uint32_t instead of long, that would be better.

knro:
ok is there an example of what kind of small image I can use? and what output to expect?

knro:
So in ISNewBLOB(..) in ccd_simulator.cpp, you simply need to change the function to say this:

ccdSim->ISNewBLOB(...) this way it gets passed and then propagated upwards

knro:
Can you please also add the spectrograph simulator to drivers.xml?

iliaplatone:
Try with  a small FIT (10x10 pixels for example) with a B/W star shape. Each star should output with this shape aberration.

PR #1081, opened 2020-01-30
Freebsd 12.0 compilation fixes and webcamd support

andkem:
This makes libindi compile on FreeBSD 12.0. It also adds support for webcamd. webcamd is essentially a a port Video 4 Linux to FreeBSD user space. Using a few defines, it is possible to compile the indi support for v4l2 for FreeBSD.

I don't know if it still compiled on earlier FreeBSD versions, but I did not try to maintain backwards compatibility, meaning I haven't tried compiling it on anything other than 12.0.

I tested the v4l2 support by forwarding the client over SSH to my Linux machine running OpenPHD2 and could capture images from the FreeBSD machine.

A build and test run has been performed on Linux and it still seems to work fine.


knro:
Thank you for making INDI works on FreeBSD!

PR #1082, opened 2020-02-01
Indicom

iliaplatone:
[indicom/INDICorrelator]
Draft for interferometers/correlators.
This class is useful for imaging using two radio dishes or more by connecting them to a correlator which has an INDI driver.
I will write a sample firmware for FPGAs in future if needed.


knro:
Thank you! I do hope we get to see all of this used in a driver soon!

PR #1096, opened 2020-02-27
Fix Guide problem with StarSense

Chris-Rowland:
StarSense HCs return an additional track mode, Ra and Dec.  This now handles this and allows guiding in this mode.


Chris-Rowland:
The build errors seem to be assoiated with the rainbow telescope driver.

Chris-Rowland:
I've tried several things to fix the set datetime error and my changes don't seem to make it into the test code.  Not sure what to do next.

PR #1098, opened 2020-03-12
Ensure we are doing signed integer math on clamping

rickbassham:
Previously, when focusing inward, -1 * the unsigned integer would wrap around to a large number, which when added to the current position would again wrap around to a large negative number. The result is that doing a relative focus inward would always get clamped to 0, racking the focuser all the way in.

### DIFF
--- a/drivers/focuser/moonlite.cpp
+++ b/drivers/focuser/moonlite.cpp
@@ -522,7 +522,7 @@ IPState MoonLite::MoveAbsFocuser(uint32_t targetTicks)
 IPState MoonLite::MoveRelFocuser(FocusDirection dir, uint32_t ticks)
 {
     // Clamp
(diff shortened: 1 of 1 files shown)


TallFurryMan:
Agreed. In reference to patch, this code is presenting the problem:

```
#include <cstdint>
#include <iostream>

using namespace std;

int main()
{
    double value = 19200.0;
    uint32_t ticks = 200;
    int dir = -1;
    int32_t newpos1 = value + dir*ticks;
    int32_t newpos2 = static_cast<int32_t>(value) + dir*static_cast<int32_t>(ticks);
    cout << "N1:" << newpos1 << endl;
    cout << "N2:" << newpos2 << endl;
}
``` 

If `d = +1`:
```
N1:19400
N2:19400
```

If `d = -1`
```
N1:-2147483648
N2:19000
```

TallFurryMan:
OK with the altered patch merged.

PR #1100, opened 2020-03-13
Tests for the CCD simulator

TallFurryMan:
A few preliminary tests for the CCD simulator.

This was part of an earlier activity to introduce aperture-based PSF star rendering, in order to implement a rough flatness and sensor centering utility for KStars Ekos.

This PR adds property tests, guide API tests and rendering tests. It's nowhere near a complete test plan, but allows to find that:
- Frame buffer is not reset just after setting parameters up.
- RA/DEC fields are undefined after instantiating the driver.


knro:
Great!! I will give it a go shortly and report back.

TallFurryMan:
Force-updated the PR to allow proper merge. No code change.

knro:
is this still WIP?

TallFurryMan:
Ah no this is final, sorry for not reminding you :)

PR #1103, opened 2020-03-15
added hour axis offset

wildi:
Hello Jasem

I added the the hour axis offset to the ccd simulator.

Most of the code is only in effect if the angle (CP,TCP) > 0, with CP: celestial, TCP: HA axis pole.

I derived from the center locations of the simulated images the offset of the hour axis and they are in good agreement. I'm investigating the differences further.
 
Kind regards, wildi

E.g. I set lamnda to .8 deg and theta to 90. deg (theta denotes the great circle, where the HA axis lies in), results:

King h         :         90.21 degree
King lambda    :       2803.13 arcsec
King lambda    :          0.78 deg


knro:
Thanks! What's the use case exactly for this? Also, what's wrong with debug logging? I just checked and it works fine here.

PR #1106, opened 2020-03-24
Debug/implement new Xagyl driver

nadalle:
This patch set fixes all issues I could find with Jasem's new Xagyl driver, when running against my physical wheel.

Major features:
1. Strict command/response with no timers or tcflush. The device firmware crashes on me when it receives commands out of sequence.
2. Relative commands like Jitter/Offset/etc. are implemented using loops.
3. Device error messages are reported to the user.
4. Device reboot/etc. work correctly and poll the device.

I'm not sure if offset etc. ever worked with the old driver (the device's command set seems to imply they're meant to be used interactively!) but they work now.

My device cannot use the v3+ Xagyl firmware, so I cannot verify that the Pulse Width command works as intended.

### DIFF
--- a/drivers/filter_wheel/xagyl_wheel.cpp
+++ b/drivers/filter_wheel/xagyl_wheel.cpp
@@ -466,7 +466,7 @@ bool XAGYLWheel::SelectFilter(int f)
     if (!optionalResponse(opt))
         return false;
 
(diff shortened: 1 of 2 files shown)


knro:
Please update astyle to max 124 max code length:

--max-code-length=124

PR #1108, opened 2020-03-28
Added again the addition of FOCUSPOS to the FITS header

joxda:
Apparently the code to actually write the focus position to the FITS header was lost, this PR is to add it back in.


libs/indibase/indiccd.cpp
```diff
@@ -1811,6 +1809,12 @@ void CCD::addFITSKeywords(fitsfile * fptr, CCDChip * targetChip)
         fits_update_key_dbl(fptr, "ROTATANG", RotatorAngle, 3, "Rotator angle in degrees", &status);
     }
 
+    // JJ ed 2020-03-28
```
TallFurryMan: Proper comment would be better here I think.

joxda: Added

PR #1113, opened 2020-04-02
Weather: adds checkParameterState()

pardo-bsso:
This comes from the discussion at https://github.com/indilib/indi-3rdparty/issues/83

This extracts the check of a parameter (added with addParameter) against
the user defined limits.

    IPState checkParameterState(std::string param);
    IPState checkParameterState(INumber parameter);

Returns:

  - IPS_IDLE:  The given parameter name is not valid.
  - IPS_OK:    The given parameter is within the safe zone.
  - IPS_BUSY:  The given parameter is in the warning zone.
  - IPS_ALERT: The given parameter is in the danger zone.

### DIFF
--- a/libs/indibase/indiweatherinterface.cpp
+++ b/libs/indibase/indiweatherinterface.cpp
@@ -187,6 +187,43 @@ bool WeatherInterface::setCriticalParameter(std::string param)
     return false;
 }
 
+IPState WeatherInterface::checkParameterState(INumber parameter)
+{
+    double warn = *(static_cast<double *>(parameter.aux0));
+    double rangeWarn = (parameter.max - parameter.min) * (warn / 100);
+
+    if ((parameter.value < parameter.min) || (parameter.value > parameter.max))
+    {
+        return IPS_ALERT;
+    }
+    //else if (parameter.value < (parameter.min + rangeWarn) || parameter.value > (parameter.max - rangeWarn))
+    // FIXME This is a hack to prevent warnings parameters which minimum values are zero (e.g. Wind)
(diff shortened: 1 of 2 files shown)


libs/indibase/indiweatherinterface.cpp
```diff
@@ -187,6 +187,43 @@ bool WeatherInterface::setCriticalParameter(std::string param)
     return false;
 }
 
+IPState WeatherInterface::checkParameterState(INumber parameter)
```
knro, line 190: IMO, this should be

IPState checkParameterState(const INumber *parameter) const

libs/indibase/indiweatherinterface.cpp
```diff
@@ -187,6 +187,43 @@ bool WeatherInterface::setCriticalParameter(std::string param)
...
+    }
+    else
+    {
+        return IPS_OK;
+    }
+
+    return IPS_IDLE;
+}
+
+IPState WeatherInterface::checkParameterState(std::string param)
```
knro, line 214: same as above, IPState checkParameterState(const std::string &param) const

libs/indibase/indiweatherinterface.cpp
pardo-bsso: Hi, thanks for the review.

Making both methods const is a nice addon as a safety measure against unintended changes in internal state.

But I'm not sure I understand why passing the INumber as a pointer and not a reference like the string version is desireable.

knro: You can also pass by reference.

pardo-bsso: Thanks.
I updated this pull request with those changes.

PR #1116, opened 2020-04-10
AP drivers, undefined park position in HA/Dec coordinate system

wildi:
Hello Jasem

I corrected and tested the changes to the PARK3 position. The person who tested it was INDI forum member Spartacus, the person who originally reported the bug.

Besides I added 0.1 degree to park position alt = latitude + .1 I corrected the call of calcParkPosition and added isSimulation() where necessary.

The addition .1 degree is a kind of a minimal fix to guarantee that the mount is more or less at HA = -6 hours. Transforming Alt=latitude, Az=0, 180 deg to HA/Dec coordinate system yields not a defined HA angle. This is the source of the error.

Currently all park positions are defined within Alt/Az coordinate system. Since AP mounts are all equatorials this is introduces an unnecessary intermediate step, namely delegating the calculation of the HA/Dec coordinates to the AP GTOCPX controllers. That can be better done within the INDI driver.

Kind regards, wildi

### DIFF
--- a/drivers/telescope/lx200ap_experimental.cpp
+++ b/drivers/telescope/lx200ap_experimental.cpp
@@ -1366,15 +1366,6 @@ bool LX200AstroPhysicsExperimental::Park()
         ln_equ_posn equatorialPos;
 
         ln_get_equ_from_hrz(&horizontalPos, &observer, ln_get_julian_from_sys(), &equatorialPos);
-
-        double lst;
-        double lha;
-        lst = get_local_sidereal_time(LocationN[LOCATION_LONGITUDE].value);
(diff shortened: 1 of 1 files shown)


knro:
@mikefulb Can you please check?

mikefulb:
I don't really have time to review or test the driver unfortunately.  I 
don't have a good INDI development setup to build code on.

The park positions were working for me when I was using the driver but 
that was over a year ago since I last was using it.

AP uses alt/az park position definitions and I based my work on 
conversations with Howard@AP.   Might be good to call him and chat about 
your changes.

Michael

PR #1117, opened 2020-04-15
Change Telescope SImulator to simulate mount axes

Chris-Rowland:
This is a large change to the telescope simulator.  The mount simulation has been totally reworked so that it uses Angle, Axis and Alignment classes, all coded in the simscope_helper files.
Angle provides a standard way to manage angles, handling the wrap around a turn and implementing arithmetic and logic operators that handle this.  Almost all of the checks for angle ranges vanish inside this.
Axis implements a generic axis that can be used for both mount axes, providing slew and move properties and methods.
Alignment handles the conversions between mount and observed coordinates, including managing a GEM mount and applying a set of mount model terms.  This is based on Patric Wallace's paper on telescope pointing.
The simulator has been refactored to use these classes.  Initially only to support the current mount type, a GEM with Pier Side but the additional features - suporting other mount types, making reporting the pier side optional, adding more tracking rates, can be implemented once this is released.


knro:
Thank you Chris, this is a great addition to INDI! 

@TallFurryMan @sterne-jaeger any comments?

sterne-jaeger:
I'll take a closer look on it over the weekend. Since we are changing a very central class, we need to ensure that we don't get regressions - especially for bug fixing and further development.

Generally spoken, the approach looks good and I really appreciate that we extend the simulator so that we can test more situations that before. Much appreciated, Chris!

Chris-Rowland:
Thanks,  agree it needs testing carefully, however it does only affect 
the simulator.  There's no change to code used with real mounts.

In some ways I wonder if I've gone far enough:
  - The angle class will I think be useful for everyone using angles, it 
may be worth making this a library function once it's found to be good.
  - It may have helped the readability if the Axis and Alignment classes 
were encapsulated into a Mount class which separated the mount logic 
from the INDI logic.

Chris

knro:
@Chris-Rowland  Another interesting point is this: Can we use this to test and fine-tune the INDI Alignment Subsystem?

Chris-Rowland:
On 17/04/2020 09:55, Jasem Mutlaq wrote:
>
> @Chris-Rowland <https://github.com/Chris-Rowland> Another interesting 
> point is this: Can we use this to test and fine-tune the INDI 
> Alignment Subsystem?
>
Hope so.  That was part of the reason that I didn't use the INDI 
Alignment subsystem.

Of course if there are differences the question will be which is correct?

Chris

sterne-jaeger:
I did a first check, everything seems to work as before. I was a little bit confused first, since I expected a new tab with all the simulation properties - which I knew from from an intermediate version and is excluded currently.

So in a nutshell the INDI client view is identical to the old simulator except that the option "Simulate pier side" is no longer available. I know, it will come back with the simulator tab. Fine for me now.

I found one bug, but this was a bug in the old version of the simulator as well. As far as I understand it, the pier side calculation is only right for the SE and SW quadrants of the sky. NE and NW have to be flipped. From my understanding, the pier side should be
pier side east: pointing SW and NE
pier side west: pointing SE and NW

From my perspective, we could merge it into the master branch.

Wolfgang

Chris-Rowland:
On 18/04/2020 08:32, sterne-jaeger wrote:
>
> I did a first check, everything seems to work as before. I was a 
> little bit confused first, since I expected a new tab with all the 
> simulation properties - which I knew from from an intermediate version 
> and is excluded currently.
>
I thought you asked for the implementation to be simplified - no 
multiple mount types, no alignment.  That's why I removed the simulator 
tab.  I'm planning to put it back almost as soon as this is out but 
could do so now if you and Jared prefer.
>
> So in a nutshell the INDI client view is identical to the old 
> simulator except that the option "Simulate pier side" is no longer 
> available. I know, it will come back with the simulator tab. Fine for 
> me now.
>
This went with the removal of the Simulator tab and will return when 
that does.
>
> I found one bug, but this was a bug in the old version of the 
> simulator as well. As far as I understand it, the pier side 
> calculation is only right for the SE and SW quadrants of the sky. NE 
> and NW have to be flipped. From my understanding, the pier side should be
> pier side east: pointing SW and NE
> pier side west: pointing SE and NW
>
No.  The way it works is correct, it's your understanding that is 
incorrect. East should be SW and NW and West SE and NE.

Pier side is a bad name, it should be called Pointing State with states 
being Normal and Inverted.  All the pointing state depends on is the 
declination axis position, normal for the range -90 to 0 to +90 and 
inverted for the range +90 to 180 to 270/-90.

sterne-jaeger:
> I thought you asked for the implementation to be simplified - no multiple mount types, no alignment.  That's why I removed the simulator tab.  I'm planning to put it back almost as soon as this is out but could do so now if you and Jared prefer.

I know, I know! Just told the story that I fiddled around a while until I found out that it was the new one running and not out of odd reasons the old one. Definitely my fault. It was the first time that I had to check out a PR, so I thought first I did something wrong with git.

Regarding the pier side: my understanding was always that we are talking about the side of pier where the scope is positioned when pointing to a certain position. That's why I claimed it being a bug. But as already said, let's leave it as it is and discuss it afterwards. Let's move forward.

sterne-jaeger:
Sorry folks, I checked it again with the old simulator, the new pier side calculation is wrong and the old simulator behaved differently (and right from my perspective).

Assuming being on the northern hemisphere and a GEM, the pier side and pointing direction need to be as follows:
pointing SW or NE: pier side EAST
pointing SE or NW: pier side WEST

Chris-Rowland:
In that case the old simulator was wrong.

I've checked with the new simulator and my AVX mount and they are both 
the same:
New Simulator:
Direction
	Hour Angle
	Azimuth
	Pier Side
	Primary Axis
	Secondary Axis
NE
	-7
	33
	W->E
	69
	118
SE
	-2
	129
	W->E
	154
	151
SW
	+2
	251
	flip! E->W
	25
	42
NW
	+5
	312
	E->W
	84
	58
NNW
	+10
	344
	E->W
	154
	52


AVX
Direction
	Hour Angle
	Azimuth
	Pier Side
NE
	-7
	34
	W->E
SE
	-1.5
	133
	W->E
SW
	+2
	254
	flip E->W
NW
	+6.5
	323
	E->W


And for me the old simulator shows:
Direction
	Hour Angle
	Azimuth
	Pier Side
NE
	-6.75
	37
	W->E
SE
	-1
	144
	W->E
SW
	+2
	261
	flip E->W
NW
	+5.5
	323
	E->W
NNW
	+10
	346
	E->W


They are all the same and as far as I'm concerned all correct, including 
the old simulator.  The only flip change happens with the move from the 
east to the west and is there to ensure that the OTA stays above the 
mount so it can't collide.

Wolfgang, I know the name is confusing but it really is that the pier 
side only depends on the dec axis value.  The physical position of the 
OTA relative to the mount is something different.

It is critical to get this right because without this scope and dome 
synchronising can't be done properly, neither can guiding using PHD2, 
nor any sort of mount management that relies on knowing the pointing state.

I can't proceed with the changes to Ekos until this is resolved.

Chris

sterne-jaeger:
Chris, I'm deeply sorry. You are right, the pier side is working exactly as you implemented it. I cross checked it now physically with my own mount, it works the same way.

So everything is fine, we can continue with your other changes.

Chris-Rowland:
Phew! Thanks.  At least your mount is correct, there's someone on the 
forum with a Losmandy mount that isn't.

Chris

PR #1123, opened 2020-04-24
Fix telescope Simulator

Chris-Rowland:
This fixes the bug with motion control.
Also abort now stops all motion.
Solar and lunar tracking rates added.
The simulation tab is still disabled.


knro:
You think the simulation tab is ready or needs more time?

Chris-Rowland:
On 24/04/2020 19:59, Jasem Mutlaq wrote:
>
> You think the simulation tab is ready or needs more time?
>
I'm enabling and using it pretty routinely.  What I can say is that 
setting alignment values changes things but it's difficult to see 
exactly what effect they have without more feedback.  Sync now works, 
even with alignment parameters.

All that's needed is to define USE_SIM_TAB.  It will be useful if people 
do so and provide feedback.  Maybe the best thing is to publish the code 
as is.  People who don't care will get the non lignment version and 
those who do will probably be compiling from sources and can enable 
this.  i could push a version with #define USE_SIM_TAB commented out if 
it helps.

Chris

PR #1127, opened 2020-04-26
Sensors

iliaplatone:


iliaplatone:
@knro please wait before merging, thanks.

iliaplatone:
This can be merged now.

iliaplatone:
There's no problem about baud rates, but I need a way to store high precision timestamps, expecially for correlators and receivers like rtl-sdr and limesdr. Some limesdr devices have separate clock inputs and should require at least sysclock tick timing resolution.
Maybe a custom FIT entry?
Let me know about this please.

knro:
FITS entry would work fine in this case. I presume you don't want high precision timestamp for the logs per se?

iliaplatone:
Perfect! I need them in post-processing. Which keyword should be used?

knro:
Well, I'd leave this up to you. We cannot use DATE-OBS since that is the standard keyword, so anything else is OK. Perhaps they already have some keyword for this in IRAF or other packages?

iliaplatone:
What about DATE? Is it used anywhere else?

knro:
how about TIMESTAMP?

iliaplatone:
Yes, Timestamp is more appropriate, thank you, I made a pull request, please check it out, sorry.

PR #1131, opened 2020-04-29
Sensors

iliaplatone:


libs/indibase/connectionplugins/connectionserial.h
```diff
@@ -42,7 +42,7 @@ class Serial : public Interface
      * \brief Supported baud rates
      * \note: Default baud rate is 9600. To change default baud rate, use setDefaultBaudrate(..) function.
      */
-    typedef enum { B_9600, B_19200, B_38400, B_57600, B_115200, B_230400 } BaudRate;
+    typedef enum { B_9600, B_19200, B_38400, B_57600, B_115200, B_230400, B_460800, B_576000, B_921600 } BaudRate;
```
knro, line 45: Ok hopefully these baud rates would be OK for MacOS.

libs/indicom.c
```diff
@@ -315,16 +317,39 @@ void IDLog(const char *fmt, ...)
...
 {
     static char ts[32];
+    char iso8601[32];
     struct tm *tp;
-    time_t t;
+    struct timespec tm;
+
+    timespec_get(&tm, TIME_UTC);
+    time_t t = (time_t)tm.tv_sec;
+    long n    = tm.tv_nsec % 1000000000;
```
knro, line 347: the regular timestamp function should not be changed as it would break the logs.

libs/indibase/connectionplugins/connectionserial.h
iliaplatone: Yes, but I just want to be sure, unfortunately these were changed for tests and forgot to not include them into the commit.

libs/indicom.c
iliaplatone: this is done

PR #1137, opened 2020-05-03
Update CCD and Guide simulators to improve image accuracy

Chris-Rowland:
The conversion of the JNow position to a J2000 catalogue position for
the GSC catalogue was  not done correctly, only the Precession was
removed, leaving aberration and nutation.

This adds libastro.cpp,.h files which contain converters to and from
J2000 catalogue and apparent positions.  When converting apparent to
J2000 the aberration, nutation and precession are all removed.  These
converters may be useful elsewhere, this isn't the only place
conversion from aparrent to J2000 is needed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -185,6 +185,7 @@ IF (UNITY_BUILD)
 ENDIF ()
 
 SET(indiclient_C_SRC ${indiclient_C_SRC}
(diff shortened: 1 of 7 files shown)


TallFurryMan:
Chris, you need to state that you distribute your libastro files under the same, or more permissive, license as the rest of the indi repository.

PR #1140, opened 2020-05-07
Sensors

iliaplatone:


knro:
So what the rationale behind this change? some drivers start not supporting streaming then they enable it later?

iliaplatone:
This change is .. for coders because in some drivers like rtl-sdr the DSP (not Streaming) is being initialized into setCapabilities, and there's no initProperties in DSP manager... Just to make sure it has been initialized before usage, for coherence I made the same in Streamer, even if not meaningful.
I saw initialization of Streamer and DSP into initProperties, by the old way one had to instantiate DSP into the constructor instead.

PR #1148, opened 2020-05-17
Improve telescope Simulator mount Error handling

Chris-Rowland:
This implements the simulator polar align error handling using 3D rotates instead of trigonometric methods.  This works better close to and at the pole.

Extensive unit testing for this and the supporting Vector and Angle classes has been added.


Chris-Rowland:
I'm not sure how to resolve that issue, one compiler has decided that sqrt is not a memeber of std.

Issue #1097, opened 2020-03-01
skywatcherAltAzSimple crashes at serial connect PC Ubuntu 18.04


gubihu:
The driver exits after serial connect and the indiserver tries to restart it, without success.

That happens both the Ubuntu packaged version (INDI Library: 1.8.4, Code 1.8.4-tgz. Protocol 1.7.) and with freshly compiling the github HEAD version.

Strangely enough the same driver compiled from the same source works on Raspberry PI.

I have traced down and it seems there is a nullpointer exception in indidriver.c:IUUpdateSwitch(), it gets NULL for svp parameter sometimes (but not everytime)when called from SkywatcherAltAzSimple::ISNewSwitch()

I cannot find the root cause of the problem, but managed to hack it:

This hack seems to mitigate the problem:
[indidriver.c.diff.txt](https://github.com/indilib/indi/files/4272663/indidriver.c.diff.txt)

`
diff --git a/indidriver.c b/indidriver.c
index 64a3cd1..ee5ee88 100644
--- a/indidriver.c
+++ b/indidriver.c
@@ -230,6 +230,10 @@ int IUUpdateSwitch(ISwitchVectorProperty *svp, ISState *states, char *names[], i
     ISwitch *sp;
     char sn[MAXINDINAME];
 
+    if (svp == 0) {
+       IDLog("IUUpdateSwitch svp is NULL\n");
+       return -1;
+    }
     /* store On switch name */
     if (svp->r == ISR_1OFMANY)
     {
`

This one also works, and eliminates the trigger of the previos log:
[skywatcherAltAzSimple.cpp.diff.txt](https://github.com/indilib/indi/files/4272666/skywatcherAltAzSimple.cpp.diff.txt)

`
diff --git a/drivers/telescope/skywatcherAltAzSimple.cpp b/drivers/telescope/skywatcherAltAzSimple.cpp
index 1e8047a..212afdd 100644
--- a/drivers/telescope/skywatcherAltAzSimple.cpp
+++ b/drivers/telescope/skywatcherAltAzSimple.cpp
@@ -501,7 +502,12 @@ bool SkywatcherAltAzSimple::ISNewNumber(const char *dev, const char *name, doubl
 
 bool SkywatcherAltAzSimple::ISNewSwitch(const char *dev, const char *name, ISState *states, char *names[], int n)
 {
-    IUUpdateSwitch(getSwitch(name), states, names, n);
+       ISwitchVectorProperty *svp = getSwitch(name);
[...]

PR #1150, opened 2020-05-19
fix: IUUpdateSwitch crashes when ISwitchVectorProperty *svp is null p…

gubihu:
…ointer
This is for issue #1097.
It does not solve the root cause (why does not the swithc properties found in the first place?)
But anyways it is nicer to return from the function when getting a null pointer than crashing the driver.

### DIFF
--- a/indidriver.c
+++ b/indidriver.c
@@ -234,6 +234,10 @@ int IUUpdateSwitch(ISwitchVectorProperty *svp, ISState *states, char *names[], i
     ISwitch *sp;
     char sn[MAXINDINAME];
 
+    if (svp == 0) {
+    	IDLog("IUUpdateSwitch svp is NULL\n");
+    	return -1;
+    }
     /* store On switch name */
     if (svp->r == ISR_1OFMANY)
     {


indidriver.c
```diff
@@ -234,6 +234,10 @@ int IUUpdateSwitch(ISwitchVectorProperty *svp, ISState *states, char *names[], i
     ISwitch *sp;
     char sn[MAXINDINAME];
 
+    if (svp == 0) {
```
TallFurryMan, line 237: "nullptr" instead? Else looks good obviously, but we'd better find the root cause for sure.

knro:
I think a crash here is appropriate since the reason is in the calling function. It has to be figured out and fixed. This should not be ignored, maybe it should be **asserted**, but not just ignored.

gubihu:
I don't agree here. Blindly dereferencing a pointer and crashing is _never_ appropriate. Especially not when that pointer is somehow related to a user editable config file. Crash just causes user frustration and does not help diagnosing the root cause.
My proposal was not to ignore it, but report it to the LOG, and return an error return code from the function. That helps diagnosing the root cause, the diagnoser can see the log file (even user supplied log file), and that is also an appropriate place to put a break point at debug sessions.
Anyway I will make further effort to find the root cause, but I may need some help from someone who understand better the structure of ISwitchVectorProperty mechanism.

TallFurryMan:
I tend to agree with @knro, in the sense that getting an unexpected null pointer and hiding it is not serving any purpose (even if logged). But that means we should first log, then crash. So let's do a compromise here: let's log the issue, then assert on debug only, then return. What do you think?

knro:
@TallFurryMan Yes an assert on debug only with log would be a good solution.

gubihu:
Should I modify my code branch according to this, and reissue the pull request?
I need some time for that because I want to test before reissue...

indidriver.c
gubihu: nullprt is a C++11 construct. This is a pure C source file.

knro: Correct. assert should be used in such a case.

TallFurryMan: @gubihu agreed, sorry.

PR #1151, opened 2020-05-23
Fix skywatcher altaz crash

gubihu:
Reissuing PR #1150 (fix for bug #1097), now issuing assert() on debug builds.
Also prevent calling IUUpdateSwitch() with svp==nullprt from drivers/telescope/skywatcherAPIMount.cpp and drivers/telescope/skywatcherAltAzSimple.cpp.
The root cause is still unknown, but it is related to how .indi/*_congif.xml is created when .indi is empty. After one successful driver open/close the config.xml changes so that the error does not occur  anymore.


gubihu:
Thanks for merging.
I still want to find the root cause. If anyone could help, it would be much appreciated!

PR #1154, opened 2020-05-24
Logitech HD C270 webcam data added to v4l2_ccd driver

stroblhofwarte:
Add Logitech HD C270 USB camera to the v4l2 driver. Set pixel size to 2.8um.


TallFurryMan:
Looks good, thanks.

PR #1161, opened 2020-06-03
Some asserts and one memory error.

lboclboc:
Hi, found one real memory problem with uninitialized data in newLilXML that initParser can have problem with freeing pointers that might be randomly initialized.
Also added some asserts (only active in debug builds).
And some other rewrites that I hope are ok.

### DIFF
--- a/base64.h
+++ b/base64.h
@@ -24,6 +24,8 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110 - 1301  USA
 
 #pragma once
 
+#include <stddef.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -39,6 +41,11 @@ extern "C" {
     \param inlen number of bytes to convert
     \return 0 on success, -1 on failure.
  */
+extern int to64frombits_s(unsigned char *out, const unsigned char *in, int inlen, size_t outlen);
+
+#if defined(__GNUC__)
+__attribute__((deprecated("unsafe function, use to64frombits_s instead")))
+#endif
 extern int to64frombits(unsigned char *out, const unsigned char *in, int inlen);
 
 /** \brief Convert base64 to bytes array.
--- a/indiapi.h
+++ b/indiapi.h
@@ -24,6 +24,8 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110 - 1301  USA
 
 #pragma once
 
+#include <stdio.h>
+
 /** \mainpage Instrument Neutral Distributed Interface INDI
  *
 \section Introduction
@@ -495,3 +497,9 @@ typedef struct _IBLOBVectorProperty /* BLOB vector property descriptor */
  * with actual array, not pointer.
  */
 #define NARRAY(a) (sizeof(a) / sizeof(a[0]))
+
+/**
+ * @brief Bails out if memory pointer is 0. Prints file and function.
+ */
+#define assert_mem(p) if((p) == 0) { fprintf(stderr, "%s(%s): Failed to allocate memory\n", __FILE__, __func__); exit(1); }
+
--- a/indidriver.c
+++ b/indidriver.c
@@ -1774,6 +1774,7 @@ void IUSaveConfigBLOB(FILE *fp, const IBLOBVectorProperty *bvp)
 
         encblob        = malloc(4 * bp->bloblen / 3 + 4);
         l              = to64frombits(encblob, bp->blob, bp->bloblen);
+        assert(l <= (4 * bp->bloblen / 3 + 4));
         size_t written = 0;
 
         while ((int)written < l)
@@ -2317,6 +2318,7 @@ void IDSetBLOB(const IBLOBVectorProperty *bvp, const char *fmt, ...)
         {
             encblob = malloc(4 * bp->bloblen / 3 + 4);
             l       = to64frombits(encblob, bp->blob, bp->bloblen);
+            assert(l <= (4 * bp->bloblen / 3 + 4));
             printf("    enclen='%d'\n", l);
             printf("    format='%s'>\n", bp->format);
             size_t written = 0;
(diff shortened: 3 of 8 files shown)


indidriver.c
```diff
@@ -1774,6 +1774,7 @@ void IUSaveConfigBLOB(FILE *fp, const IBLOBVectorProperty *bvp)
 
         encblob        = malloc(4 * bp->bloblen / 3 + 4);
         l              = to64frombits(encblob, bp->blob, bp->bloblen);
+        assert(l <= (4 * bp->bloblen / 3 + 4));
```
TallFurryMan, line 1777: This assert (and the others you added) should part of the guarantees of `to64frombits()` and not a verification from the caller point of view. So move your `assert()` to that function.
Also, `encblob` could be returned as `null` by `malloc`, and that situation is not properly managed. Could you add a bailout path for this?

libs/lilxml.c
```diff
@@ -235,6 +237,7 @@ XMLEle **parseXMLChunk(LilXML *lp, char *buf, int size, char ynot[])
         if (!ltpos)
         {
             lp->ce->pcdata.s = (char *)moremem(lp->ce->pcdata.s, lp->ce->pcdata.sm + size);
+            assert(lp->ce->pcdata.s);
```
TallFurryMan, line 240: Getting a null pointer here is valid, could you add a bailout path instead?

libs/lilxml.c
```diff
@@ -211,8 +213,8 @@ void delXMLEle(XMLEle *ep)
 //#define WITH_MEMCHR
 XMLEle **parseXMLChunk(LilXML *lp, char *buf, int size, char ynot[])
 {
-    XMLEle **nodes = (XMLEle **)malloc(sizeof(XMLEle *));
-    int nnodes     = 1;
+    unsigned int nnodes     = 1;
+    XMLEle **nodes = (XMLEle **)malloc(nnodes * sizeof *nodes);
```
TallFurryMan, line 217: Could you add a check and bailout path here instead?

libs/lilxml.c
```diff
@@ -146,8 +147,9 @@ void lilxmlMalloc(void *(*newmalloc)(size_t size), void *(*newrealloc)(void *ptr
 /* pass back a fresh handle for use with our other functions */
 LilXML *newLilXML()
 {
-    LilXML *lp = (LilXML *)moremem(NULL, sizeof(LilXML));
-    memset(lp, 0, sizeof(LilXML));
+    LilXML *lp = (LilXML *)moremem(NULL, sizeof *lp);
+    memset(lp, 0, sizeof *lp);
+    assert(lp);
```
TallFurryMan, line 152: Your assert is executed too late. Instead, could you add a bailout path instead when moremem return null?

indidriver.c
lboclboc: I don't quite agree with that its to64frombits obligation to check memory size since it have no information about it. But your second point I will fix.

TallFurryMan: I completely agree on `to64frombits` not having the size information. I think that's a shortcoming of its prototype, like `memcpy` and `strcpy` had in the past. Now, there is `dlen` inside `to64frombits`, which does a allocation estimation and could be use for that purpose.
I propose you add a new prototype, `to64frombits_s`, that takes the size of the output buffer as argument. There you may assert on the size of the `encblob` destination. We can then deprecate the use of `to64frombits` progressively afterwards. What do you think?

TallFurryMan: The only rework in that case would be to call `to64frombits` from `to64frombits_s` after asserting the size of the destination using the same allocation length estimation done for `dlen`.

lboclboc: Sounds like a good idea, ill have a look

base64.h
```diff
@@ -39,6 +41,11 @@ extern "C" {
     \param inlen number of bytes to convert
     \return 0 on success, -1 on failure.
  */
+extern int to64frombits_s(unsigned char *out, const unsigned char *in, int inlen, size_t outlen);
+
+#if defined(__GNUC__)
+__attribute__((deprecated("unsafe function, use to64frombits_s instead")))
```
TallFurryMan, line 47: Excellent.

indiapi.h
```diff
@@ -495,3 +497,9 @@ typedef struct _IBLOBVectorProperty /* BLOB vector property descriptor */
  * with actual array, not pointer.
  */
 #define NARRAY(a) (sizeof(a) / sizeof(a[0]))
+
+/**
+ * @brief Bails out if memory pointer is 0. Prints file and function.
+ */
+#define assert_mem(p) if((p) == 0) { fprintf(stderr, "%s(%s): Failed to allocate memory\n", __FILE__, __func__); exit(1); }
```
TallFurryMan, line 504: Oh sorry, by bailing out I meant bailing out from the function, not from the program!

TallFurryMan: This is probably OK actually as it doesn't change the behavior, except when it avoids corruption. At some point the memory shortage should be managed by freeing resources that can be freed and retrying. Future!

TallFurryMan: One thing you can do is `assert(0); return;` depending on the prototype of the function you add your `assert_mem` into. With this, debug code asserts, and production code returns to the caller without damaging the memory.

lboclboc: Thanks, I agree. Its better to die hard so the user have a chance to restart instead of possibly wasting a lot of time on a system that behaves erratic.

knro: let me know once this is ready.

lboclboc: As far as I know it should be good now

lboclboc: Oh, I missed @TallFurrMan last comment.
I still think it's better to exit here since the code is not really written to handle memory allocation in general and would require quite a bit of coding. Standing in the middle of the night fotographing I would rather do a clean restart instead of having a half broken exposure sequence running.

lboclboc: Oh, I missed @TallFurrMan last comment.
I still think it's better to exit here since the code is not really written to handle memory allocation in general and would require quite a bit of coding. Standing in the middle of the night fotographing I would rather do a clean restart instead of having a half broken exposure sequence running.

PR #1163, opened 2020-06-06
Ap park3

wildi:


drivers/telescope/lx200ap_experimental.cpp
```diff
@@ -120,12 +122,13 @@ bool LX200AstroPhysicsExperimental::initProperties()
     IUFillSwitch(&SwapS[1], "EW", "East/West", ISS_OFF);
     IUFillSwitchVector(&SwapSP, SwapS, 2, getDeviceName(), "SWAP", "Swap buttons", MOTION_TAB, IP_RW, ISR_1OFMANY, 0,
                        IPS_IDLE);
-
+#ifdef no
+    //2020-06-02, wildi, I'm not fond of this CMR sync
```
knro, line 126: Can you elaborate further on this?

drivers/telescope/lx200ap_experimental.cpp
```diff
@@ -198,16 +209,32 @@ bool LX200AstroPhysicsExperimental::updateProperties()
...
+        IDSetSwitch(&ParkSP, nullptr);
+
+
+#ifdef no
+        // 2020-05-30, wildi, this is the wrong place to load config data,
+        // see celestron driver.
+        // The RA/Dec values are 0,90 at startup and should not be
+        // displayed until they are set in UnPark()
         // load in config value for park to and initialize park position
         loadConfig(true, ParkToSP.name);
```
knro, line 239: I don't think loadConfig is the correct way here. You have functions to read each value separately now from the config file without having issue an ISNewSwitch(...) function call which is not required at this stage.

IUReadConfigSwitch should be used.

drivers/telescope/lx200ap_experimental.cpp
```diff
@@ -397,35 +428,69 @@ bool LX200AstroPhysicsExperimental::initMount()
...
-        IDSetNumber(&MeridianDelayNP, nullptr);
+        APUTCOffsetNP.s  = IPS_OK;
+        IDSetNumber(&APUTCOffsetNP, nullptr);
+
+        return true;
+    }
+    if (!strcmp(name, HourangleCoordsNP.name))
+    {
+        HourangleCoordsNP.s = IPS_BUSY;
+        IDSetNumber(&HourangleCoordsNP, nullptr);
```
knro, line 466: Why do you need to explicitly set it to IPS_BUSY and send it over to client here?

drivers/telescope/lx200ap_experimental.cpp
```diff
@@ -501,7 +567,7 @@ bool LX200AstroPhysicsExperimental::ISNewSwitch(const char *dev, const char *nam
         IDSetSwitch(&APGuideSpeedSP, nullptr);
         return true;
     }
-
+#ifdef no
```
knro, line 570: should we use a better identifier than **no** ?

drivers/telescope/lx200ap_experimental.cpp
```diff
@@ -544,15 +610,52 @@ bool LX200AstroPhysicsExperimental::ISNewSwitch(const char *dev, const char *nam
     if (!strcmp(name, UnparkFromSP.name))
     {
         IUUpdateSwitch(&UnparkFromSP, states, names, n);
-        int unparkPos = IUFindOnSwitchIndex(&UnparkFromSP);
-
-        LOGF_DEBUG("Unpark from pos set to (%d).", unparkPos);
+        ParkPosition unparkPos = (ParkPosition)IUFindOnSwitchIndex(&UnparkFromSP);
```
knro, line 613: use static_cast<...>

drivers/telescope/lx200ap_experimental.cpp
```diff
@@ -544,15 +610,52 @@ bool LX200AstroPhysicsExperimental::ISNewSwitch(const char *dev, const char *nam
...
+        else
+        {
+            UnPark();
+            IUResetSwitch(&ParkSP);
+            ParkS[1].s = ISS_ON;
+        }
+
+        ParkSP.s = IPS_OK;
+        IDSetSwitch(&ParkSP, nullptr);
+        return true;
```
knro, line 657: Why override INDI::Telescope handling? because of the IPS_ALERT?

drivers/telescope/lx200ap_experimental.cpp
```diff
@@ -591,12 +710,85 @@ bool LX200AstroPhysicsExperimental::ISNewSwitch(const char *dev, const char *nam
...
+    getMountStatus(&isAPParked);
+    if (!isAPParked)
+    {
+        // in case of simulation, the coordinates are set on parking
+        HourangleCoordsNP.s = IPS_BUSY;
+        IDSetNumber(&HourangleCoordsNP, nullptr);
+        HourangleCoordsN[0].value = get_local_hour_angle(lst, currentRA);
+        HourangleCoordsN[1].value = currentDEC;
+        HourangleCoordsNP.s = IPS_OK;
+        IDSetNumber(&HourangleCoordsNP, nullptr );
```
knro, line 727: why are we setting this twice? Also, maybe we should add a check that in case there is no change from last update, it shouldn't spam the client with unnecessary messages?

drivers/telescope/lx200ap_experimental.cpp
```diff
@@ -591,12 +710,85 @@ bool LX200AstroPhysicsExperimental::ISNewSwitch(const char *dev, const char *nam
...
         return true;
     }
 
+    if (getAPUTCOffset(PortFD, &val) < 0)
+    {
+        LOG_ERROR("Reading UTC offset  %d");
+    }
+    if (getLocalTime24(PortFD, &val) < 0)
+    {
+        LOGF_DEBUG("Reading local time failed :GL %d", -1);
```
knro, line 755: Shouldn't this be LOG_ERROR or LOG_WARN at least?

drivers/telescope/lx200ap_experimental.cpp
```diff
@@ -1693,9 +2213,10 @@ bool LX200AstroPhysicsExperimental::MoveNS(INDI_DIR_NS dir, TelescopeMotionComma
     }
 
     bool rc = LX200Generic::MoveNS(dir, command);
-
+#ifdef no
```
knro, line 2216: Why is this block not in use?

libs/indibase/inditelescope.cpp
```diff
@@ -1384,7 +1384,10 @@ bool Telescope::ISNewSwitch(const char *dev, const char *name, ISState *states,
             IUResetSwitch(&ParkOptionSP);
 
             bool rc = false;
-
+	    // 2020-06-01, wildi, why not writing while parked, slewing, ...?
+	    // In case the driver died, one can not set it to true again,
+	    // without editing the ParkData.xml
+#ifdef no
```
knro, line 1390: Parking option can only be set when unparked and idle. This change might be breaking so we'd avoid it for this release.

drivers/telescope/lx200ap_experimental.cpp
wildi: AP recommends to use :CM during the init phase and says explicitly not to use :CMR. :CMR is used to "re-calibrate" on a last slew while the mount is tracking. It does this even if the mount is on the "wrong side" of the pier. That means, the declination's axis points below horizon, as seen from the intersection of the hour with the declination axis in the direction of the telescope. :CMR needs a slew command before it can be used or as AP expresses it the object's co-ordinates needs to be set in memory. AP does not describe, what happens if used incorrectly. If I remember it correctly: a :CMR on the wrong side followed by goto command including a meridian flip will move the mount in the wrong direction. This behavior has to be tested with a physical mount. I recommend to leave it disabled until the test have been performed.

drivers/telescope/lx200ap_experimental.cpp
wildi: I came to the same conclusion and I disabled this section from line 233 to 283 using #ifdef...#endif.

drivers/telescope/lx200ap_experimental.cpp
wildi: I wanted the user to realize that HA is updated.

drivers/telescope/lx200ap_experimental.cpp
wildi: I used the #ifdef as a temporary indicator. I suggest, that all with #ifdef no labeled sections will be removed right before the driver will be renamed (loosing "experimental"). I recommend to leave it.

drivers/telescope/lx200ap_experimental.cpp
wildi: I changed all occurrences to  static_cast(<ParkPosition>...)

drivers/telescope/lx200ap_experimental.cpp
wildi: I agree (see above  "HA is update"). I reconsidered all  HourangleCoordsNP occurrences.

drivers/telescope/lx200ap_experimental.cpp
wildi: Yes and no: this is in ReadScopeStatus(). There is an error message if UTC offset can not be read, hence the user is informed. I recommend to leave it. 

I changed LOGF_* to LOG_*

drivers/telescope/lx200ap_experimental.cpp
wildi: There were about four bool variables in use. E.g., motionCommanded was never used. In general I think instead of using a boolean it is better to use the INDI states. I disabled all these variables in favor of a cleaner UnPark() method.

libs/indibase/inditelescope.cpp
wildi: I understand. The reason why I came up with this is, I had to edit ParkData.xml in order that the mount could be unparked cleanly.

drivers/telescope/lx200ap_experimental.cpp
wildi: Because I do not like the idea that the mount is unparked on startup. This mode is ok, if there is somebody beside the mount. In unattended case it could be dangerous.

E.g. if the mount is physically at park 2 position and INDI thinks it is at park 3 and Last Park is set then the set co-ordinates are wrong. 

Further I noticed that using INDI::Telescope Parking is set to IPS_OK, Park(ed) and if I click on UnPark(ed) it says:

2020-06-06T08:13:41: [INFO] Telescope already unparked. 
but the success message:

LOGF_INFO("UnPark: Current parking position ...

did not appear.

As I wrote in the source, that's a matter to be discussed.

PR #1164, opened 2020-06-06
Implement Parking for LX200 classic

SevenW:
I have implemented Parking for the LX200 classic. The classic does not natively support parking so it is emulated by eventually going to LAND mode to stop tracking.

There is however one thing that is not perfect. When Unparking, the alignment mode needs to be set to its previous value: Polar or AltAz. This state is however not saved. Please advice.

### DIFF
--- a/drivers/telescope/lx200classic.cpp
+++ b/drivers/telescope/lx200classic.cpp
@@ -131,51 +131,6 @@ bool LX200Classic::updateProperties()
             SetAxis1ParkDefault(LocationN[LOCATION_LATITUDE].value >= 0 ? 0 : 180);
             SetAxis2ParkDefault(LocationN[LOCATION_LATITUDE].value);
         }
-        
-        if (LX200Generic::isParked())
-        {
-            LOG_INFO("Try to reset object coordinates for EKOS UI");
-            double parkAz  = GetAxis1Park();
-            double parkAlt = GetAxis2Park();
-            
-            char AzStr[16], AltStr[16];
-            fs_sexa(AzStr, parkAz, 2, 3600);
-            fs_sexa(AltStr, parkAlt, 2, 3600);
-            LOGF_DEBUG("Parking to Az (%s) Alt (%s)...", AzStr, AltStr);
-
-            //LX200 classic set target in RA/Dec
-            ln_lnlat_posn observer;
-            observer.lat = LocationN[LOCATION_LATITUDE].value;
-            observer.lng = LocationN[LOCATION_LONGITUDE].value;
-            if (observer.lng > 180)
-                observer.lng -= 360;
-
-            ln_hrz_posn horizontalPos;
-            // Libnova south = 0, west = 90, north = 180, east = 270
-
-            horizontalPos.az = parkAz + 180;
-            if (horizontalPos.az > 360)
-                horizontalPos.az -= 360;
-            horizontalPos.alt = parkAlt;
-
-            ln_equ_posn equatorialPos;
-
-            ln_get_equ_from_hrz(&horizontalPos, &observer, ln_get_julian_from_sys(), &equatorialPos);
-
-            LOGF_DEBUG("Reseting to RA (%f) DEC (%f)...", equatorialPos.ra / 15.0, equatorialPos.dec);
-            if ((setObjectRA(PortFD, equatorialPos.ra / 15.0) < 0) || (setObjectDEC(PortFD, equatorialPos.dec) < 0))
-            {
-                LOG_ERROR("Error setting Reset RA/Dec.");
-                return false;
-            }
-            
-            currentRA = equatorialPos.ra / 15.0;
-            currentDEC = equatorialPos.dec;
-            NewRaDec(currentRA, currentDEC);
-
-            
-        }
-        
 
         return true;
     }
@@ -392,13 +347,9 @@ bool LX200Classic::ISNewSwitch(const char *dev, const char *name, ISState *state
         // Unpark Alignment Mode
         if (!strcmp(name, UnparkAlignmentSP.name))
         {
-            //IUResetSwitch(&UnparkAlignmentSP);
             IUUpdateSwitch(&UnparkAlignmentSP, states, names, n);
             UnparkAlignmentSP.s = IPS_OK;
             IDSetSwitch(&UnparkAlignmentSP, nullptr);
-            
-            index = IUFindOnSwitchIndex(&UnparkAlignmentSP);
-            LOGF_INFO("ISNewSwitch() UnparkAlignmentSP %d", index);
 
             return true;
         }
@@ -423,8 +374,6 @@ bool LX200Classic::saveConfigItems(FILE *fp)
 //Parking
 bool LX200Classic::Park()
 {
-    //const struct timespec timeout = {0, 100000000L};
-    
     double parkAz  = GetAxis1Park();
     double parkAlt = GetAxis2Park();
     
@@ -461,71 +410,13 @@ bool LX200Classic::Park()
     UnparkAlignmentS[curAlignment].s = ISS_ON;
     UnparkAlignmentSP.s = IPS_OK;
     IDSetSwitch(&UnparkAlignmentSP, nullptr);
-    
-    int index = IUFindOnSwitchIndex(&UnparkAlignmentSP);
-    LOGF_INFO("Park() UnparkAlignmentSP %d", index);
-
     saveConfig(true, UnparkAlignmentSP.name);
 
-
-    if (isSimulation())
+    if (!Goto(equatorialPos.ra / 15.0, equatorialPos.dec))
     {
-        Goto(equatorialPos.ra / 15.0, equatorialPos.dec);
-    }
-    else
-    {
-        //// If scope is moving, let's stop it first.
-        //if (EqNP.s == IPS_BUSY)
-        //{
-            //if (!isSimulation() && abortSlew(PortFD) < 0)
-            //{
-                //AbortSP.s = IPS_ALERT;
-                //IDSetSwitch(&AbortSP, "Abort slew failed.");
-                //return false;
-            //}
-
-            //AbortSP.s = IPS_OK;
-            //EqNP.s    = IPS_IDLE;
-            //IDSetSwitch(&AbortSP, "Slew aborted.");
-            //IDSetNumber(&EqNP, nullptr);
-
-            //if (MovementNSSP.s == IPS_BUSY || MovementWESP.s == IPS_BUSY)
-            //{
-                //MovementNSSP.s = MovementWESP.s = IPS_IDLE;
-                //EqNP.s                          = IPS_IDLE;
-                //IUResetSwitch(&MovementNSSP);
-                //IUResetSwitch(&MovementWESP);
-
-                //IDSetSwitch(&MovementNSSP, nullptr);
-                //IDSetSwitch(&MovementWESP, nullptr);
-            //}
-
-            //// sleep for 100 msec
-            //nanosleep(&timeout, nullptr);
-        //}
-
-        //if ((setObjectRA(PortFD, equatorialPos.ra / 15.0) < 0) || (setObjectDEC(PortFD, equatorialPos.dec) < 0))
-        //{
-            //LOG_ERROR("Error setting Park RA/Dec.");
-            //return false;
-        //}
-
-        //int err = 0;
-
-        ///* Slew reads the '0', that is not the end of the slew */
-        //if ((err = Slew(PortFD)))
-        //{
-            //LOGF_ERROR("Park Error Slewing to Az %s - Alt %s", AzStr, AltStr);
-            //slewError(err);
-            //return false;
-        //}
-        
-        if (!Goto(equatorialPos.ra / 15.0, equatorialPos.dec))
-        {
-            ParkSP.s = IPS_ALERT;
-            IDSetSwitch(&ParkSP, "Parking Failed.");
-            return false;
-        }
+        ParkSP.s = IPS_ALERT;
+        IDSetSwitch(&ParkSP, "Parking Failed.");
+        return false;
     }
 
     EqNP.s     = IPS_BUSY;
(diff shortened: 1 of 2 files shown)


Rushwind13:
@SevenW It looks like we are trying to solve the same problem (check the other PR against indilib:master) -- do you want to try to fix this together?

drivers/telescope/lx200classic.cpp
```diff
@@ -325,3 +346,226 @@ bool LX200Classic::saveConfigItems(FILE *fp)
...
+    {
+        setAlignmentMode(PortFD, LX200_ALIGN_LAND);
+        
+        //IUResetSwitch(&AlignmentSP);
+        //AlignmentSP.s = IPS_OK;
+        //AlignmentS[LX200_ALIGN_LAND].s = ISS_ON;
+        //IDSetSwitch(&AlignmentSP, nullptr);
+    }
+
+    LX200Generic::ReadScopeStatus();
```
Rushwind13, line 567: I think that you have to call the base class method first -- there are code branches that could return false from there.

drivers/telescope/lx200classic.cpp
```diff
@@ -325,3 +346,226 @@ bool LX200Classic::saveConfigItems(FILE *fp)
...
+
+    if (isSimulation())
+    {
+        mountSim();
+        return true;
+    }
+
+    if ((TrackState == SCOPE_PARKING) &&  (isSlewComplete()))
+    {
+        setAlignmentMode(PortFD, LX200_ALIGN_LAND);
```
Rushwind13, line 559: Can you save off the current state of AlignmentSP, here (using IUFindOnSwitchIndex()), and then use it to restore the value in UnPark()?

SevenW: The state needs to be stored persistently over power cycles of the indi-computer and the LX200 mount. So keeping the state in a member variable will not work.

Nicest for me would be if this state is also stored in the ParkData.xml. However that change is a bit involved.

I think I may be implementing an additional switch similar to AligmentSP that serves the purpose of defining (no more and no less) the UnPark recovery state. 

Please share your insights.

drivers/telescope/lx200classic.cpp
SevenW: What a nice coincidence!

You are right on this. I think the two alternate exit paths are already included in my override, but it is not safe. Calling first the base class spoils the state that I need to query. 

I have seen a full code clone of the base class in another example. I tried to avoid that. I wil take a further look on this code.

Rushwind13: hmm, I didn't think about it spoiling my state... let me look that over some more in my code

drivers/telescope/lx200classic.cpp
Rushwind13: I'm having this exact same problem. My code works as long as I don't stop and restart Ekos. I think we need to find a way to persist this in ParkData.xml (which does store a boolean called "parkstatus", so it seems like that's on the right track... ) or the mount .xml config. Maybe there's a way to store it across two config Switches, like "Parked Alignment" and "Tracking Alignment". That seems wrong somehow :D

SevenW: Yes. @knro what is your preference?

SevenW:
@Rushwind13 I have looked a bit more in the code. Solving through ParkData.xml is the nicest solution and I think it may be not too difficult. Shall I give it a try later today?

SevenW:
I was thinking of adding a getUnparkAlignment and setUnparkAlignment method to inditelescope. I need to verify if it loads current xml files without this property correctly. All drivers not using this feature are not bothered with the additional content.

Rushwind13:
I'd love to see your implementation, yes :) I've been trying really hard not to change the base classes, but as long as we can pass the required parameters through without causing a breaking change for everyone else, that sounds like a good plan.

Rushwind13:
LX200FS2 overrides INDI::Telescope::saveConfigItems() and adds additional fields in its own implementation method (indi_lx200fs2.cpp line 162); I assume that's the proper way, as LX200Telescope does the same.

knro:
So adding a property to save this across restarts is the usual way to resolve this.

knro:
@Rushwind13 That's the way to do it, correct.

drivers/telescope/lx200classic.cpp
knro: Using a property to save this across restarts is the way to go for now.

SevenW:
I have added the the UnparkAlignment property to store the alignment mode required for the unpark function. Some more code cleanups and simplifications.

drivers/telescope/lx200classic.cpp
```diff
@@ -322,6 +365,190 @@ bool LX200Classic::saveConfigItems(FILE *fp)
...
+    
+    if (LX200Generic::ReadScopeStatus())
+    { 
+        if ((TrackState == SCOPE_PARKED) && (curTrackState == SCOPE_PARKING) && !isSimulation())
+        {
+            setAlignmentMode(PortFD, LX200_ALIGN_LAND);
+            //Update the UI
+            LX200Generic::getAlignment();
+        }
+    }
```
Rushwind13, line 551: If he base class returns false, you need to return false.

drivers/telescope/lx200classic.cpp
```diff
@@ -322,6 +365,190 @@ bool LX200Classic::saveConfigItems(FILE *fp)
...
+    return true;
+}
+
+bool LX200Classic::ReadScopeStatus()
+{
+    int curTrackState = TrackState;
+    
+    if (LX200Generic::ReadScopeStatus())
+    { 
+        if ((TrackState == SCOPE_PARKED) && (curTrackState == SCOPE_PARKING) && !isSimulation())
```
Rushwind13, line 545: At least one of these needs to be a || or else this can never be true

drivers/telescope/lx200classic.cpp
```diff
@@ -322,6 +365,190 @@ bool LX200Classic::saveConfigItems(FILE *fp)
...
+        }
+    }
+
+    SetParked(false);
+    return true;
+}
+
+bool LX200Classic::SetCurrentPark()
+{
+    ln_hrz_posn horizontalPos;
```
Rushwind13, line 497: Most of this is boilerplate copied from... lx200telescope? Is here any way for us to elevate this code so both of us can use it without copy pasta these entire methods?

drivers/telescope/lx200classic.cpp
SevenW: I think this is correct. curTrackState was when entering ReadScopeStatus, and if TrackState in the mean time has changed to SCOPE_PARKED, then it evaluates to true.

Rushwind13: Sorry, I missed the “cur”. I retract the comment 😄

drivers/telescope/lx200classic.cpp
SevenW: I took it from here: https://github.com/indilib/indi/blob/9f37544774c0cd22fe4289d60a3ff32a9711ca81/drivers/telescope/lx200ss2000pc.cpp#L329

I would not call it boiler plate. It converts AltAz parking coordinates to RaDec. And in one case the other way. The LX200 classic does not support AltAz position setting. Most later LX200 scopes do. Hence the need to control it in RaDec.  

Would there be a more compact helper function anywhere else?

drivers/telescope/lx200classic.cpp
SevenW: Fixed this. Will upload after some testing.

drivers/telescope/lx200classic.cpp
Rushwind13: If there’s not already a “convertAltAz( &RA, &Dec )” and “convertRADec( &Alt, &Az)”, there should be 😄 but that’s like indibase type stuff

drivers/telescope/lx200classic.cpp
```diff
@@ -544,11 +500,68 @@ bool LX200Classic::ReadScopeStatus()
...
+            }
             //Update the UI
             LX200Generic::getAlignment();
         }
     }
 
     return true;
 }
+
+void LX200Classic::azAltToRaDecNow(double az, double alt, double &ra, double &dec)
```
Rushwind13, line 518: This! Yes! 🖖😌

drivers/telescope/lx200classic.cpp
```diff
@@ -412,7 +396,7 @@ bool LX200Classic::Park()
     IDSetSwitch(&UnparkAlignmentSP, nullptr);
     saveConfig(true, UnparkAlignmentSP.name);
 
-    if (!Goto(equatorialPos.ra / 15.0, equatorialPos.dec))
+    if (!Goto(parkRA, parkDEC))
```
Rushwind13, line 399: This is all so much cleaner with the utility function handling the mathy  bits

drivers/telescope/lx200classic.cpp
SevenW: I have moved the coordinates system conversion code to two separate methods. Those two methods could be moved down the specialization chain, or two a separate helper module.

drivers/telescope/lx200classic.cpp
SevenW: I agree. Much nicer.

drivers/telescope/lx200classic.cpp
SevenW: Implemented it like this.

drivers/telescope/lx200classic.cpp
SevenW: If you like to reuse, can you figure out where to move to?

drivers/telescope/lx200classic.cpp
```diff
@@ -322,6 +365,203 @@ bool LX200Classic::saveConfigItems(FILE *fp)
...
 
     return true;
 }
+
+bool LX200Classic::UnPark()
+{
+    if (isSimulation() == false)
+    {
+        // Parked in Land alignment. Restore previous mode.
+        if (setAlignmentMode(PortFD, IUFindOnSwitchIndex(&UnparkAlignmentSP)) < 0)
```
Rushwind13, line 418: So my mount doesn't implement AlignmentMode *facepalm*, I will work on turning "RAMotorOff()" and "RAMotorOn()" into methods, and maybe we can integrate in the parent class.

drivers/telescope/lx200classic.cpp
Rushwind13: Yep, let me find a good spot for it... (will probably find it already exists :D )

Rushwind13: TransformCelestialToTelescope() does this, looks like it is provided by alignment/AlignmentSubsystemForDrivers.h (DSC, Temma, and Skywatcher are the only folks that are using it, I will give it a try)

drivers/telescope/lx200classic.cpp
SevenW: As my mount can be both poalr and altAz, a generic function would control both motors. So I propose to use a more generic name than RA....

SevenW:
Tried it, and works as long as the driver runs. But does not help to be rally persistent. The implementation with the additional property works very well.

Verstuurd vanaf mijn iPhone

drivers/telescope/lx200classic.cpp
Rushwind13: It looks like LX200Telescope just calls "slewToPark()" (a direct LX200 command), and several of the child classes don't support that, and had to implement Park() and UnPark() from scratch. 
Given the way the classes are laid out, LX200Generic is the only place in between the two, but that seems to want to be an empty passthrough class.

This behavior is implemented in indicom.h and there's a really gorgeous implementation in alignment/*Math*.h, and all the ln_ decorators are because these are functions directly from libnova.

Maybe there should be LX200Telescope::SlewToPark() which either calls slewToPark() (the internal command), or this snippet.

Or maybe this is LX200Generic::Park() -- the class should stay empty, but this is a bit that's not in the base that is in several of the children (Classic, FS-2, SS2000PC, AP, AP_GTOCP2, AP_Experimental). 

I don't know the right answer -- just seems like if the answer going forward for parking will be "copy paste these exact 5 functions", we've built the object model wrong somehow...

drivers/telescope/lx200classic.cpp
Rushwind13: TrackingStop()/TrackingStart()?

SevenW: Sounds good

drivers/telescope/lx200classic.cpp
SevenW: I went for two days back and forth through the code stack to figure it out. With the classic there were three aspects different form assumption made elsewhere:

- The slew to park command as you mention
- Lack of setting positions in AzAlt. Only RaDec possible
- The mount can be AltAz or Polar, and therefore needs to remember the state.

The LX200telescope class may be able to implement (part of) this. However it seems the situation is that there are a lot of different choices made by mount manufacturers.

Rushwind13: Agreed, I wonder about “intent” in the LX200 family. I feel like LX200Telescope is a bit prescriptive, but that’s  what you get when everyone copies an existing standard... sort of. Sigh. Maybe this just gets copy pasta

knro:
Do you want to move the coordinate conversion functions up to LX200Telescope ? maybe they can be re-used by other drivers?

SevenW:
There is obviously a need for this to be a helper function. I think it should either be even in a lower base class or really a helper function. Searching on ... reveals severla hits and checking a few of them reveal that the altaz->eq coverison is guite popular. Also in GotoAltAz functions. Not only in Park of LX200 type scopes. See
https://github.com/indilib/indi/blob/d28dbe5e6a9c373c4e8efb65fcbca14956df659c/drivers/telescope/celestrongps.cpp#L619
https://github.com/indilib/indi/blob/baed690f6cfd9efb417d5b060f771fd6ded28f16/drivers/telescope/ieqpro.cpp#L703

I am happy to add the new methods to another layer, but I think you need to guide here..

Otherwise please pull in this code for the time being, and lets put in an enhancement request to further discuss the location of the helper functions. I am happy to work with you to move some code around.

My current implementation does the job and in general the functions are a bit cleaner than most of the examples I have walked through.

drivers/telescope/lx200classic.cpp
SevenW: I have proposed to move the AltAz<->Equatorial conversion to another location, but with another pull request. Lets take time to figure out the best location.

knro:
I agree, there is no need for rush here. INDI::Telescope is a good place for this, but can also be moved to indicom, but you wouldn't be able to use LocationNP property. Extra checks need to be added to account for various mount capabilities.

PR #1165, opened 2020-06-06
LX200 FS-2 Park

Rushwind13:
DO NOT MERGE 

I have attempted to make FS-2 park properly by turning off the motors when parked (set rate to CENTERING, move East). 

I added a Switch to toggle the feature, and code to stop the mount when parking is complete. 
This works great, and the mount parks and unparks, and slews just fine...

...until I completely stop Ekos and restart it. 
At this point, Ekos turns on tracking, but doesn't unpark the mount -- this leaves the mount in a weird state where it can't be slewed.

So I still have errors in state machine.

Can someone help me to understand how to maintain this behavior across reboots?


knro:
What do you mean Ekos turn on tracking exactly? Where is that command issued?

Rushwind13:
With my enhancement in place, I can hear the RA drive stop running (and the current draw goes from 0.85A when tracking to 0.25A) when parked. 

If I then go to Ekos window and click Disconnect and Stop, the mount stays silent.
If I then start Ekos,  I can hear the RA drive start up, and the current goes back up to 0.85A.

I am fairly certain this is when LX200Telescope::initProperties() does this on line 228 of lx200telescope.cpp 

TrackState = SCOPE_IDLE;

So I am calling LX200Generic::initProperties() first, then trying to get everything to the correct state in my LX200FS2::initProperties() method. (line 51 -- SetParked(false);)

drivers/telescope/lx200fs2.cpp
```diff
@@ -174,6 +214,28 @@ bool LX200FS2::Park()
...
 
+bool LX200FS2::ReadScopeStatus()
+{
+    bool retval = LX200Generic::ReadScopeStatus();
+    // For FS-2 v1.21 owners, stop tracking once Parked.
+    if (retval && 
+        StopAfterParkS[0].s == ISS_ON &&
+        isConnected() && 
+        !isSimulation() &&
+        TrackState == SCOPE_PARKING)
```
SevenW, line 225: Could it be that this is no longer SCOPE_PARKING after calling the base class?

Rushwind13: Simply adding SCOPE_PARKED to this logic messed things up, but maybe I could figure out how to intercept this in a smarter way

knro:
TrackState = SCOPE_IDLE is how the state is initilized by default. However, in updateProperties() you can call InitPark() and then this would reset the TrackState depending on whether the mount was parked in ParkData.xml or not. So if it was parked, TrackState = SCOPE_PARKED right there and I don't see how it would case the tracking to engage?

Rushwind13:
I will check InitPark() and make sure the condition is properly handled, thanks for the workflow hunt 😊

knro:
@Rushwind13 What's the update on InitPark usage?

Rushwind13:
I cleaned things up a bit, and added this to InitPark(), but there are still some broken use cases.

1) Mount goes to idle for a moment after Parking slew stops and before changing from Parking to Parked. This causes the mount to runaway to the east during TrackingStop().
2) Once the state machine gets messed up, the mount stops being able to slew (Arrow buttons work, but Goto commands do not cause the mount to start moving).

It feels like I'm missing something on the state change, and I'd like to be a little more defensive in TrackingStop() around "am I already stopped?"

Rushwind13:
I'm not 100% sure that I'm setting the mount's speed properly. Like, I'm making then updateSlewRate() call, but do I need to wait for that to complete before making a movement request? I tried adding an Abort() :Q# into the stream, but I still get runaway slews when parking.

Rushwind13:
Can I assume that adding sleep() and other timing hacks is a bad idea?

knro:
yes sleep() should be avoided. is there a way to query slew rate after setting it to ensure it has indeed changed?

Rushwind13:
I added a set of states for the mount to walk through in order to park and unpark, so that the commands come no more often than once/update. This seems to have calmed things down quite a bit, but I am still having trouble restoring the original mount status after unpark. Will keep poking.

Rushwind13:
Now it does what I think are all the steps needed to stop and then restart and restore the motors. With a clean reboot, I can get park and unpark to happen once. Then the mount idles, but refuses to slew. I can only move using the Ekos hand controller buttons.

I'm sort of lost, at this point. What do I need to do to the mount to get it to unpark without losing its mind?

PR #1172, opened 2020-06-13
UDP support for skywatcherAPI

### LINKED ISSUE #1158: indi_skywatcherAltAzMount does not connect wirelessly to Az-GTI
**Describe the bug**
Using exactly the same parameters than I use for connection in indi_azgti_telescope driver: 
192.168.43.195
11880
UDP
The driver attempts to connect but nothing happens.
The mount and the Raspberry Pi 4 (w/Astroberry server) connects to my phone's hotspot (192.168.43.1).
If I do the same with indi_azgti_telescope it connects immediately.
Last version of Astroberry (Version 2.0.1)

gubihu:
SkywatcherAPI::TalkWithAxis() changed to use tty_read_section() to read the response for a command. This way both Serial and UDP connection can work. Fix #1158

### DIFF
--- a/drivers/telescope/skywatcherAPI.cpp
+++ b/drivers/telescope/skywatcherAPI.cpp
@@ -752,6 +752,7 @@ bool SkywatcherAPI::TalkWithAxis(AXISID Axis, char Command, std::string &cmdData
     bool StartReading   = false;
     bool EndReading     = false;
     bool mount_response = false;
+    char response[257];
 
     SendBuffer.push_back(':');
     SendBuffer.push_back(Command);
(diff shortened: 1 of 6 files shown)


knro:
Great work!

@kecsap Any comments?

kecsap:
Ok.

knro:
@gubihu yes there are lots of duplicated code there because IIRC, one was copied over to make the Simple version. At any rate, feel free to make whatever necessary changes since this driver is not utilized much (compared to EQMod for example) and can use improvements.

gubihu:
@knro Yes, it is quite obvious from the code, that one was copied over to make an other version. What I'm trying to tell, that these code parts are quite generic, any other driver using skywatcherAPI would have to do quite the same. 
Are there any other driver out there that use the skywathcerAPI?  Those will break now because of not giving implementation to the new function.

knro:
No it's just these two drivers. I don't think even the SkyWatcherAPIMount driver is used by end-users.

PR #1177, opened 2020-06-17
CCD Simulator focusing

sterne-jaeger:
Currently, the CCD simulator supports focusing only in combination with the Focus Simulator. This change creates the option to attach a real (absolute) focuser and using the CCD simulator so that focusing can be tested during daytime.

**Test Plan**
Create a EKOS profile combining an absolute focuser with the CCD Simulator. Use the Simulator settings of the CCD Simulator to configure the focus position and the maximal focuser position. Using this setup, try different focuser positions and check, whether the stars change their FHWM values.

### DIFF
--- a/drivers/ccd/ccd_simulator.cpp
+++ b/drivers/ccd/ccd_simulator.cpp
@@ -178,7 +178,7 @@ bool CCDSim::initProperties()
                        SIMULATOR_TAB, IP_RW, ISR_1OFMANY, 60, IPS_IDLE);
 
     // Simulate focusing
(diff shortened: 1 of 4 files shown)


knro:
So in the current method optimal focus around ~36,000, shouldn't this be the same?

sterne-jaeger:
That’s easy to be solved, sure. What about the old solution? Should I remove it? It is currently commented out in the CCD Simulator, but still active in the Focus Simulator.

knro:
If there are no drivers using that, then sure remove it.

knro:
@sterne-jaeger Any update on this PR?

PR #1206, opened 2020-08-07
Update gps_simulator.cpp

rtlprmft:
Adding the option to save the simulated coordinates


knro:
This should be in 1 PR and not 2. I'll close both now, and please submit as one.

rtlprmft:
A reasonable explanation how to do that would help... this github stuff is a mess if you are not the experiences type...

PR #1208, opened 2020-08-09
Feature- Satellite tracking with TLE for 10Micron mounts

IslamAlam:
This feature provide satellite tracking for the mount LX200 by creating a new tab for providing the [Two-line element set
](https://en.wikipedia.org/wiki/Two-line_element_set) 

It was developed for the satellite tracking at the observatory of Technical University of Munich by @ee5cd7d5d


knro:
This is awesome!! Can this be in any way generalized to more mounts? It looks to be using specific 10Micron commands.

drivers/telescope/lx200_10micron.cpp
```diff
@@ -233,6 +268,7 @@ bool LX200_10MICRON::ReadScopeStatus()
     }
     DEBUGFDEVICE(getDefaultName(), DBG_SCOPE, "CMD <%s> RES <%s>", cmd, data);
 
+    // TODO: check if this needs changing when satellite tracking
```
d33psky, line 271: the mount command protocol docs of 2.14.11 list only 8 fields for :Ginfo#
what doc version did you write against ?

drivers/telescope/lx200_10micron.cpp
```diff
@@ -514,6 +550,167 @@ bool LX200_10MICRON::setLocalDate(uint8_t days, uint8_t months, uint16_t years)
...
+    }
+    return 0;
+}
+	
+bool LX200_10MICRON::SetTLEfromDatabase(int tleN)
+{
+    char command[12];
+    snprintf(command, sizeof(command), ":TLEDL%01d#", tleN);
+
+    LOG_INFO("Setting TLE from Database");
```
d33psky, line 614: just an idea : can we add tleN here as %d ?

drivers/telescope/lx200_10micron.cpp
```diff
@@ -514,6 +550,167 @@ bool LX200_10MICRON::setLocalDate(uint8_t days, uint8_t months, uint16_t years)
...
+            pch = strtok (NULL, "\n");
+        }
+    }
+    return 0;
+}
+	
+bool LX200_10MICRON::SetTLEfromDatabase(int tleN)
+{
+    char command[12];
+    snprintf(command, sizeof(command), ":TLEDL%01d#", tleN);
```
d33psky, line 612: question: is %d insufficient here ? does it need to be %01d ? I don't see any hints to that in the docs.

jochym:
The type of api this mount (LX200) uses was generally also my idea regarding implementing sat tracking in CAUX and maybe EQmod drivers. Goto is useless for this, but if you have control over tracking rates (like CAUX and EQmod) you can track anything. My idea was to submit the tracking table to the driver - to avoid putting the whole celestial mechanics solver into the driver. But maybe the solver can be implemented as a plugin, using some external library. Similarly to the way the alignment plugin is implemented. The first step may be just formalizing custom tracking rate API. Having this one can implement tracking with external code driving the indi driver.

drivers/telescope/lx200_10micron.cpp
ee5cd7d5d: Hi, I am using a command protocol we got sent from 10micron upon request from the university. It should apply to mounts with 2.15.8 and above. I think with this comment I meant the output of :Gstat#, because it has states specific for satellite tracking

drivers/telescope/lx200_10micron.cpp
ee5cd7d5d: I believe so

drivers/telescope/lx200_10micron.cpp
ee5cd7d5d: Hi, I think %d should be enough

ee5cd7d5d:
> This is awesome !
> I just left a few questions and minor improvement idea.
> Maybe rename the PR title to "Feature- Satellite tracking with TLE for 10Micron mounts" because it uses the 10Micron-specific satellite tracking features that may not exist in other LX200 protocol mounts.
> Do you have any image results of this code online ?
> Thanks for the PR !

I have a forum thread from ages ago: https://indilib.org/forum/mounts/6177-update-on-the-lx200-10-micron-driver.html#58016 I just posted some pictures there. More to come soon!

ee5cd7d5d:
> The type of api this mount (LX200) uses was generally also my idea regarding implementing sat tracking in CAUX and maybe EQmod drivers. Goto is useless for this, but if you have control over tracking rates (like CAUX and EQmod) you can track anything. My idea was to submit the tracking table to the driver - to avoid putting the whole celestial mechanics solver into the driver. But maybe the solver can be implemented as a plugin, using some external library. Similarly to the way the alignment plugin is implemented. The first step may be just formalizing custom tracking rate API. Having this one can implement tracking with external code driving the indi driver.

I did put some thought into how to implement this feature correctly, and I could not come up with anything clever given the current state of the command protocol. I believe now there is a command protocol 2.16.x right @IslamAlam ? So maybe there is something new in the last version. One can implement custom tracking rates for the LX200, even individually in RA and E, so in theory one could update the custom tracking rates as the satellite pass occurs. The problem with this is that the mount does not provide any way to do this continuously and I do not want to start writing code for calculating optimal tracking rate update frequency and such when there is a built-in satellite tracking functionality. It is technically possible for this mount, but it seems like a "hack".

ee5cd7d5d:
> This is awesome !
> I just left a few questions and minor improvement idea.
> Maybe rename the PR title to "Feature- Satellite tracking with TLE for 10Micron mounts" because it uses the 10Micron-specific satellite tracking features that may not exist in other LX200 protocol mounts.
> Do you have any image results of this code online ?
> Thanks for the PR !

Thanks for your awesome replies guys! @IslamAlam can you please rename the pull request as suggested, and replace the %01d with %d? Thanks a lot!

IslamAlam:
> > This is awesome !
> > I just left a few questions and minor improvement idea.
> > Maybe rename the PR title to "Feature- Satellite tracking with TLE for 10Micron mounts" because it uses the 10Micron-specific satellite tracking features that may not exist in other LX200 protocol mounts.
> > Do you have any image results of this code online ?
> > Thanks for the PR !
> 
> Thanks for your awesome replies guys! @IslamAlam can you please rename the pull request as suggested, and replace the %01d with %d? Thanks a lot!

it is done!

knro:
@IslamAlam Thanks!

@d33psky Is this good to go?

d33psky:
Yes, looks fine to me.
-- Hans

siyu6974:
Wow cool, I'm working on a very similar project. My objective is to track the ISS with my EQ5 mount (pretty generic mount with no encoder or fancy stuff). The first stage is to use TLE to "blind track" the object and then the second stage is to use guide camera to compensate for the crappiness of the mount gears.

PR #1211, opened 2020-08-17
Added Spinel camera pixel size info to v4l2driver.cpp

kaourimo:


knro:
Last should be **true** since this was reported by a user.

PR #1230, opened 2020-09-02
change RPM Release to contain a timestamp

xsnrg:
This PR will change the name of the RPM package to contain a timestamp of when the package is created. This helps in 2 ways:

1. RPMs can now be automatically updated. They key off the filename, and if the filename had not changed, more steps were needed to udpate.
2. It is much easier to trace the RPM package back to a time in git history


xsnrg:
needs to be cleaned up. lots of junk commits

PR #1231, opened 2020-09-02
modify libindi.spec to add a timestamp to the RPM package generated

xsnrg:
This PR adds a timestamp to the RPM filename for 2 reasons:

1. rpm updates can happen automatically as they key off the version in the file name to notice updates
2. the timestamp can be tied back to the git log


xsnrg:
https://copr.fedorainfracloud.org/coprs/xsnrg/libindi-bleeding/builds/

PR #1239, opened 2020-09-24
Fixed Losmandy Gemini failed to complete park issue

grbsystems:
The Gemini controlled mount would respond to a park command and would physically park itself.

The indi driver would never report the park as complete, and so shutdown actions would not be performed by Ekos.

The issue was caused by the call to updateMovementState when in the SCOPE_PARKING state.  Inspection shows that this will reset the state to SCOPE_SLEWING, and hence the park state will never be correctly set.

The fixed removes the call to updateMovementState when the scope is parking.  I also removed a redundant call to explicitly set the track state and left some of the debug logging I added in place.

The fix has been tested using the mount and CCD simulator by having a sequence that slews to a star, takes one image and then completes. The mount using a Gemini II mini successfully slews and parks as expected.

If required I can further clean up the logging code added.  I would recommend against this as I added it to debug an issue, and as such is likely to prove useful in the future.


knro:
Thanks, is SCOPE_TRACKING set elsewhere?

grbsystems:
updateMovementState is called in a few places, I did not track them all.  It normalizes the gemini reported state to the enum.  With the flow specific to the issue, the final call to set the track state explicitly was redundant, because it had already been set to tracking.

In fact, explicitly setting it may have caused an issue because the gemini has a mode where it doesn't track, but it will slew.   This is correctly handled by updateMovementState.  The explicit setting of track state could erroneously set the state to SCOPE_TRACKING when it should report no tracking, SCOPE_IDLE. 

Happy to do some further due diligence on this if required.

grbsystems:
Also, the tracking state is set explicitly in a few places, but on the face of it they all make sense in context.

knro:
Thanks, let's try to get feedback from other users now that it's merged. Thank you for your great efforts.

grbsystems:
No problem.  Nice to give something back.

drivers/telescope/lx200gemini.cpp
```diff
@@ -442,17 +444,16 @@ bool LX200Gemini::ReadScopeStatus()
...
     else if (TrackState == SCOPE_PARKING)
     {
         updateParkingState();
-        updateMovementState();
 
         if (isSlewComplete())
         {
+            LOG_DEBUG("Park is complete ...");
             SetParked(true);
             sleepMount();
```
palmito9, line 458: sleepMount() needs to be commented/removed for the mount not to go into SCOPE_SLEWING after SCOPE_PARKING

palmito9 (commented):
Everything else works perfectly

drivers/telescope/lx200gemini.cpp
palmito9: Sorry, let me be more explicit:
The initial commit fixed the issue were the mount went back to tracking after parking completed, thus preventing further actions (like a flat sequence).
However the status in ekos and mount control showed "Slewing" instead of "Parked".

To reproduce, I simply unparked the mount set tracking on, manually slew a bit and then hit park.
Status went: Tracking->Parking->Slewing

After further investigation I found out sleepMount() was causing this (The mount itself was reporting SLEWING after Gv# command). Once sleepMount() was commented the status correctly displayed "Parked" after parking.

I tested further starting from parked state with a sequence of 10 lights followed by 10 flats (with park mount checked). It worked perfectly, mount unparked, slewed, tracked, then parking and parked.

PR #1243, opened 2020-10-01
Improve on deadlocks and integrity

TallFurryMan:
These fixes were tested to help resolve deadlock and sigsegv issues with KStars.

They are not expected to change any functionality from the specification.

The first commit - the delay after setting the disconnection state - is the most fragile. It embeds an arbitrary duration and requires clients to check the `isConnected()` flag before trying to interact with properties. However, it resolves most if not all deadlock situations with little side-effect (waiting a bit more for disconnection is non-critical).

Note that KStars also requires a few fixes to successfully manage its own shortcomings (part of another pull request).

Although the `INDI::Property` API is very light, those changes include a few unitary tests. Integrity tests are skipped because it is not obvious yet that we do want integrity checks embedded in INDI.

### DIFF
--- a/libs/indibase/baseclient.cpp
+++ b/libs/indibase/baseclient.cpp
@@ -56,6 +56,7 @@
 #endif
 
 #define MAXINDIBUF 49152
+#define DISCONNECTION_DELAY_US 500000
 
 INDI::BaseClient::BaseClient() : cServer("localhost"), cPort(7624)
 {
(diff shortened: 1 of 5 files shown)


knro:
Thanks Eric. So the delay after disconnect as your pointed out is rather arbitrary. If you remove this delay, then you can reproduce a deadlock? I cannot reproduce any deadlocks here.

TallFurryMan:
KStars deadlocks on the listener thread `join()` when it attempts to use a property while the server is disconnecting. This is a Qt signal deadlock.

The solution is to check server's connectivity with `isConnected()` before attempting to use the property in the signal, but that fails because the disconnection thread does not yield between setting the disconnection flag and joining.

I am finishing the related MR for KStars, I added a test to show the problem. Very basically to test yourself, in `test_ekos_simulator`, move the Ekos startup/terminate at the beginning/end of the `testColorSchemes` test case so that Ekos is cycled for each fixture, and remove the hack waiting for 10 seconds for devices to connect from `KTRY_EKOS_START_SIMULATORS` so that properties are still flowing in when the test finishes. And poof, deadlocks nearly immediately.

Of course, it would be easier if it only deadlocked. But it also sigsegv for other reasons also taken care of here and in KStars, which makes investigation slightly more painful :)

PR #1256, opened 2020-11-08
Fix/streammanager - fix recording long exposure time

pawel-soja:
Hi!

Waiting for the frame to close its stream is annoying due to the long exposure times.
I moved the counting of time and frame rate from the processing thread to be able to stop the stream.

:)


pawel-soja:
In the meantime, This is a fix for a serious issue.

If you want to take 1 photo with an exposure time of 30 minutes, you have to wait 60 minutes to take another.

PR #1258, opened 2020-11-08
fix/performance/streammanager - Do not block the stream thread when sending FPS status, fix frames discarding.

pawel-soja:
I found another performance issue that has been around for a long time. I didn't notice him last time.

This simple fix sends the FPS value without blocking threads.
This is important when running the indi server remotely.


pawel-soja:
The bug fix causes that the rejected frames are not counted to the FPS statistics, which are then used to count the duration of the recorded video.
Additionally, in the Indi settings you can see the divisor effect (FPS).

PR #1265, opened 2020-11-21
Fix for when observer location is less than 1 degree East

kenkelso:
If the observer longitude was less than 1 degree East the code wrongly sent West to the mount because it checked to see if the degrees where > 0 not accounted for the case such as my location which is 000:07:24 East. Changing to >= fixes this issue.


knro:
shouldn't it then use if longitude > 0 ? that's a double value so it would work as well.

kenkelso:
Yes that would work to Jasem. 

Ken

knro:
Ok please change to that since it's safer.

kenkelso:
Wilco

PR #1267, opened 2020-11-23
Format cmake code

nicolasfella:
The inconsistent coding style and the lack of indentation for some if
statements makes it hard to read the code.

Done using https://github.com/cheshirekow/cmake_format with some manual
intervention.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,61 +1,69 @@
 cmake_minimum_required(VERSION 3.0)
-PROJECT(libindi C CXX)
+project(libindi C CXX)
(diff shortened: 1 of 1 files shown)


knro:
is it possible to style it differently? we use astyle for C/C++ code, is there something similar for CMake?

nicolasfella:
cmake-format has a range of options (https://cmake-format.readthedocs.io/en/latest/format-usage.html), but I'm not familiar with the details

knro:
I will check for a plugin to keep the file reformatted at all times.

PR #1283, opened 2020-12-13
Add get_hrz_from_equ_sidereal_time and get_equ_from_hrz_sidereal_time to indicom.c

iliaplatone:
This PR adds get_ra_dec_coordinates function to indicom.c. Internal functions needed if not compiling with libnova


libs/indicom.c
```diff
@@ -1647,36 +1647,52 @@ void get_equ_from_hrz(struct ln_hrz_posn *object, struct ln_lnlat_posn *observer
 
     ln_get_equ_from_hrz(&libnova_object, observer, JD, position);
 }
-
 #endif // HAVE_LIBNOVA
 
-double get_local_hour_angle(double sideral_time, double ra)
```
knro, line 1653: is there a reason this is removed? it's not used by any drivers?

iliaplatone:
This was an error, will push shortly to fix it, sorry

libs/indicom.c
iliaplatone: It was not removed however, only moved.. Restored now in its previous position

libs/indicom.c
```diff
@@ -1656,27 +1655,44 @@ double get_local_hour_angle(double sideral_time, double ra)
...
-        az = 360 - az;
-    *Alt = alt;
-    *Az = az;
+    if(ha < -6.0 || ha > 6.0)
+        az = 180.0-az;
+    *Alt = rangeDec(alt);
+    *Az = range360(az);
+}
+
+void get_ra_dec_coordinates(double Alt, double Az, double Lat, double lst, double* Ra, double *Dec)
```
knro, line 1675: why not use libnova functions?

iliaplatone: Ah.. yes.. err.. why?

iliaplatone: Sorry, these functions don't need external dependency, the main reason is that need arbitrary sidereal time argument. Since that's obtained by a libnova function maybe calling there libnova funcions again maybe will better

knro:
Wouldn't a better name be **indi_get_hrz_from_equ** ..etc? this way these function would be simple wrapper for the libnova ones to correct for the Azimuth discrepency.

iliaplatone:
this would need to rename all occurrencies. I just added the functions that need sidereal time.

PR #1284, opened 2020-12-14
MacOS Fix - replace non-standard uint type with unsigned

pawel-soja:
Fix for "error: unknown type name 'uint'; did you mean 'int'" on MacOS


knro:
Why not use uint32_t ?

pawel-soja:
I'm not using fixed width integer types where it's not important.
If you have a preference to use uint32_t, I can change that.

knro:
Sure go ahead, I'm trying to use fixed width every where in the code base now to avoid any ambiguities.

PR #1286, opened 2020-12-14
Indiapi/refactoring - IWidgetVectorProperty proposal to simplify the code

pawel-soja:
I would suggest starting to simplify the code that is duplicated.
If you accept my changes, I will continue to work.
I am happy to answer any doubts.

I tried to transfer the behavior from the original implementation.

I have noticed inconsistencies.
only in some places is it used:
<code>
bool pRegistered = oneProp->getRegistered();
</code>
or:
<code>
if (pType == INDI_BLOB)
    oneProp->setRegistered(false);
</code>
I don't know all the library code yet.

### DIFF
--- a/indiapi.h
+++ b/indiapi.h
@@ -253,8 +253,8 @@ typedef struct _IWidgetVectorProperty
     double timeout;
     /** current property state */
     IPState state;
-    /** Widget comprising this vector */
-    IWidget *widets;
+    /** Widget comprising this vector - first cast to valid type to iterate */
+    void *widets;
     /** Dimension of widgets[] */
     int  nwidgets;
     /** ISO 8601 timestamp of this event */
--- a/indiapi.h
+++ b/indiapi.h
@@ -218,6 +218,52 @@ IP_RW  /*!< Read & Write */
  */
 
 /**
+ * @struct IWidget
+ * @brief Common data for
+ * IText, INumber, ISwitch, ILight, IBLOB
+ */
+typedef struct _IWidget
+{
+    /** Index name */
+    char name[MAXINDINAME];
+    /** Short description */
+    char label[MAXINDILABEL];
+} IWidget;
+
+/**
+ * @struct IWidgetVectorProperty;
+ * @brief Common data for
+ * ITextVectorProperty, INumberVectorProperty,
+ * ISwitchVectorProperty, ILightVectorProperty,
+ * IBLOBVectorProperty
+ */
+typedef struct _IWidgetVectorProperty
+{
+    /** Device name */
+    char device[MAXINDIDEVICE];
+    /** Property name */
+    char name[MAXINDINAME];
+    /** Short description */
+    char label[MAXINDILABEL];
+    /** GUI grouping hint */
+    char group[MAXINDIGROUP];
+    /** Client accessibility hint */
+    IPerm perm;
+    /** Current max time to change, secs */
+    double timeout;
+    /** current property state */
+    IPState state;
+    /** Widget comprising this vector */
+    IWidget *widets;
+    /** Dimension of widgets[] */
+    int  nwidgets;
+    /** ISO 8601 timestamp of this event */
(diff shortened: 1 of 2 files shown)


knro:
Thanks! I can see how this could be useful, but is this intended an an internal structure unavailable to driver that is used to simplify the code? All changes have to be in line with the INDI wire protocol.

pawel-soja:
First I would like to clean up what is now without modifying the structures.
Each "widget" has the same header structure, so if you want to read it, you don't need to cast to a particular type.
Therefore, the new structures are only for 'static_cast'.

The next step I would suggest is to get rid of std::vector searching, and use std::map for example.

The above modifications will increase the readability of "defaultdevice", "basedevice" etc. Reduce the amount of code, using standard C++11 efficient containers.
It will not change the way you write drivers.

After cleaning, I suggest you implement a convenient API (only C++) for writing drivers, leaving the old compatible of course.
A new way of creating buttons etc, would make it possible to attach handlers in the form of a class method or a lambda function.
The new API will use the lowest layer of the previous one so as not to duplicate the code.

In summary, all the modifications I would make will be backwards compatible.
They will concern the internal structure of the library and drivers.

If you mean XML when you say "INDI wire protocol", rest assured, nothing will change.

Indi Library looks a bit like a patch on the patch.
I am still new, are there any bigger plans for the development of the project?
I really like Indi/KStars - all in one place.

When I bought my first telescope, I was prepared to create the entire framework, I didn't know that Indi / KStars exists ;)
If you like my job, I can continue in the long term.

knro:
Thank you for the clarification. Some structures like ISwitchVectorProperty have different structure, how would the static casting work?

pawel-soja:
Thank you! I need to buy a bigger monitor.

I made the adjustments. Without the d-pointer, any modification to the class variables, all dependencies must be recompiled.

Is this hotfix acceptable? Is there something I don't know?

pawel-soja:
Where are we standing? Is my proposal up to date? I would like to go on.

While analyzing the code, I found something disturbing.

The variable n is incremented with each iteration of the XML element. However, it is only used for specific tags.
https://github.com/indilib/indi/blob/c80fa88163bb5681af5698bbd4bd30e4c5fd7b0b/libs/indibase/basedevice.cpp#L641-L685

knro:
Thanks for the heads up. Give me until tomorrow to review all of this again. Will also check the issue reported above. Too much work today.

knro:
I agree with you on nXMLEle should be sufficient. The code from 2003 onward was making heavy use of realloc all over the place as you can see. A separate MR for this would be appreciated.

knro:
After reviewing the patch, here are my primary concerns:

1. Re-arranging the struct members might actually break C code that uses aggregate initialization. This style is no longer used in libindi or indi-3rdparty, but it could be still used by external drivers written in C.
2. Introducing the un_used elements in ILightVectorProperty might have the same effect as well as #1.

Both of these issues are not show stoppers, but I'd like to see the input from other core developers as well before we can proceed.

@iliaplatone @d33psky @jochym @TallFurryMan Please share your thoughts.

iliaplatone:
If I can say my opinion, there's the chance to ease up coding with this patch, c drivers and clients could use a separate API with C wrappers or context API by internally call c++ methods.
The API should be simplified, yes, but needs also to be C portable, for both clients and drivers, also just to invoke INDI from different languages (JNI, P/Invoke, Python invoke etc..) and at first sight this patch seemed was made for this.

pawel-soja:
Hi!

**Don't merge it!**

I analyzed the library and drivers carefully.
Now I understand everything. It will be a huge problem if someone used the library outside the project.
I have a much better idea.
I'll be back soon.

Thank you all!

iliaplatone:
Yes, it's much better this way.. I watched indicom, the functions I worked on, allright.
The rest of the code is much more readable.
So in future we'd use vectors instead o mallocs right?

PR #1288, opened 2020-12-15
use of the common mkpath function

pawel-soja:
I unified the various libraries using INDI::mkpath (IndiUtility)


knro:
Awesome, looks good!!

Issue #1262, opened 2020-11-17
Primaluce Sesto Senso 2 - user profile / hold current control support


gilesco:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2420%20Funded-%2300A156.svg" />](https://issuehunt.io/r/indilib/indi/issues/1262)

Some telescopes require a Hold Current to be set in order to prevent slippage. The Sesto Senso 2 supports setting such a current, but it is not available to be set in the driver.
Currently the driver loads the factory preset "Slow" profile and this is hardcoded into the driver source code.

Ideally, I would like to be able to configure the all available settings through the driver, and also be able to load and save user profiles (There are 3 user profile slots available).
If not possible, at least be able to select an existing user profile (we can use the Wifi connection to create and configure these profiles.

Required Functionality: Load, Edit, Save the 3 user customisable profiles, and all their related settings (specifically Hold Current needed, but speed, acceleration / deceleration also beneficial).

Workaround for the time being:
I am yet to use the Sesto Senso 2, but I intend to create my own profile and then modify the code to select it instead of the "slow" profile.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $20.00)

- [<img src='https://avatars2.githubusercontent.com/u/63476340?v=4' alt='gilesco' width=24 height=24> gilesco](https://issuehunt.io/u/gilesco) ($20.00)

### Submitted pull Requests
- [#1315 Sesto Senso 2: Motor settings](https://issuehunt.io/r/indilib/indi/pull/1315)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/indilib/indi/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

knro:
Thank you for the feedback. Can you submit a PR for this?

gilesco:
Ah, I'm not really sure of how to implement any functionality, all I know is that if I set up a user profile called "Esprit" outside of the driver that I can substitute "slow" in line 1028 of sestosenso2.cpp for "Esprit" and it will probably load my profile instead of the slow profile:

bool CommandSet::loadSlowPreset(char *res)
{
    return sendCmd("{\"req\":{\"cmd\":{\"RUNPRESET\":\"Esprit\"}}}", "RUNPRESET", res);
}

On my unit {"req":{"get": ""}}} results with:

[...]

PR #1289, opened 2020-12-16
Sesto Senso 2: Support for displaying 12V input voltage.

lantictac:
Added warning and alert state if 12V power is insufficent.
Voltage polled alongside temprature.


knro:
Looks good! Great work for 1st patch!

lantictac:
Thanks Jasem! Going to have a go at implementing presets support and perhaps controlling the hold voltage.

gilesco:
If you get that working then be sure to close #1262 as I've donated $20 to whoever gets it done 🥇

PR #1290, opened 2020-12-17
 Sesto Senso 2: Log & display firmware version

lantictac:
Also moved firmware info to Info tab to declutter main controls tab a little.


lantictac:
Actually, I just realized I should be working on a feature branch...

Issue #1270, opened 2020-11-30
indi_lx200_10micron: setting Long in telescope location


fenriques:
On a 10Micron mount, but probably on other drivers as well, when entering a Longitude value in 'Site Management', the driver updates the mount coordinates but skips the seconds (ss) value.
For example, entering '02:38:30' in Ekos, will results in 'E002 39 00' (and not 'E002 38 30') on the 10micron handpad.

Latitude behavior is different: seconds are stored correctly.

Looking in the INDI log file, the reason seems to be different decimal places for Long (5) and Lat (only 2) when setting the **Observer location** (second line of the log below).

`[2020-11-27T19:03:52.453 CET INFO ][           org.kde.kstars.indi] - LX200 10micron :  "[INFO] Site location updated to Lat  37:44:24 - Long    2:38:30 "`
`[2020-11-27T19:03:52.458 CET INFO ][           org.kde.kstars.indi] - LX200 10micron :  "[INFO] Observer location updated: Longitude (2.63556) Latitude (37.74) "`

Tested using Kstars 3.50, Indi 1.8.7 but was also on previous versions.

I wasn't able to find in the source code where the Observer location is set.
Ferrante

knro:
Thank you for the report. The reason is that 10Micron is based off LX200 command set. If you check in [lx200driver:1020](https://github.com/indilib/indi/blob/5abd3d5cb7f39f1411c4647b8f5cfc51e82f0b82/drivers/telescope/lx200driver.cpp#L1020) you'll see it only sets degrees and minutes. The 10Micron driver should probably override [updateLocation](https://github.com/indilib/indi/blob/5abd3d5cb7f39f1411c4647b8f5cfc51e82f0b82/drivers/telescope/lx200telescope.h#L89) function to send the full sexagesimal representation correctly to the mount.

fenriques:
thanks Jasem. But if nothing else breaks, wouldn't it be better to fix  lx200driver::setSiteLongitude method directly? 
I'm not much into this class hierarchy, but other lx200-based drivers are affected as well by this issue.

knro:
The Deg:Min convention is used by the official LX200 command set. It's possible to perhaps check if long format is supported and then send the seconds accordingly like it's done for the equatorial coordinates.

knro:
@d33psky Please check this out

d33psky:
I'll be using the `controller_format` to switch styles (short, long, longer formats currently)

d33psky:
[...]

PR #1296, opened 2020-12-23
Fix for #1270, use d m s.f with f being tenths for LX200_GEO_LONGER_FORMAT

d33psky:
This allows to set the higher precision on the mount : 
```
#:Gg#
-005*32#

#:U1#
#:Gg#
-005*31:46#

#:U2#
#:Gg#
-005:31:46.2#
```
These values however are not shown yet on the Site Management (seconds are, fractions of seconds are not).
I'll add this later and update this PR.

### DIFF
--- a/drivers/telescope/lx200driver.cpp
+++ b/drivers/telescope/lx200driver.cpp
@@ -46,7 +46,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110 - 1301  USA
 #define RB_MAX_LEN    64
 
 
-int controller_format;
+int controller_format; /* For possible values see enum TFormat */
 char lx200Name[MAXINDIDEVICE];
 /* ESN DEBUG */
 unsigned int DBG_SCOPE = 8;
@@ -80,9 +80,9 @@ int getCommandInt(int fd, int *value, const char *cmd);
 /* Get tracking frequency */
 int getTrackFreq(int fd, double *value);
 /* Get site Latitude */
-int getSiteLatitude(int fd, int *dd, int *mm);
+int getSiteLatitude(int fd, int *dd, int *mm, double *ssf);
 /* Get site Longitude */
-int getSiteLongitude(int fd, int *ddd, int *mm);
+int getSiteLongitude(int fd, int *ddd, int *mm, double *ssf);
 /* Get Calender data */
 int getCalendarDate(int fd, char *date);
 /* Get site Name */
@@ -119,7 +119,7 @@ int setUTCOffset(int fd, double hours);
 /* Set Track Freq */
 int setTrackFreq(int fd, double trackF);
 /* Set current site longitude */
-int setSiteLongitude(int fd, double Long);
+int setSiteLongitude(int fd, double CartographicLongitude);
 /* Set current site latitude */
 int setSiteLatitude(int fd, double Lat);
 /* Set Object Azimuth */
@@ -534,8 +534,17 @@ int getSiteName(int fd, char *siteName, int siteNum)
     return 0;
 }
 
-int getSiteLatitude(int fd, int *dd, int *mm)
+int getSiteLatitude(int fd, int *dd, int *mm, double *ssf)
 {
+    // :Gt# from 10Micron docs explaining the extensions to the standard LX200 protocol.
+    // Get current site latitude.
+    // Returns the latitude of the current site formatted as follows:
+    // Emulation and precision              Return value
+    // Any emulation, low precision         sDD*MM# (sign, degrees, minutes)
+    // LX200 emulation, high precision      sDD*MM# (sign, degrees, minutes)
+    // Extended emulation, high precision   sDD*MM:SS# (sign, degrees, arcminutes, arcseconds)
+    // Any emulation, ultra precision       sDD:MM:SS.S# (sign, degrees, arcminutes, arcseconds, tenths of arcsecond)
+    // Positive implies north latitude.
     DEBUGFDEVICE(lx200Name, DBG_SCOPE, "<%s>", __FUNCTION__);
     char read_buffer[RB_MAX_LEN]={0};
     int error_type;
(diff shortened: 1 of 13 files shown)


d33psky:
I'd like to set the WIP and DO NOT MERGE flags but cannot. Can this access be changed ?

knro:
It's marked as WIP when I just checked.

d33psky:
weird, I only see tag 'Proposal', and I cannot edit that either.

knro:
IIRC, LX200 only uses d:m for the geographic coordinates regardless of the format used for equatorial coordinates. i.e. SHORT and LONG equatorial coordinates do not affect geographic coordinates, they're still d:m and not d:m:s

d33psky:
10Micron has this on it : 
```
:Gt#
Get current site latitude. Returns the latitude of the current site formatted as follows:
Emulation and precision              Return value
Any emulation, low precision         sDD*MM# (sign, degrees, minutes)
LX200 emulation, high precision      sDD*MM# (sign, degrees, minutes)
Extended emulation, high precision   sDD*MM:SS# (sign, degrees, arcminutes, arcseconds)
Any emulation, ultra precision       sDD:MM:SS.S# (sign, degrees, arcminutes, arcseconds, tenths of arcsecond)
Positive implies north latitude.
```
I'll play with sscanf to match all and propose some code that should not break anything.

d33psky:
@knro ready for review. Also @ee5cd7d5d would appreciate if you could review too :)

knro:
Thanks! Any drivers in INDI 3rdparty are also affected by this?

drivers/telescope/lx200driver.cpp
```diff
@@ -562,19 +571,35 @@ int getSiteLatitude(int fd, int *dd, int *mm)
 
     DEBUGFDEVICE(lx200Name, DBG_SCOPE, "RES <%s>", read_buffer);
 
-    if (sscanf(read_buffer, "%d%*c%d", dd, mm) < 2)
+    *ssf = 0.0;
+    if (sscanf(read_buffer, "%d%*c%d:%lf", dd, mm, ssf) < 2)
```
ee5cd7d5d, line 575: Shouldn't this be %.1lf? Only one decimal place?

drivers/telescope/lx200driver.cpp
```diff
@@ -599,13 +624,15 @@ int getSiteLongitude(int fd, int *ddd, int *mm)
 
     DEBUGFDEVICE(lx200Name, DBG_SCOPE, "RES <%s>", read_buffer);
 
-    if (sscanf(read_buffer, "%d%*c%d", ddd, mm) < 2)
+    *ssf = 0.0;
+    if (sscanf(read_buffer, "%d%*c%d:%lf", ddd, mm, ssf) < 2)
```
ee5cd7d5d, line 628: %.1lf?

drivers/telescope/lx200telescope.cpp
```diff
@@ -640,7 +640,7 @@ bool LX200Telescope::updateLocation(double latitude, double longitude, double el
     if (isSimulation())
         return true;
 
-    if (!isSimulation() && setSiteLongitude(PortFD, 360.0 - longitude) < 0)
+    if (!isSimulation() && setSiteLongitude(PortFD, longitude) < 0)
```
ee5cd7d5d, line 643: Is 360.0 - longitude the same as flipping the sign?

ee5cd7d5d (commented):
Works for me! I might test it on my hardware too, if I come to it :)

drivers/telescope/lx200driver.cpp
d33psky: well because the third parameter is completely optional I did not want to limit significance there. If there's a ss.ss then it can be used for rounding at least of the last .s digit. So I deliberately left out the 1. Does this make sense to you ?

drivers/telescope/lx200telescope.cpp
d33psky: this `360 -` actually caused problems for me, my longitude is -5 , and then this results in 365 degrees which my mount refuses to set. There were more places with longitude 'calculations' and these became only visible when I upgraded the logging to always log both sexagesimal and the double. The visual display in KStars uses carthographic coordinates because that is what users read on maps and get from GPS, INDI internally converted that halfway to 0-360 which is actually not needed for the api in the first place. The new standard is carthographic everywhere so that my brain does not hurt, and convert for lx200 during actual writing to and reading from the mount. This conversion is the double * -1.

d33psky:
I checked 3rdparty, There's two LX200 drivers in there and both are not affected by these changes. 
LX200Skywalker::setSiteLongitude already does its own ":Sg%03d*%02d:%02d#", d, m, s as an improvement over LX200Generic.
LX200StarGo::setSiteLongitude does ":Sg+%03d*%02d:%02d#", d, m, s, same improvement. And it does its own mapping to have longitude between -180 and +180.

drivers/telescope/lx200driver.cpp
d33psky: same here

d33psky:
That is unfortunate. Where does it say that ?
In any case users will expect to be able just put in their GPS coordinates.
We can at least extend it to also accept those.
So from -180 to 360 then ?

drivers/telescope/lx200driver.cpp
ee5cd7d5d: Right! Yes makes sense

drivers/telescope/lx200telescope.cpp
ee5cd7d5d: Aaaah yeah. I see, I did understand your logic to be honest (conversion from cartographic to lx200) I did not understand the previous conversion (the 360-) but since it was working before I had to ask :)

knro:
This is the standard since 2003 as defined in [INDI Standard Properties](https://indilib.org/developers/developer-manual/101-standard-properties.html#h2-general). So clients usually take care of sending the location in the correct format, the user doesn't usually enter it manually.

d33psky:
> IIRC, LX200 only uses d:m for the geographic coordinates regardless of the format used for equatorial coordinates. i.e. SHORT and LONG equatorial coordinates do not affect geographic coordinates, they're still d:m and not ds

I forgot about that ^ remark. I'm re-using the equatorial coordinates precision for the geographic coordinates precision where I should not. I'll change the code today to actually detect the geographic format on a read and use that for the writes. That should work everywhere and even follow live precision changes (U U1 U2) that may be set elsewhere.

d33psky:
If ^ is OK I'll also make a PR for indi-3rdparty/ to rename checkLX200Format there 6 times.

libs/indibase/inditelescope.cpp
```diff
@@ -104,7 +104,7 @@ bool Telescope::initProperties()
     IUFillTextVector(&TimeTP, TimeT, 2, getDeviceName(), "TIME_UTC", "UTC", SITE_TAB, IP_RW, 60, IPS_IDLE);
 
     IUFillNumber(&LocationN[LOCATION_LATITUDE], "LAT", "Lat (dd:mm:ss.s)", "%012.8m", -90, 90, 0, 0.0);
-    IUFillNumber(&LocationN[LOCATION_LONGITUDE], "LONG", "Lon (dd:mm:ss.s)", "%012.8m", -180, 180, 0, 0.0);
+    IUFillNumber(&LocationN[LOCATION_LONGITUDE], "LONG", "Lon (dd:mm:ss.s)", "%012.8m", -180, 360, 0, 0.0);
```
knro, line 107: This cannot be used as well, it's 0 to 360 as the standard properties specify. All drivers follow this convention.

d33psky: OK, set back to 0-360 which is what it was.

d33psky:
I think this PR is ready to be merged in.

knro:
Thanks, I'll go ahead with the merge. Please send a PR for 3rd party to make it work with this change.

d33psky:
Done -> https://github.com/indilib/indi-3rdparty/pull/272

PR #1300, opened 2020-12-26
Refactor/satellite tracking properties -> Moving properties to a higher abstraction layer

ee5cd7d5d:
I moved satellite tracking properties to `INDI::Telescope` as suggested [here](https://invent.kde.org/education/kstars/-/merge_requests/169). I refactored the `indi_lx200_10micron` implementation to accommodate the property changes.
@d33psky I have seen you have recently implemented changes in other aspects of the driver. I would appreciate some feedback :)


libs/indibase/inditelescope.h
```diff
@@ -781,6 +789,20 @@ class Telescope : public DefaultDevice
 
         const char * getPierSideStr(TelescopePierSide ps);
 
+        // Satellite tracking
+        IText TLEtoTrackT[1] {};
+        ITextVectorProperty TLEtoTrackTP;
+        IText SatPassWindowT[2] {};
+        ITextVectorProperty SatPassWindowTP;
+        enum
```
d33psky, line 797: let's name this enum

drivers/telescope/lx200_10micron.cpp
```diff
@@ -60,7 +58,7 @@ LX200_10MICRON::LX200_10MICRON() : LX200Generic()
 
     SetTelescopeCapability( TELESCOPE_CAN_GOTO | TELESCOPE_CAN_SYNC | TELESCOPE_CAN_PARK | TELESCOPE_CAN_ABORT |
         TELESCOPE_HAS_TIME | TELESCOPE_HAS_LOCATION | TELESCOPE_HAS_PIER_SIDE |
-        TELESCOPE_HAS_TRACK_MODE | TELESCOPE_CAN_CONTROL_TRACK | TELESCOPE_HAS_TRACK_RATE );
+        TELESCOPE_HAS_TRACK_MODE | TELESCOPE_CAN_CONTROL_TRACK | TELESCOPE_HAS_TRACK_RATE | TELESCOPE_CAN_TRACK_SATELLITE );
```
d33psky, line 61: I'll move these flags to one per line later.

drivers/telescope/lx200_10micron.cpp
```diff
@@ -784,19 +757,30 @@ bool LX200_10MICRON::CalculateTrajectory(int year, int month, int day, int hour,
...
-                year, month, day, hour, minute, nextpass);
-    date_pass.years = year;
-    date_pass.months = month;
-    date_pass.days = day;
-    date_pass.hours = hour;
-    date_pass.minutes = minute;
-    date_pass.seconds = 0.0;
-    JD = ln_get_julian_day(&date_pass);
+    struct ln_date start_pass;
+    if (extractISOTime(start_pass_iso, &start_pass) == -1)
```
d33psky, line 761: does this thing require HAVE_LIBNOVA ?

indicom.h has it in #ifndef _WIN32
indicom.c has it in #if defined(HAVE_LIBNOVA)

drivers/telescope/lx200_10micron.cpp
```diff
@@ -771,7 +744,7 @@ bool LX200_10MICRON::SetTLEfromDatabase(int tleN)
     return 0;
 }
 
-bool LX200_10MICRON::CalculateTrajectory(int year, int month, int day, int hour, int minute, int nextpass, ln_date date_pass)
+bool LX200_10MICRON::CalculateSatTrajectory(char *start_pass_iso, char *end_pass_iso)
```
d33psky, line 747: good to see ISO8601 in use :) maybe name it start_pass_isodatetime for clarity ?

drivers/telescope/lx200_10micron.cpp
```diff
@@ -1184,9 +1150,9 @@ bool LX200_10MICRON::ISNewSwitch(const char *dev, const char *name, ISState *sta
             IDSetSwitch(&AlignmentSP, nullptr);
             return true;
         }
+        if (strcmp(TrackSatSP.name, name) == 0)
+          {
```
d33psky, line 1154: spacing

drivers/telescope/lx200_10micron.cpp
```diff
@@ -1274,27 +1240,51 @@ bool LX200_10MICRON::ISNewText(const char *dev, const char *name, char *texts[],
...
             }
+
+           return true;
+        }
+        if (strcmp(name, "SAT_PASS_WINDOW") == 0)
+        {
+
+          IUUpdateText(&SatPassWindowTP, texts, names, n);
+          if (0 == CalculateSatTrajectory(SatPassWindowT[0].text, SatPassWindowT[1].text))
+            {
```
d33psky, line 1273: spacing

drivers/telescope/lx200_10micron.h
```diff
@@ -138,9 +138,7 @@ class LX200_10MICRON : public LX200Generic
     bool SetTLEtoFollow(const char *tle);
     bool SetTLEfromDatabase(int tleN);
     bool TrackSat();
-    bool CalculateTrajectory(int year, int month, int day, int hour, int minute, int nextpass, ln_date date_pass);
-    struct ln_date today;
-    struct ln_date date_pass;
+    bool CalculateSatTrajectory(char *start_pass, char *end_pass);
```
d33psky, line 141: let's use the same names as in the implementation

d33psky (commented):
Nice work :)
I left some comments and questions

libs/indibase/inditelescope.h
ee5cd7d5d: Done

drivers/telescope/lx200_10micron.cpp
ee5cd7d5d: I went ahead and did it :)

drivers/telescope/lx200_10micron.cpp
ee5cd7d5d: It does require HAVE_LIBNOVA, in order to use `ln_date` and `ln_get_julian_day`. What should I do in this respect?

drivers/telescope/lx200_10micron.cpp
ee5cd7d5d: Done

drivers/telescope/lx200_10micron.cpp
ee5cd7d5d: Which spacing? If its the one before `IUUPdateText`, just fixed it :+1:

drivers/telescope/lx200_10micron.cpp
ee5cd7d5d: ?

drivers/telescope/lx200_10micron.h
ee5cd7d5d: Done

drivers/telescope/lx200_10micron.cpp
d33psky: the { should move 2 positions to the left

drivers/telescope/lx200_10micron.cpp
d33psky: I think the `#ifndef _WIN32` in indicom.h should be replaced with `#if defined(HAVE_LIBNOVA)` and the extractISOTime calls should have `#if defined(HAVE_LIBNOVA)` as well. @knro do you agree ?

drivers/telescope/lx200_10micron.cpp
d33psky: thanks :)

drivers/telescope/lx200_10micron.cpp
knro: Right, we use astyle for the formatting. It's documented in the README.md in the Github page.

drivers/telescope/lx200_10micron.cpp
```diff
@@ -771,7 +754,7 @@ bool LX200_10MICRON::SetTLEfromDatabase(int tleN)
     return 0;
 }
 
-bool LX200_10MICRON::CalculateTrajectory(int year, int month, int day, int hour, int minute, int nextpass, ln_date date_pass)
+bool LX200_10MICRON::CalculateSatTrajectory(char *start_pass_isodatetime, char *end_pass_isodatetime)
```
knro, line 757: const char * or better const std::string & if you can

libs/indibase/inditelescope.cpp
```diff
@@ -170,6 +170,28 @@ bool Telescope::initProperties()
...
+        IUFillText(&TLEtoTrackT[0], "TLE", "TLE", "");
+        IUFillTextVector(&TLEtoTrackTP, TLEtoTrackT, 1, getDeviceName(), "SAT_TLE_TEXT", "Orbit Params", SATELLITE_TAB, 
+        IP_RW, 60, IPS_IDLE);
+
+        char curTime[32] = {0};
+        std::time_t t = std::time(nullptr);
+        struct std::tm *utctimeinfo = std::gmtime(&t);
+        strftime(curTime, sizeof(curTime), "%Y-%m-%dT%H:%M:%S", utctimeinfo);
+        
+        IUFillText(&SatPassWindowT[1], "SAT_PASS_WINDOW_END", "End UTC", curTime);
```
knro, line 184: use enum instead of 0 and 1 indexes. I think I'll require enums for all new code.

libs/indibase/inditelescope.h
```diff
@@ -781,6 +789,20 @@ class Telescope : public DefaultDevice
...
         const char * getPierSideStr(TelescopePierSide ps);
 
+        // Satellite tracking
+        IText TLEtoTrackT[1] {};
+        ITextVectorProperty TLEtoTrackTP;
+        IText SatPassWindowT[2] {};
+        ITextVectorProperty SatPassWindowTP;
+        enum SATELLITE_TRACKING
+        {
+            SAT_TRACK,
```
knro, line 799: Please document the enum using Doxygen style coding.

Also, I think the class itself should probably document how this is used with a simple example. This is to allow prospective developers from understand what these properties are exactly for and how to utilize them in their own drivers.

knro (changes requested):
Few minor changes and documentation updates.

libs/indibase/inditelescope.h
ee5cd7d5d: Sure! Is a docstring fine regarding the example? Or is it maybe better to just point at the implementation in the LX200 so they see how it was done in that case?

drivers/telescope/lx200_10micron.cpp
ee5cd7d5d: Done

drivers/telescope/lx200_10micron.cpp
ee5cd7d5d: I formatted the entire file using astyle and the config stated in the README. Let me know if something was done incorrectly

libs/indibase/inditelescope.cpp
ee5cd7d5d: That's great! No magic numbers is always better. I created an enum and documented it

drivers/telescope/lx200_10micron.cpp
ee5cd7d5d: I can do that, it's just two lines. I am just unsure what would be the behavior in that case: if it does not have LIBNOVA it should simply return an error `LOG_ERROR("Libnova not available, can't process date")` or something like that? Or do you have some other suggestion?

knro:
All drivers have access to libnova. It's just that a client utilizing indicom might not have access to libnova. So as far DRIVERS are concerned, libnova exists.

libs/indibase/inditelescope.h
ee5cd7d5d: I went ahead and added docstrings with a `\ref` to the LX200_10micron implementation. Let me know if you think that is sufficient

drivers/telescope/lx200_10micron.cpp
knro: No need. As I said, all drivers have access to libnova.

PR #1302, opened 2020-12-28
Indibase/refactoring - update INDI::Property

pawel-soja:
Hi!

I found an unfinished "INDI::Property" class. I have moved some of the scattered code into this class.

For those who are bored:
I am thinking to use "BaseDevice" and "Property" to implement the idea:
https://github.com/pawel-soja/indi/tree/widgets/examples/example_widgets

### DIFF
--- a/indidriver.c
+++ b/indidriver.c
@@ -168,7 +168,7 @@ static char *escapeXML(const char *src, size_t size)
 /* tell Client to delete the property with given name on given device, or
  * entire device if !name
  */
-void IDDelete(const char *dev, const char *name, const char *fmt, ...)
+void IDDeleteVA(const char *dev, const char *name, const char *fmt, va_list ap)
 {
     pthread_mutex_lock(&stdout_mutex);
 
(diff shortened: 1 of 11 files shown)


knro:
Great! Are you thinking of creating Qt-like widgets with signals/slots? That would be awesome!

Btw, have you tested this with drivers that utilize dynamic properties (read from skeleton XML)? Like EQMod and induino drivers.

pawel-soja:
Yes, I would implement something like Qt, but using std::function. It gives great opportunities to manipulate arguments.
First, I want to clean up as much code as possible so as not to miss something when implementing widgets.

indidriver.c
```diff
@@ -2313,13 +2366,12 @@ void IDSetLight(const ILightVectorProperty *lvp, const char *fmt, ...)
...
-        va_start(ap, fmt);
         char message[MAXINDIMESSAGE];
-        printf("  message='");
         vsnprintf(message, MAXINDIMESSAGE, fmt, ap);
+
+        // #PS: why not escapeXML_fputs?
+        printf("  message='");
         printf("%s'\n", entityXML(message));
-        va_end(ap);
     }
```
pawel-soja, line 2375: Sometimes the "entityXML" function is used,
sometimes "escapeXML_fputs", but nothing here.
Right?

With the next review, I can improve it.

indidriver.c
```diff
@@ -2354,12 +2413,10 @@ void IDSetBLOB(const IBLOBVectorProperty *bvp, const char *fmt, ...)
...
     if (fmt)
     {
-        va_list ap;
-        va_start(ap, fmt);
         printf("  message='");
+        // #PS: missing escapeXML_fputs ?
         vprintf(fmt, ap);
         printf("'\n");
-        va_end(ap);
     }
```
pawel-soja, line 2420: Sometimes the "entityXML" function is used,
sometimes "escapeXML_fputs", but nothing here.
Right?

With the next review, I can improve it.

knro:
Any issues with EQMod and/or induino drivers with this change?

knro:
While checking with INDI EQMod, KStars just crashed in customerproperties.cpp. Still investigating the issue.

pawel-soja:
I missed something. I'll fix it.

pawel-soja:
I tested indi_eqmod_telescope and indi_duino with KStars (no hardware).
It looks OK to me.
Have you recompiled indi-3rdparty?
Can you give more details?

knro:
Looks like it was a KStars bug. Ok thanks, I will merge this now as I ran all the tests and it looks OK. I will test in the live observatory now as well. Good work, this is a very good step to migrating the rest oft he code to modern C++!

rlancaste:
A user in the INDI forum just reported that a change made in this merge broke the OS X build.  I just tested and confirmed it.  The issue appears to be a change you made near the top of indiproperty.cpp.  Here is the compile time error message:  "error: aliases are not supported on darwin"

knro:
@pawel-soja Can you please check?

PR #1307, opened 2020-12-31
Basedevice/refactoring - Another small step into the future.

pawel-soja:
Hi Jasem!

I cleaned up a bit. I have separated a directory where I will implement high-level widget/property.
Fixed a bug with counting correct elements from XML ('n' variable).
This is probably all this year.

Happy New Year!
Paweł Soja


knro:
Thank you Pawel, great work as always! I am happy to report that I tested all your changes yesterday in the observatory and it all worked without issues. I will keep testing today if the weather is clear. Happy new year!

PR #1308, opened 2020-12-31
Add new helper methods to the AlignmentSubsystemForDrivers (Alt/Az)

rickbassham:
Now adding the helper methods for Alt/Az mounts. I would appreciate if someone with an Alt/Az mount could test this. The test cases I wrote seem to work well, though.

```cpp
      /** \brief Adds an alignment point to the model database, usually called from Sync.
       * \param[in] actualRA actual Right Ascension in decimal hours
       * \param[in] actualDec actual Declination in decimal degrees
       * \param[in] mountAlt Altitude where the mount thinks it is in decimal degrees
       * \param[in] mountAz Azimuth where the mount thinks it is in decimal degrees
       * \return true if the alignment point was added to the database, otherwise false
       *
       * This will return false if either the alignment point was already added, or if the location
       * is not set. Call UpdateLocation to set the current location.
       */
      bool AddAlignmentEntryAltAz(double actualRA, double actualDec, double mountAlt, double mountAz);

      /** \brief Converts an actual sky location to coordinates to send to the mount, usually called
       * in Goto.
       * \param[in] actualRA actual Right Ascension in decimal hours
       * \param[in] actualDec actual Declination in decimal degrees
       * \param[out] mountAlt Altitude to send to the mount
       * \param[out] mountAz Azimuth to send to the mount
       * \return true if we converted actualRA/actualDec to mountAlt/mountAz, otherwise false
       *
       * This will return false if we have fewer than 2 alignment points added, or if the location
       * is not set. Call UpdateLocation to set the current location.
       */
      bool SkyToTelescopeAltAz(double actualRA, double actualDec, double &mountAlt, double &mountAz);

      /** \brief Converts a mount location to actual sky coordinates, usually called in ReadScopeStatus.
       * \param[in] mountAlt Altitude where the mount thinks it is in decimal degrees
       * \param[in] mountAz Azimuth where the mount thinks it is in decimal degrees
       * \param[out] actualRA actual Right Ascension in decimal hours
       * \param[out] actualDec actual Declination in decimal degrees
       * \return true if we converted mountRa/mountDec to actualRA/actualDec, otherwise false
       *
       * This will return false if we have fewer than 2 alignment points added, or if the location
       * is not set. Call UpdateLocation to set the current location.
       */
      bool TelescopeAltAzToSky(double mountAlt, double mountAz, double &actualRA, double &actualDec);
```

### DIFF
--- a/libs/indibase/alignment/AlignmentSubsystemForDrivers.cpp
+++ b/libs/indibase/alignment/AlignmentSubsystemForDrivers.cpp
@@ -170,6 +170,96 @@ namespace INDI
             return false;
         }
 
+        bool AlignmentSubsystemForDrivers::AddAlignmentEntryAltAz(double actualRA, double actualDec, double mountAlt, double mountAz)
+        {
+            ln_lnlat_posn location;
+            if (!GetDatabaseReferencePosition(location))
+            {
+                return false;
+            }
+
+            struct ln_hrz_posn AltAz
+            {
+                0, 0
+            };
+            AltAz.alt = range360(mountAlt);
+            AltAz.az = range360(mountAz);
+
+            AlignmentDatabaseEntry NewEntry;
+            TelescopeDirectionVector TDV = TelescopeDirectionVectorFromAltitudeAzimuth(AltAz);
+
+            NewEntry.ObservationJulianDate = ln_get_julian_from_sys();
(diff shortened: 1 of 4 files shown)


knro:
@jochym @xsnrg Can you please check?

PR #1309, opened 2021-01-03
add new ccd rotation prop to ccd simulator

rickbassham:
To make testing https://invent.kde.org/education/kstars/-/issues/56 easier, add a manual rotation prop to the ccd simulator.


knro:
This rotation sim property was actually removed from the simulator a while ago! It's now in the ABS_ROTATOR_ANGLE property in INDI::Rotator which INDI::CCD monitors. So for your testing, you can add a Rotator Simulator to your equipment profile and change the value as desired. It would have the same effect.

rickbassham:
Since we are trying to do manual rotations with that ticket, the code in Ekos will use the rotator to do the rotation, so having a setting you can change manually would still be needed to test it.

rickbassham:
And the snoop on the rotator will still work with this change. This just gives us another way to change it.

knro:
Alright, merged!

PR #1312, opened 2021-01-05
fix: inline methods that use unlinked functions

pawel-soja:
Is it possible to add automated tests for OS X?

I think I fixed the weak linking problem.
If the methods are not used (apply / define / save), there will be no reference to functions that are compiled only for drivers.

Additionally, I prepared the class for inheritance.
In some places, strongly typed, it is unnecessarily reduced to the INDI::Property type.
Later I will add another layer as template for IText/INumber/ISwitch/ILight/IBLOB that inherits from INDI::Property.

### DIFF
--- a/libs/indibase/property/indiproperty.cpp
+++ b/libs/indibase/property/indiproperty.cpp
@@ -122,9 +122,11 @@ Property::Property(IBLOBVectorProperty   *property)
 { }
 
 Property::~Property()
-{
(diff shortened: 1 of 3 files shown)


knro:
@rlancaste Can you please check if this works fine on MacOS?

rlancaste:
Yes, this built just fine

PR #1313, opened 2021-01-06
Indidriver/refactoring

pawel-soja:
The most important change is moving the ROSC implementation to one place. This is a step to get rid of unnecessary dependencies between source files. Additionally, it will make it easier to change the method of storing items (list or map).

### DIFF
--- a/eventloop.c
+++ b/eventloop.c
@@ -165,7 +165,7 @@ int addCallback(int fd, CBF *fp, void *ud)
             break;
     if (cp == &cback[ncback])
     {
-        cback = cback ? (CB *)realloc(cback, (ncback + 1) * sizeof(CB)) : (CB *)malloc(sizeof(CB));
+        cback = realloc(cback, (ncback + 1) * sizeof(CB));
         cp    = &cback[ncback++];
     }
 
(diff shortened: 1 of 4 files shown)


pawel-soja:
Is there any reason to keep eg the indidriver in C? Can I convert to C ++ while keeping the headers for C?

knro:
Wouldn't this break all C-based drivers? Not that we have any in the repo AFAIK, but perhaps drivers written and maintained externally.

pawel-soja:
If a library is used outside of a project as shared (*.so), it doesn't matter what it's written in. It is worth considering such a transition to C++, because most of the code is the implementation of containers that unnecessarily rewrite the content each time an element is added.

knro:
Alright, I suppose the move the full C++ implementation is inevitable sooner or later. There is one outstanding issue that requires attention regarding INDI::BaseDevice and INDI::BaseClient handling for INDI::Property. Right now, the code uses raw pointers to create and pass INDI::Property to the client. Problem arises when a property is removed from the basedevice, and then the request is sent to the client to remove it.

Since each lives in its own thread, you end up with dangling pointers on the client side for properties removed in the basedevice. This can also lead to some race conditions that would cause possible deadlocks though this is mostly resolved. What would be the approach to handle this? I suppose shared_ptr would make more sense to use now than raw pointers to begin with, but perhaps we need to introduce better thread-safety for all the components in INDI.

pawel-soja:
I'll get it ;)
First, I will get rid of unnecessary dependencies between files.
I see that it was a big problem once and some functions ended up in the "indicom" file. Function declarations in files with a different name than the implementation are quite confusing.
BaseDevice is compiled with "indiclient" without zlib, where it is required for "BaseDevice::setBLOB".
CMakeLists would also need to be split into individual modules. The current version of the file increases compilation time unnecessarily.
A large number of these little bugs prevent me from continuing to work on Qt-Like widgets.

So let me push some minor tweaks to make this project happen. When the project is tidy, I'll redo it all to C++.

PR #1315, opened 2021-01-08
Sesto Senso 2: Motor settings

lantictac:
While this includes code for editing, saving and apply user presets, it seems there's an issue with the Sesto Senso 2 firmware v1.3 that prevents applying the user presets. So for now I've disabled the apply/save user preset switches until a firmware fix is confirmed. If the fix becomes available the functionality can be enabled by removing the two commented TODO code blocks within SestoSenso2::updateProperties().

The builtin defaults can be applied without issue.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#1262: Primaluce Sesto Senso 2 - user profile / hold current control support](https://issuehunt.io/repos/58881442/issues/1262)
---
</details>

### DIFF
--- a/drivers.xml
+++ b/drivers.xml
@@ -381,11 +381,11 @@
         </device>
         <device label="Sesto Senso 2" manufacturer="Primaluce Lab">
             <driver name="Sesto Senso 2">indi_sestosenso2_focus</driver>
-            <version>0.6</version>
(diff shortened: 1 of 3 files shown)


lantictac:
image
image
image

PR #1318, opened 2021-01-11
Refactoring - less dependencies between files / less mallocs

pawel-soja:
While browsing the sources, I immediately made corrections.

- less mallocs
- less unnecessary dependencies between files
- Separation of the common part in indidriver (escapeXML_vprintf)
- null-terminated strings in tty_error_msg
- others to improve readability

### DIFF
--- a/eventloop.c
+++ b/eventloop.c
@@ -42,6 +42,7 @@
 #include <sys/time.h>
 
 #include "eventloop.h"
+#include "indidevapi.h"
 
 /* info about one registered callback.
  * the malloced array cback is never shrunk, entries are reused. new id's are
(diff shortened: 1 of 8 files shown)


knro:
Excellent! Btw, regarding indicom, I created long ago a class for (serial connection)[https://github.com/indilib/indi/blob/master/libs/indibase/connectionplugins/ttybase.cpp] but I never had the chance to continue the work to get it adopted fully in the drivers. The base class implements the generic tty connection and then drivers that needs special behavior can subclass. Right now in indicom it's set globally and as a set of if/else which is bad.

@iliaplatone Can you please revise the changes?

PR #1320, opened 2021-01-12
Refactoring/INDI::BaseDevice - D-Pointer implemented

pawel-soja:
Hi!

I extracted the data of the INDI::BaseDevice class.
By the way, I improved a few places.


knro:
Bravo! I love the changes! Btw, I have been thinking of using shared_ptr for INDI::Property, thoughts?

pawel-soja:
Take it easy, in due time. I have an idea how to embrace the whole thing. I just need to separate the implementations well between the client and the server.

PR #1321, opened 2021-01-12
Add snoop device to joystick controller

rickbassham:
Add snoop device to joystick controller to allow devices not named "Joystick" to act as a joystick. Defaults to "Joystick", so it should not be a breaking change.

### DIFF
--- a/libs/indibase/indicontroller.cpp
+++ b/libs/indibase/indicontroller.cpp
@@ -97,8 +97,8 @@ bool Controller::initProperties()
     IUFillSwitchVector(&UseJoystickSP, UseJoystickS, 2, device->getDeviceName(), "USEJOYSTICK", "Joystick", OPTIONS_TAB,
                        IP_RW, ISR_1OFMANY, 0, IPS_IDLE);
 
-    IUFillText(&JoystickDeviceT[0], "JOYSTICK_DEVICE", "Device", "Joystick");
-    IUFillTextVector(&JoystickDeviceTP, JoystickDeviceT, 1, device->getDeviceName(), "JOYSTICK_DEVICE", "Snoop Joystick",
+    IUFillText(&JoystickDeviceT[0], "SNOOP_JOYSTICK_DEVICE", "Device", "Joystick");
+    IUFillTextVector(&JoystickDeviceTP, JoystickDeviceT, 1, device->getDeviceName(), "SNOOP_JOYSTICK", "Snoop Joystick",
                      OPTIONS_TAB, IP_RW, 60, IPS_IDLE);
 
     return true;
@@ -168,7 +168,7 @@ bool Controller::ISNewText(const char *dev, const char *name, char *texts[], cha
 {
     if (strcmp(dev, device->getDeviceName()) == 0)
     {
-        if (!strcmp(name, "JOYSTICK_DEVICE"))
+        if (!strcmp(name, "SNOOP_JOYSTICK"))
         {
             IUUpdateText(&JoystickDeviceTP, texts, names, n);
             JoystickDeviceTP.s = IPS_IDLE;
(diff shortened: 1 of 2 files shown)


rickbassham:
Not sure if it is related to this commit yet, but I'm getting a SEGFAULT in KStars in `void INDIListener::registerProperty(INDI::Property *prop)` when the telescope has saved it's options and joystick is enabled. Trying to track it down.

knro:
I've been trying to track this down. Make sure libindi is also updated and installed to the system then re-compile KStars. It's related to property getting removed from INDI::BaseClient but KStars still trying to access it, or something like that.

rickbassham:
The cause is calling `deleteProperty` when loading the saved config. It seems to cause a race condition somewhere (but I couldn't find it yet). In this case, I don't actually need to call it, so I just removed the offending call.

rickbassham:
@knro  I wonder if `INDIListener` needs some mutexes/locks?

rickbassham:
https://invent.kde.org/education/kstars/-/merge_requests/189

knro:
@rickbassham I'd appreciate any help with INDIListener (and GUIManager as well). Basically, there is a possibility for delProperty to get executed and the property to be deleted but on some occasions, INDIListener tries to access the list of properties and hits a dangling pointer for the property that was just deleted. Normally, the removeProperty would be invoiced and the property is removed from the list of properties managed by INDIListener, but something this does not happen because it's asynchrnous.

Before, the signal/slot connection were set to QueuedBlockingConnection. That means that the signal thread (INDI::BaseClient) would be **blocked** until the main KStars thread finishes process this call. However, while this solves the race condition, it results in a deadlock when disconnecting from INDI server. It happens when like this:

1. Main thread --> BaseClient::disconnectServer --> shuts the baseclient thread and waits for the thread to join.
2. BaseClient thread (the one that connects to INDI server and communicate with it) process message to/from the server and if signated by the main thread to terminate, it would. Normally, it would then delete the properties when that happens. So removeProperty is invoked. Now the main thread receives this and because the policy is QueuedBlockingConnection, it **blocks** the BaseClient thread until the main thread finishes. But wait.. the main thread is actually now waiting in #1 above for the BaseClient thread to join.. hence the deadlock.

I informed @pawel-soja as well about this IIRC so hopefully we can come up with a better long term solution for this issue. There should be a better management for property/device removal that is efficient & safe.

rickbassham:
Do the tests on fedora and opensuse usually fail, or do I have another issue to track down here?

knro:
Usually failures on these two platforms are due to compiling errors, not test failing. Need to check.

libs/indibase/indicontroller.cpp
```diff
@@ -97,6 +97,10 @@ bool Controller::initProperties()
     IUFillSwitchVector(&UseJoystickSP, UseJoystickS, 2, device->getDeviceName(), "USEJOYSTICK", "Joystick", OPTIONS_TAB,
                        IP_RW, ISR_1OFMANY, 0, IPS_IDLE);
 
+    IUFillText(&JoystickDeviceT[0], "SNOOP_JOYSTICK_DEVICE", "Device", "Joystick");
```
rickbassham, line 100: @knro I've commented out everything locally except the additions to `initProperties` and `test_alignment` still fails on Fedora with a segfault. I haven't been able to get gdb to break on the segfault, so figuring this out is turning out to be quite difficult. Any tips or help would be appreciated.

knro: Not sure, nothing from your commit stands out.

knro:
@xsnrg Can you please check why this is failing on Fedora? any hints?

rickbassham:
Going to close this PR and open a new one that is squashed.

PR #1324, opened 2021-01-16
Refactoring - Widget decorators

pawel-soja:
I have decided to split my work as milestones that will be extended.

This pull request includes the creation of decorators for existing IXXXVectorProperty/IXXX structures.

Additionally, I used "#define" for switch-case to avoid confusion when referencing structure fields.

Decorators, hereinafter called views, will enable implementation in C++ in the future.
Decorators to help get rid of direct calls to low-level functions.
Structures from indiapi.h can be projected onto decorators at any time for easy manipulation.
Everything about a given "Widget" in one place.

These classes will be the basis for the further implementation of Qt-Like widgets.


knro:
Thank you @pawel-soja. Wouldn't this break existing clients? Is there a way to make it backward-compatible and perhaps deprecate the functions we want to migrate from?

pawel-soja:
Something wrong with the compatibility of the code above?
My intention was to be backwards compatible.
I was wondering whether to rewrite the whole thing to C++ or make a decorator wrapper first. Unfortunately, the time I can devote is limited and I decided to make a decorator first.

knro:
I stand corrected! it works fine as is with clients, the PropertyView threw me off when I first looked at it! :D 

Great job again!

PR #1325, opened 2021-01-18
Frame Based V4L Stacking and lazy stop capture for better loop perf

pmneo:
By this, i get deterministic stacked results, with the same amount of stacked frames.

### DIFF
--- a/drivers/video/v4l2driver.cpp
+++ b/drivers/video/v4l2driver.cpp
@@ -69,6 +69,7 @@ V4L2_Driver::V4L2_Driver()
     divider = 128.;
 
     is_capturing = false;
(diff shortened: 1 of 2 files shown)


TallFurryMan:
Excellent work 👍 Any details on performance improvement, and on which cameras this was tested?

pmneo:
I have tested with a Aptina 0130 based Webcam. Each 500ms Frame took about 3.5 Seconds. A 1-3 Seconds Stack only captured 1 frame ... 
With my changes it took about 1.2 s for a 500ms Loop (dropping early captured Frames) and a 1s Loop about 1.6s with stabile Stack of 2 Frames.

CS
pmneo

PR #1327, opened 2021-01-19
DefaultDevice - D-Pointer implementation

pawel-soja:
Hi!

A little cleaned up.

- D-Pointer implementation for the DefaultDevice class
- Add defineProperty function instead of defineXXX
- Access to POLLMS via getters/setters.

### DIFF
--- a/drivers/agent/agent_imager.cpp
+++ b/drivers/agent/agent_imager.cpp
@@ -294,9 +294,9 @@ bool Imager::initProperties()
     IUFillBLOB(&FitsB[0], "IMAGE", "Image", "");
     IUFillBLOBVector(&FitsBP, FitsB, 1, getDefaultName(), "IMAGE", "Image Data", DOWNLOAD_TAB, IP_RO, 60, IPS_IDLE);
 
-    defineNumber(&GroupCountNP);
-    defineText(&ControlledDeviceTP);
-    defineText(&ImageNameTP);
+    defineProperty(&GroupCountNP);
+    defineProperty(&ControlledDeviceTP);
+    defineProperty(&ImageNameTP);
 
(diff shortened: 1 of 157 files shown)


pawel-soja:
A bit problematic that DefaultDevice had POLLMS and some classes used their own variable with the same name.
Where POLLMS is obscured, I used the name POLLMS_OVERRIDE.

I have not delved into this mechanism, but I can see that it repeats itself a lot.

knro:
Thanks! Do we need a similar PR for 3rdparty repo? Otherwise, all the drivers would break over there.

knro:
Ok, again I stand corrected. The drivers are working fine in 3rd party repo, just need a similar update to bring all in line.

knro:
all drivers should use setters/getters for POLLMS now

pawel-soja:
Yes, I have added temporary access to keep it compatible.

I will make changes in 3rd if a new version of DefaultDevice is available there. I understand that if I do it now, the tests will crash?

knro:
If you're talking about CI tests, unfortunately, we couldn't figure out how to implement them for 3rd party repo. The idea is that it should be built against the latest libindi for commits pushed to master. This required creating a docker container to fetch the latest nightly build from the PPA. However, it never materialized and not sure if there is a better way to run CIs on the 3rd part repo.

pawel-soja:
I took a look at what is currently being done. It looks pretty decent. Better any tests than none.
Currently it is enough to run circleci for 3rd-indi or did I miss something?
There is a warning on the warning in this project :)

One more thing, I would like to make sure, how is the PPA update currently being uploaded?

knro:
There is no proper Circle CI integration yet for the INDI 3rd party repo :(

The PPA is built automatically nightly from Canonical Launchpad service.

Issue #1329, opened 2021-01-20
celestrongps reports guide rates (to PHD2) at wrong scale


interplanetarychris:
I'm not quite sure where to reference it in the code, but it would appear that the Celestron GPS INDI driver is reporting RA/DEC guide rates at a percentage number from 0-100, when applications like PHD2 are expecting it as a fraction between 0-1. 

While it doesn't prevent guiding from working, PHD2 ends up using much smaller pulses than it needs to, and reports incorrect information about the capability of the mount.

https://github.com/indilib/indi/blob/d6e9d74a82bbf6914709bab7063bb9bd68f45687/drivers/telescope/celestrongps.cpp#L1250

PHD2 developers cite the **[INDI Standard Properties](https://indilib.org/developers/developer-manual/101-standard-properties.html#h3-telescopes)** SLEW_GUIDE member description as their source for the interface scale and units, noting that it should be fixed in the INDI driver to be compliant with this property:

_TELESCOPE_SLEW_RATE -> SLEW_GUIDE -> "0.5x to 1.0x sidereal rate or slowest possible speed."_

knro:
Right, the guide rate scale should be from 0 to 1. Can you submit a PR to fix that?

interplanetarychris:
> Right, the guide rate scale should be from 0 to 1. Can you submit a PR to fix that?

Sorry, I don't follow the code well enough to distinguish what's used internally, what's interfaced to the mount itself, and what's reported externally.

interplanetarychris:
I took my best attempt @knro  - hopefully a code review is quicker than making the change yourself!

Chris-Rowland:
The guide rate as a percentage of sidereal is done to match what Celestron use in their HC.  I think this should remain for the UI at least.

I've had a look in the INDI documentation and can see nothing about how the guide rate should be specified, or even any mention of guide rate.  I could have missed this, where is this defined?

Where are the "interface standards" that PHD mention defined and published?  Is this "standard" just what PHD require?

It is quite likely that this is not implemented according to the "standard" for other drivers as well.

Chris

interplanetarychris:
@Chris-Rowland 
I've asked Bruce with PHD2 to point to their preferred standard.  He had initially pointed to the INDI documentation "SLEW_RATE.  I don't see that the INDI API clearly indicates guide rate as a variable, or its units.

[...]

PR #1331, opened 2021-01-20
Fix celestrongps guide rate scaling

interplanetarychris:
This is my first development work for INDI to address Issue https://github.com/indilib/indi/issues/1329

While I was able to successfully compile locally on two platforms, running from a DEV version of indi alongside the distributed version still eludes me, so I was unable to test locally.

The "IUFillNumber" parameters should be reviewed closely for appropriateness, as I took my best guess there based on the change and the function defintion.

### DIFF
--- a/drivers.xml
+++ b/drivers.xml
@@ -63,27 +63,27 @@
         </device>
         <device label="Celestron GPS" manufacturer="Celestron">
             <driver name="Celestron GPS">indi_celestron_gps</driver>
-            <version>3.3</version>
+            <version>3.5</version>
         </device>
         <device label="Celestron CPC" manufacturer="Celestron">
             <driver name="Celestron GPS">indi_celestron_gps</driver>
-            <version>3.3</version>
+            <version>3.5</version>
         </device>
         <device label="Celestron NexStar" manufacturer="Celestron">
             <driver name="Celestron GPS">indi_celestron_gps</driver>
-            <version>3.3</version>
+            <version>3.5</version>
         </device>
         <device label="Celestron AVX" manufacturer="Celestron">
             <driver name="Celestron GPS">indi_celestron_gps</driver>
-            <version>3.3</version>
+            <version>3.5</version>
         </device>
         <device label="Celestron CGEM" manufacturer="Celestron">
             <driver name="Celestron GPS">indi_celestron_gps</driver>
-            <version>3.3</version>
+            <version>3.5</version>
         </device>
         <device label="Celestron CGX" manufacturer="Celestron">
             <driver name="Celestron GPS">indi_celestron_gps</driver>
-            <version>3.3</version>
+            <version>3.5</version>
         </device>
         <device label="Astro-Electronic FS-2">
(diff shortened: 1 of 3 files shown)


interplanetarychris:
I was able to successfully build and test with my CGX mount, after uninstalling the APT distribution of indi-full. 

The guiding rate can now be appropriately set and reported to PHD2 as a rate between 0-1.   

However, it does not appear to READ the mount guide rates upon initialization/connect.  I have not had an opportunity to slew/track/guide with my mount, so I can not yet confirm if this is due to my mount being in a sufficient state, or a deficiency of the code.  

Local weather should allow me to test this by Friday.

Chris-Rowland:
You can check for read and write of guide rate by doing a quick align to get the mount working.  No need to wait for dark.  You can use the HC to check what is being used and a full driver debug will show the commands going to and from the HC.

I would have retained the 0 to 100 range for the display because this corresponds to the rate reported by the HC. Having a different value may generate support queries.

I would have multiplied the guide rate by 255 rather than 256.  I think I used 256 so that it was possible to get the full 100% setting.

The ability to read and set the guide rate from the mount will depend on the mount and HC versions.  I would expect that a reasonably up to date CGX would be fine.  The rates use the pass through commands to send commands to the motor controllers and aren't the easiest to read but it should be reasonably clear from the code.

Chris

interplanetarychris:
@Chris-Rowland - My scope and HC are locked in a small "moteloscope" which doesn't permit safe movement of the scope.  I tried a "Move to Switch" of RA/DEC, which is triggered by the INDI "Align" command, but that didn't result in the guide rates updating.  Same for enabling/disabling tracking.   When I can open it up in clear weather, I can further verify with alignment, pointing, and HC details, etc.

> I would have retained the 0 to 100 range for the display because this corresponds to the rate reported by the HC. Having a different value may generate support queries.

I'll see about introducing new variables to report different units to the display and external requestors.  I'm still not sure what mechanism is reported to PHD2 -- it appears that it uses the greater of the two values in the `GuideRateNP` vector.

> I would have multiplied the guide rate by 255 rather than 256. I think I used 256 so that it was possible to get the full 100% setting.

This code preceded my small tweaks -- in studying it, using the result of min(value*256,255) seemed appropriate, as a guide value of 0.0x doesn't make any sense.

> The ability to read and set the guide rate from the mount will depend on the mount and HC versions. I would expect that a reasonably up to date CGX would be fine. The rates use the pass through commands to send commands to the motor controllers and aren't the easiest to read but it should be reasonably clear from the code.

The code appears to carefully account for this, and this might be the reason why its not restored upon subroutine initialization -- it waits till its confident that the mount supports guiding until it queries the values.

I was also thinking about handling of appropriate ranges/defaults when the mount doesn't support guiding, or perhaps doesn't support changing the rate, or is asked for something outside its configurable range by INDI.  I think the CGX just supports between 0.5-1.0x sidereal.

Still trying to figure out how to turn on full driver/debug in my INDI/EKOS/Kstars setup.

PR #1335, opened 2021-01-22
fix: add null pointer check to serial connect

rickbassham:
Fix a SEGFAULT when there are no system ports defined.


rickbassham:
I was getting seg faults when there are no system ports defined but connecting with simulation enabled on a Mac.

PR #1341, opened 2021-02-01
Driver for Digital Dome Works (DDW) from Technical Innovations

jpaana:
Fairly simple dome driver as the controller handles most things by itself. The only slightly complicated aspect was that the controller may respond to hand controller commands at any point, but that was pretty straight forward to poll. 

Missing functionality in this first version:
- initiating dome calibration
- using direct LX200 protocol mount interception / slaving the controller supports (may not be needed nor wanted versus regular INDI dome slaving)
- relative movements and CW/CCW hand controller emulation
- shutter closing/dome moving order when parking is probably not optimal if parking position is not the same as home position where electrical connections are (shutter can only be operated there)

### DIFF
--- a/drivers/dome/ddw_dome.cpp
+++ b/drivers/dome/ddw_dome.cpp
@@ -92,11 +92,6 @@ void ISSnoopDevice(XMLEle *root)
 DDW::DDW()
 {
     setVersion(1, 0);
-    m_ShutterState     = SHUTTER_UNKNOWN;
-
-    status        = DOME_UNKNOWN;
-    targetShutter = SHUTTER_CLOSE;
-
     SetDomeCapability(DOME_CAN_ABORT | DOME_CAN_ABS_MOVE | DOME_CAN_PARK | DOME_HAS_SHUTTER);
 }
 
@@ -105,8 +100,8 @@ bool DDW::initProperties()
     INDI::Dome::initProperties();
 
     IUFillNumber(&FirmwareVersionN[0], "VERSION", "Version", "%2.0f", 0.0, 99.0, 1.0, 0.0);
-    IUFillNumberVector(&FirmwareVersionNP, FirmwareVersionN, 2, getDeviceName(), "FIRMWARE",
-                       "Firmware", INFO_TAB, IP_RO, 60, IPS_IDLE);
+    IUFillNumberVector(&FirmwareVersionNP, FirmwareVersionN, 2, getDeviceName(), "FIRMWARE", "Firmware", INFO_TAB,
+                       IP_RO, 60, IPS_IDLE);
 
     SetParkDataType(PARK_AZ);
 
@@ -163,8 +158,10 @@ bool DDW::Handshake()
     LOGF_DEBUG("Initial response: %s", response.c_str());
 
     // Response should start with V
-    if(response[0] != 'V')
+    if (response[0] != 'V')
         return false;
+
+    parseGINF(response.c_str());
     return true;
 }
 
@@ -185,7 +182,7 @@ bool DDW::updateProperties()
 
     if (isConnected())
     {
-        defineNumber(&FirmwareVersionNP);
+        defineProperty(&FirmwareVersionNP);
         SetupParms();
     }
     else
@@ -209,7 +206,7 @@ bool DDW::ISNewNumber(const char *dev, const char *name, double values[], char *
     return INDI::Dome::ISNewNumber(dev, name, values, names, n);
 }
 
-int DDW::writeCmd(const char* cmd)
+int DDW::writeCmd(const char *cmd)
 {
     int nbytes_written = 0, rc = -1;
     char errstr[MAXRBUF];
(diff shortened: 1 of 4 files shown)


knro:
Thanks, looking good. Any documentation for INDI Website to accompany this driver?

jpaana:
Not yet, I can't get screenshots from actual hardware until I get to our club observatory, but there isn't any custom controls, just the ones generated by the dome base class. But I'll try to put something together.

knro:
Ok, let's get some skeleton at least working so it shows up in the devices list in the INDI website dome section.

drivers/dome/ddw_dome.cpp
```diff
@@ -0,0 +1,560 @@
...
+    int rc = writeCmd("GINF");
+    if (rc != TTY_OK)
+        return false;
+
+    std::string response;
+    rc = readStatus(response);
+    if (rc != TTY_OK)
+        return false;
+
+    LOGF_DEBUG("Initial response: %s", response.c_str());
```
knro, line 158: Have you considered using [sendCommand](https://github.com/indilib/indi/blob/fc26c0129693a0c9390f55d7fe71ffbb0ad3176f/drivers/dome/domepro2.h#L126
) function? It is generic enough to cover for most use cases.

drivers/dome/ddw_dome.cpp
```diff
@@ -0,0 +1,560 @@
...
+    int windDir;
+    int windSpd;
+    int temperature;
+    int humidity;
+    int wetness;
+    int snow;
+    int windPeak;
+    int scopeAzimuth;
+    int deadzone;
+    int offset;
```
knro, line 294: best to initialize to some value? 0?

drivers/dome/ddw_dome.cpp
```diff
@@ -0,0 +1,560 @@
...
+    }
+    return IPS_OK;
+}
+
+/************************************************************************************
+ *
+* ***********************************************************************************/
+IPState DDW::ControlShutter(ShutterOperation operation)
+{
+    LOGF_INFO("Control shutter %d", (int)operation);
```
knro, line 502: LOGF_DEBUG?

knro (commented):
Minor comments!

drivers/dome/ddw_dome.cpp
```diff
@@ -0,0 +1,560 @@
...
+* ***********************************************************************************/
+bool DDW::ISNewSwitch(const char *dev, const char *name, ISState *states, char *names[], int n)
+{
+    return INDI::Dome::ISNewSwitch(dev, name, states, names, n);
+}
+
+bool DDW::ISNewNumber(const char *dev, const char *name, double values[], char *names[], int n)
+{
+    return INDI::Dome::ISNewNumber(dev, name, values, names, n);
+}
```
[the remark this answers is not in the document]
jpaana: Yes, I took the base code from my ScopeDome driver, which does use these, so there is some leftovers

drivers/dome/ddw_dome.cpp
```diff
@@ -0,0 +1,560 @@
...
+    return true;
+}
+/************************************************************************************
+ *
+* ***********************************************************************************/
+
+bool DDW::SetDefaultPark()
+{
+    // By default set position to 90
+    SetAxis1Park(90);
```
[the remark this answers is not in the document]
jpaana: I think I'll make home position the default, this was again left over from some driver this code was originally from.

drivers/dome/ddw_dome.cpp
```diff
@@ -0,0 +1,560 @@
...
+    IUFillNumber(&FirmwareVersionN[0], "VERSION", "Version", "%2.0f", 0.0, 99.0, 1.0, 0.0);
+    IUFillNumberVector(&FirmwareVersionNP, FirmwareVersionN, 2, getDeviceName(), "FIRMWARE", "Firmware", INFO_TAB,
+                       IP_RO, 60, IPS_IDLE);
+
+    SetParkDataType(PARK_AZ);
+
+    addAuxControls();
+
+    // Set serial parameters
+    serialConnection->setDefaultBaudRate(Connection::Serial::B_9600);
```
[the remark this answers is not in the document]
jpaana: I prefer explicit baud rate setting over default that is documented somewhere else for readability though as this device does not support anything else. Might even be good if there was a way to remove the setting all together.

drivers/dome/ddw_dome.cpp
jpaana: It might work for this handshake case, normally the device doesn't immediately respond at all, but might also send random responses to hand controller events so they must be polled separately instead of simple command/response kind of protocol. Again this was partially leftover from ScopeDome driver, which uses a binary protocol with checksums etc.

drivers/dome/ddw_dome.cpp
jpaana: Style-wise perhaps, but the values are not used uninitialized as the return value of sscanf is checked so that all values are initialized.

drivers/dome/ddw_dome.cpp
jpaana: Yes

drivers/dome/ddw_dome.cpp
```diff
@@ -0,0 +1,560 @@
...
+#include <termios.h>
+#include <wordexp.h>
+#include <unistd.h>
+
+#define DDW_TIMEOUT 2
+
+// We declare an auto pointer to DDW.
+std::unique_ptr<DDW> ddw(new DDW());
+
+void ISPoll(void *p);
```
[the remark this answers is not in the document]
jpaana: The pointer or ISPoll, github is sometimes a bit vague which line the comment is about...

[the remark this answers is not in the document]
knro: sorry ISPoll is not required.

drivers/dome/ddw_dome.cpp
knro: Ok I see, alright let's stick with what works now.

jpaana:
Updates pushed

PR #1349, opened 2021-02-05
DDW shutter state fix

jpaana:
Avoid spamming shutter state updates if the state hasn't changed.

At first looked like there was some issue with the shutter status value, but it seems to have been just a testing issue that the user didn't open the shutter fully so it reported as indeterminate.


knro:
Ready to land?

jpaana:
Yes, I think this shutter issue is now solved, still figuring out why slaving doesn't move the dome without abort command in between the moves.

PR #1355, opened 2021-02-13
Added labels for USB and Dew

rishi-garrod:
Added labels for USB and Dew. Also removed the ability to switch off the led for V2 boxes. Switching off the Led causes the stepper motor to be put to sleep.


drivers/auxiliary/pegasus_upb.cpp
```diff
@@ -135,6 +136,37 @@ bool PegasusUPB::initProperties()
...
+    IUFillText(&DewControlsLabelsT[0], "DEW_LABEL_1", "Dew A", "Dew A");
+    IUFillText(&DewControlsLabelsT[1], "DEW_LABEL_2", "Dew B", "Dew B");
+    IUFillText(&DewControlsLabelsT[2], "DEW_LABEL_3", "Dew C", "Dew C");
+    IUFillTextVector(&DewControlsLabelsTP, DewControlsLabelsT, 3, getDeviceName(), "DEW_CONTROL_LABEL", "Dew Labels",
+                     DEW_TAB, IP_WO, 60, IPS_IDLE);
+
+    char dewLabel[MAXINDILABEL];
+
+    // Turn on/off power and power boot up
+    memset(dewLabel, 0, MAXINDILABEL);
```
knro, line 149: No need for memset. Just use

char dewLabel[MAXINDILABEL] = {0};

knro (approved):

Issue #1317, opened 2021-01-10
Flux and focus management of the CCD Simulator


TallFurryMan:
The magnitude and focus management of `indi_ccd_simulator` appear to be handled incorrectly.

I configured the Simulator with saturation magnitude 1.0, limiting magnitude 17.0 and max ADU value of 65000, and pointed to Vega, which is magnitude ~1.0~ 0.0, thus must saturate the sensor in one second (EDIT: Vega magnitude is 0.0, thus my experiment is wrong).

This is the frame obtained with 1s exposure at near-optimal focus position 33500. The ADU for the center of Vega, on the left of the frame, is around 13300:

![202101_FocusFluxNotFixed](https://user-images.githubusercontent.com/22483335/104130216-d2520d00-536f-11eb-9e9b-33e45e8edd3a.jpg)

This is the frame obtained with 1s exposure at defocused position 80000. The ADU for the center of Vega, on the left of the frame, is around 14800:

![202101_DefocusFluxNotFixed](https://user-images.githubusercontent.com/22483335/104130145-7ab3a180-536f-11eb-97e0-416f18ab8f76.jpg)

So there seem to be two problems:
- The flux calculated for Vega is far too small. One second on Vega should convert to nearly 65K ADUs, maybe smoothed considering the gaussian point-spread function used in the Simulator.
- The focus management could be correct in the way it spreads illumination, but the center value remains the same as the in-focus representation, which is certainly incorrect.

TallFurryMan:
I recalculated the linear function mapping the flux to ADU values based on the saturation and limiting magnitudes, and I think the problem is there.

The part stretching the gaussian also has a problem as it retains the same value whatever the focus position, which drastically increases the flux when out-of-focus, and results in strong blobs.

This affects all Ekos Focus UI tests, which become quite dependent on noise because of the really small flux values and do not detect stars anymore when near the optimal focus plane.

TallFurryMan:
I consider this fixed by #1356. Reopen if needed.

TallFurryMan:
While the situation clearly improved, I tend to think we should increase the default gain to reflect sensors currently on the market.

PR #1356, opened 2021-02-15
Updated flux calculation for CCD Simulator

TallFurryMan:
This PR updates the flux calculation in the CCD Simulator and refactors the flux calculation into a separate function, where the `k`, affine scale, and `z`, affine offset, are calculated from the limiting and saturation magnitudes so that:
- the limiting magnitude provides zero ADU whatever the exposure
- the saturation magnitude provides max ADU in one second

This PR updates the rendering of stars so that the flux is rendered as a gaussian distribution of unitary integral `1/(sqrt(2pi)*sigma) * exp( -x² / (2*sigma²) )`.

This PR was tested with a rendering of Vega and stars of various magnitudes, and shows a clear improvement compared to the previous calculation. Specifically, the previous calculation produced star centers with the same flux value whatever the seeing/focus and did not account for the dispersion in the gaussian rendering. This PR also makes the Ekos Focus tests more stable and reliable.


TallFurryMan:
However, in this PR:
- I did not optimise the flux function.
- The gaussian I used for rendering probably has a unitary integral in 1-D, but maybe not in 2-D.
- Test are not passing, they need to be reworked.

This is in the context of #1317.

knro:
Thanks! Any ideas about the failed CI?

TallFurryMan:
Yeah in the test implementation, one star is rendered with controlled parameters, and the result is checked against expected values. Because I changed the flux calculation and the gaussian function used for rendering, I need to modify the assertion values.

TallFurryMan:
I realise the vignetting management also has an issue: subframes get their own vignetting effect, instead of using the effect from the full frame...

image

I'll raise another issue and tackle this in another PR.

knro:
Any update?

TallFurryMan:
This PR is ready as is. I didn't add as many tests as I had planned, but the existing ones cover the star rendering changes.

PR #1361, opened 2021-03-01
PropertyView - client/server side + refactor

pawel-soja:
Hi,

Another portion of new methods to improve access to low-level functions.
Extracting the apply/define implementation for client/driver side.
Compiled and tested with the pyindi-client project.

@geehalel I'll give you a suggestion to simplify indiclientpython.i.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,6 +176,7 @@ SET(indiclient_CXX_SRC
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/basedevice.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/baseclient.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indiproperty.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indipropertyview_client.cpp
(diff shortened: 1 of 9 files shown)


knro:
Lovely changes! This is going to get a while getting used to, and we need to update the [documentation](https://docs.indilib.org) as well. All is currently backward compatible with INDI clients?

pawel-soja:
All changes are backwards compatible.
I would not like the PropertyView/WidgetView classes to be documented as base classes for writing drivers.

The purpose of these classes is to introduce methods for managing widgets.
Ultimately, I want to create **simple classes** such as **PropertySwitch, PropertyText, PropertyNumber, PropertyLight, PropertyBlob**. These classes will already hold an array of **WidgetView and PropertyView**.

My plan:
1. Implementation of PropertyView/WidgetView decorators (collection of used low-level functions)
2. implementation of widgets (eg PropertyText...) which have methods from PropertyView/WidgetView and additionally inherit from INDI::Property.
3. implementation of signals / slots for PropertyText...

Step 3 will get rid of ISGetProperties/ISNewSwitch/ISNewText completely, reducing the "A Simple Driver" example to a few lines.

So for new classes, the documentation will need to be updated. I am happy to correct it and create new examples. Of course, the old way of writing will still be possible.

knro:
Thank you for the detailed response Pawel. I supported your plan all the way. It is time for an INDI refresh. This is the first major refactor in over a decade. Great work!

PR #1367, opened 2021-03-06
Fix warnings, use strcpy/strcat for string literal

pawel-soja:
Hi!

`char * strncat ( char * destination, const char * source, size_t num );`
_num - Maximum number of characters to be appended. size_t is an unsigned integral type._

There is no point in using this function for a known constant string.
It is overused in some places and the target buffer length is not checked.

I added an extra condition when removing the last character so there is no leakage for the empty string.

It is possible that when I replace with new widgets (Property), I will improve part of the implementation.


knro:
I think part of it is the usual "don't use strcpy it's not safe". I know this doesn't apply to string literals but the warnings are all over the place nonetheless.

PR #1371, opened 2021-03-06
Update to main branch

seventi71:


knro:
Thank you! Any iOptron mounts covered by this driver that do NOT support PEC?

seventi71:
Jasem,

Yes, the EC (encoder) mounts, do not use PEC.
The original iEQ driver had a function to filter mounts based on what they supported and kept it in a table.

However the V3 rewrite does not. 

i can add logic that blocks PEC functionality as part of initialise, as the driver already puts a log entry in when it detects PEC in the mount.

i will make the updates to the code for this.

longer term, as a feature update, we should look at integrating the data record functionality with the Guiding algorythm. 

currently, the driver is only set to record if you are tracking, however i would like to update this to say, only record if you are guiding using specific parameters, so as to get a clean recording.

christopher

knro:
Thanks, please remove gps_simulator changes from this PR.

seventi71:
Hi Apologies for the delay on this. 

I am having issues on my mount, where the Kstars and Mount communication goes strange and inverts or goes around in circles.
This has become my pressing issue to solve, as i cannot image at the moment.

This is why i updated the simulated GPS, as it is a path of investigation into why this is happening.

Once i have solved for this, i can make updates to the driver to include only CEM/GEM mounts for PEC.

Regards
Chris.

Issue #1364, opened 2021-03-03
Dome slaving moves back and forth just before meridian flip 


fcasarramona:
Extracted from Indi forum Topic
https://indilib.org/forum/domes/8909-dome-slaving-just-before-meridian-flip-with-lx200-10micron-mount.html

When the mount has hour angles just before the meridian flip the dome starts going back and forth between the currently required position and, evidently, the position it should be in after the meridian flip.

These 30 odd degrees roughly is the position where the dome should be after the meridian flip (we confirmed that by performing it) and these 335 odd degrees is where it is supposed to be before the flip (when the log was taken).

Since this is happening with the dome simulator driver I think I can rule out my own driver as the problem, the only difference is that it is actually in reality starting to go crazy.

**Desktop (please complete the following information):**
The version we used are, from ppa:mutlaqja/ppa:
INDI Library: 1.8.8
Code 1.8.8-tgz. Protocol 1.7.
kstars 3.5.1 Stable
Raspberry PI

**Log Files**
INFO    902.046243 sec  : Dome reached requested azimuth angle.
DEBUG   902.539657 sec  : JD: 2.45927e+06 - MSD: 10.394
DEBUG   902.539715 sec  : MC.x: 0 - MC.y: 0 MC.z: 0.45
DEBUG   902.539738 sec  : HA: -10.8837  Lng: 10.8328 RA: 329.999
DEBUG   902.539751 sec  : Active Snoop, driver: LX200 10micron, property: TELESCOPE_PIER_SIDE
**DEBUG   902.539768 sec  : OTA_SIDE: 1**
DEBUG   902.539782 sec  : OTA_OFFSET: 1  Lat: 49.8328
DEBUG   902.539795 sec  : OC.x: 0.957603 - OC.y: -0.22015 OC.z: 0.264175
DEBUG   902.539812 sec  : Mount Az: 2.76338  Alt: 43.0728
DEBUG   902.539826 sec  : OV.x: 0.0352179 - OV.y: 0.729638 OV.z: 0.682927
DEBUG   902.539845 sec  : Calculated target azimuth is 29.89. MinAz: 4.78, MaxAz: 54.99
DEBUG   903.696934 sec  : JD: 2.45927e+06 - MSD: 10.3943
DEBUG   903.696951 sec  : MC.x: 0 - MC.y: 0 MC.z: 0.45
DEBUG   903.696956 sec  : HA: -10.8834  Lng: 10.8328 RA: 329.999
[...]

PR #1373, opened 2021-03-10
Improving Dome slave to telescope regarding telescope pier side

fcasarramona:
Looking for a solution of this bug (https://github.com/indilib/indi/issues/1364) I improved how Dome drivers deal with the side of the pier where the telescope is in GEM mounts.

Before this change, user can inform to Dome driver which side was the telescope, but if the mount has this information, then it takes from the mount. However, in some circuntances (when telescope was moving) it's not clear what side must be taken.

To make the dome driver behaviour more predictable and clear, now the user can choose from 5 diferent options:

**East** : Dome always positions as if telescope tube was at the east of the pier
**West**:  Dome always positions as if telescope tube was at the west of the pier
**Mount**: Dome positions as if the telescope tube was at the side the mount reports. If the mount didn't report the side, the dome behave as with a fork mount.
**Hour Angle**: Dome positions as if the telescope tube was at the side deducted by the hour angle of the target coordinates. For positive hour angle (telescope point to west of the meridian) the tube was in the east side.
**Ignore**: Dome positios as with a fork mount.

A final note: When configured as **Mount** and the telescope slews from one side  to the other side of the pier, the dome positions to the "wrong" side when the telescope begins moving, and then repositions to the correct side when the telescope end slewing. I have not found any solution for this  behaviour, so I left for a future change if a solution is found.

### DIFF
--- a/libs/indibase/indidome.cpp
+++ b/libs/indibase/indidome.cpp
@@ -1019,9 +1019,9 @@ bool Dome::ISSnoopDevice(XMLEle * root)
             const char * elemName = findXMLAttValu(ep, "name");
 
             if (!strcmp(elemName, "PIER_EAST") && !strcmp(pcdataXMLEle(ep), "On"))
-                mountOTASide = 1;
-            else if (!strcmp(elemName, "PIER_WEST") && !strcmp(pcdataXMLEle(ep), "On"))
                 mountOTASide = -1;
+            else if (!strcmp(elemName, "PIER_WEST") && !strcmp(pcdataXMLEle(ep), "On"))
+                mountOTASide = 1;
         }
         
         return true;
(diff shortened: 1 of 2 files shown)


knro:
@jpaana Can you please give some feedback? I think Peter Kennet can perhaps help in the testing?

fcasarramona:
@knro @jpaana had posted a comment about it in the bug report:
https://github.com/indilib/indi/issues/1364#issuecomment-797008202

jpaana:
Yes, in my testing it seems the pier side is consistently flipped in this change.

knro:
Right how about the "few issues" raised here: https://github.com/indilib/indi/issues/1364#issuecomment-797008202 ?

Are these resolved as well?

fcasarramona:
Corrections of errors reported by @jpaana in https://github.com/indilib/indi/issues/1364#issuecomment-797008202

jpaana:
Thanks, tested this new version and everything seems to be in order now. Pier side works as expected and setting is also saved, thanks!

PR #1375, opened 2021-03-10
Add global ISGetProperties/ISNewXXX/ISSnoopDevice propagation functions for drivers

pawel-soja:
Hi!

Currently, each driver must support the functions:
```c++
void ISGetProperties(const char *dev)
{
    mydriver->ISGetProperties(dev);
}

void ISNewSwitch(const char *dev, const char *name, ISState *states, char *names[], int n)
{
    mydriver->ISNewSwitch(dev, name, states, names, n);
}

void ISNewText(const char *dev, const char *name, char *texts[], char *names[], int n)
{
    mydriver->ISNewText(dev, name, texts, names, n);
}

void ISNewNumber(const char *dev, const char *name, double values[], char *names[], int n)
{
    mydriver->ISNewNumber(dev, name, values, names, n);
}

void ISNewBLOB(const char *dev, const char *name, int sizes[], int blobsizes[], char *blobs[],
               char *formats[], char *names[], int n)
{
    mydriver->ISNewBLOB(dev, name, sizes, blobsizes, blobs, formats, names, n);
}

void ISSnoopDevice(XMLEle *root)
{
    mydriver->ISSnoopDevice(root);
}
```
It always comes down to the same thing.
My suggestion is to do this on the indi library side.

The proposed PR is backward compatible. If the driver has already implemented these functions, the functions in indilib are not used (not linked).

When removing a function from drivers, you need to make sure that the methods in the class have an override specifier - it's not everywhere.
In the next PR, I will fix the override specifier for all drivers.

Edit: ahh yes, not all drivers inherit from DefaultDevice ... ;(

### DIFF
--- a/libs/indibase/defaultdevice.cpp
+++ b/libs/indibase/defaultdevice.cpp
@@ -20,6 +20,8 @@
 #include "defaultdevice_p.h"
 
 #include "indicom.h"
+#include "indiapi.h"
+
 #include "indistandardproperty.h"
 #include "connectionplugins/connectionserial.h"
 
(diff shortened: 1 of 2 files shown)


knro:
Thanks! That will remove a useless boilerplate code and make drivers cleaner... I don't see how this is backward compatible? I guess my C++ getting rusty.

pawel-soja:
Hi Jasem!

C/C++ doesn't matter here. Mainly it is about the compilation and linker process.

When compiling the driver, the indidriver.c file is included. This source calls the functions:
- ISGetProperties
- ISNewSwitch
- ISNewText
- ISNewNumber
- ISNewBLOB
- ISSnoopDevice

So when writing the driver, you need to implement them, otherwise there will be a compilation error, and more precisely with linking.
Linking priority is performed for a given project, if the functions do not exist, they are searched for in shared libraries (included in the project)

**So there is no conflict between the proposed PR and the already implemented functions in the drivers.**
Now, it is possible to write new drivers without implementing the mentioned functions.

knro:
Thanks got it, it's all about linking.

PR #1379, opened 2021-03-11
Bugfix - test_ccd_simulator with incorrect library, missing override

pawel-soja:
Hi!

**Add missing overide specifier** - no comment.

**Remove indiclient from ccd driver test** - The library should not be included with driver side programs.
Causes an incorrect operation ("fill method available only on driver side") when using INDI::PropertyView<> because the priority is the source indipropertyview_client.cpp.

### DIFF
--- a/drivers/focuser/onfocus.h
+++ b/drivers/focuser/onfocus.h
@@ -32,8 +32,8 @@ public:
     const char * getDefaultName();
     virtual bool initProperties() override;
     virtual bool updateProperties() override;
-    virtual bool ISNewNumber (const char *dev, const char *name, double values[], char *names[], int n);
-    virtual bool ISNewSwitch (const char *dev, const char *name, ISState *states, char *names[], int n);
+    virtual bool ISNewNumber (const char *dev, const char *name, double values[], char *names[], int n) override;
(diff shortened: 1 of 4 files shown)


pawel-soja:
Dear @knro

I care about 'remove indiclient from ccd driver test' very much. I do not know why the tests did not start.
It closes the step, matching the library to the tests of Massive Refactorization (INDI::PropertyXXX).

Thanks!

knro:
Not sure it was added there for the CCD test. However, there are some drivers that do indeed use INDI client. For example, the INDI Sky Safari bridge driver uses that.

pawel-soja:
Do you mean this fragment?
https://github.com/indilib/indi/blob/01cc8da19011d4a01a6666dcbd066d774e1d01f0/CMakeLists.txt#L1472-L1487

```CMake
target_link_libraries(indi_skysafari indidriver indiclient) 
```
The first one is attached indidriver and then indiclient. Here is all right. I do not see a problem here. In the future I will do that the order does not matter.

This pull request refers to 'test_ccd_sulator', I do not see any dependencies with other files.
https://github.com/indilib/indi/blob/01cc8da19011d4a01a6666dcbd066d774e1d01f0/test/drivers/CMakeLists.txt#L1-L16
I can swap Indiclient and Indidriver, but I do not see the need to attach indiclient.

PR #1383, opened 2021-03-15
StreamManager - Bugfix - D-Pointer implementation

pawel-soja:
Hi!

Unfortunately, in https://github.com/indilib/indi/pull/1372 I made a mistake removing the reference. Image enlargement in the preview does not work.
I noticed this in testing after implementing D-Pointer.

Changes:
- Reference bugfix - new implementation
- D-Pointer
- Use INDI::PropertyXXX - it's safe, because old IXXX / IXXXPropertyView were in the private section
- More readable public header streammanager.h
- Add missing const

Dependent libraries recompilation required - last time ;)


knro:
Thanks! I will give this a go with Canon 600D as well to verify all is well.

PR #1387, opened 2021-03-17
Implement real periodic timer.

pawel-soja:
Hi!

A lot of drivers uses periodic timers. Currently, "addTimer" must be called again.
This causes an imprecise frequency of callback calls.

```
2021-03-17T12:44:25: Driver timer_old_way: 1000.049 ms
2021-03-17T12:44:26: Driver timer_old_way: 2001.132 ms
2021-03-17T12:44:27: Driver timer_old_way: 3002.262 ms
2021-03-17T12:44:28: Driver timer_old_way: 4002.311 ms
2021-03-17T12:44:29: Driver timer_old_way: 5003.407 ms
2021-03-17T12:44:30: Driver timer_old_way: 6004.504 ms
2021-03-17T12:44:31: Driver timer_old_way: 7005.568 ms
```
I implemented the 'addPeriodicTimer' function.
```c
/** Register a new periodic timer function, \e fp, to be called with \e ud as argument after \e ms.
*
* \param ms timer period in milliseconds.
* \param fp a pointer to the callback function.
* \param ud a pointer to be passed to the callback function when called.
* \return a unique id for use with rmTimer().
*/
extern int addPeriodicTimer(int ms, TCF *fp, void *ud);
```
The added Timer stores the 'Interval' value and is not removed from the list after calling callback.
```c
static void checkTimer()
{
    TF *node = timefunc->next;

    if (node == NULL || remainingTimerNode(node) > 0)
        return;

    (*node->fp)(node->ud);

    node = dettachTimer(node);

    if (node == NULL)
        return;

    if (node->interval > 0)
    {
        node->tgo += node->interval;
        insertTimer(node);
    } else {
        free(node);
    }
}
```
Now, Callback is called predetermined times.
```
2021-03-17T12:45:36: Driver timer_new_way: 1000.057 ms
2021-03-17T12:45:37: Driver timer_new_way: 2000.061 ms
2021-03-17T12:45:38: Driver timer_new_way: 3000.077 ms
2021-03-17T12:45:39: Driver timer_new_way: 4000.081 ms
2021-03-17T12:45:40: Driver timer_new_way: 5000.075 ms
2021-03-17T12:45:41: Driver timer_new_way: 6000.084 ms
2021-03-17T12:45:42: Driver timer_new_way: 7000.083 ms
```

Source of example:
```cpp
#include "inditimer.h"
#include "indielapsedtimer.h"

static class Main
{
    INDI::Timer timer;
    INDI::ElapsedTimer elapsedTimer;

public:
    Main()
    {
        timer.callOnTimeout([&](){
            fprintf(stderr, "%.3f ms\n", elapsedTimer.nsecsElapsed() / 1000000.0);
        });
        
        timer.start(1000);
        elapsedTimer.start();
    }
} main_example;
```

I also added auxiliary features to check the remaining time to call callback.
```c

/** Returns the timer's remaining value in milliseconds left until the timeout.
 *
 * \param tid the timer callback ID returned from addTimer() or addPeriodicTimer()
 * \return  If the timer not exists, the returned value will be -1.
 */
extern int remainingTimer(int tid);

/** Returns the timer's remaining value in nanoseconds left until the timeout.
 *
 * \param tid the timer callback ID returned from addTimer() or addPeriodicTimer()
 * \return  If the timer not exists, the returned value will be -1.
 */
extern int nsecRemainingTimer(int tid);
```

High-Level Timers have also been updated and use new changes in eventloop.c.
```cpp
void TimerPrivate::start()
{
    if (singleShot)
    {
        timerId = addTimer(interval, [](void *arg){
            TimerPrivate *d = static_cast<TimerPrivate*>(arg);
            d->p->timeout();
            d->timerId = -1;
        }, this);
    } else {
        timerId = addPeriodicTimer(interval, [](void *arg){
            TimerPrivate *d = static_cast<TimerPrivate*>(arg);
            d->p->timeout();
        }, this);
    }
}
```

### DIFF
--- a/eventloop.c
+++ b/eventloop.c
@@ -270,6 +270,16 @@ static TF *dettachTimer(TF *node)
     return NULL;
 }
 
+/* find the timer by id */
+static TF *findTimer(int timer_id)
+{
+    TF *it = timefunc->next;
+    for(; it != NULL; it = it->next)
+        if (it->tid == timer_id)
+            return it;
+    return NULL;
+}
+
 /* remove the timer with the given id, as returned from addTimer().
  * silently ignore if id not found.
  */
@@ -285,6 +295,23 @@ void rmTimer(int timer_id)
             break;
         }
     }
+}
+
+/* Returns the timer's remaining value in milliseconds left until the timeout. */
+static double remainingTimerNode(TF *node)
+{
+    struct timeval now;
+    gettimeofday(&now, NULL);
+    return (node->tgo - EPOCHDT(&now));
+}
+
+/* Returns the timer's remaining value in milliseconds left until the timeout.
+ * If the timer not exists, the returned value will be -1.
+ */
+int remainingTimer(int timer_id)
+{
+    TF *it = findTimer(timer_id);
+    return it == NULL ? -1 : remainingTimerNode(it);
 }
 
 /* add a new work procedure, fp, to be called with ud when nothing else to do.
@@ -379,33 +406,24 @@ static void callCallback(fd_set *rfdp)
  */
 static void checkTimer()
 {
-    struct timeval now;
-    double tgonow;
     TF *node = timefunc->next;
 
-    /* skip if list is empty */
+    if (node == NULL || remainingTimerNode(node) > 0)
+        return;
+
+    (*node->fp)(node->ud);
+
+    node = dettachTimer(node);
+
     if (node == NULL)
         return;
 
-    gettimeofday(&now, NULL);
-    tgonow = EPOCHDT(&now);
-
-    if (node->tgo <= tgonow)
+    if (node->interval > 0)
     {
-        (*node->fp)(node->ud);
-
-        node = dettachTimer(node);
-
-        if (node == NULL)
-            return;
-
-        if (node->interval > 0)
-        {
-            node->tgo += node->interval;
-            insertTimer(node);
-        } else {
-            free(node);
-        }
+        node->tgo += node->interval;
+        insertTimer(node);
+    } else {
+        free(node);
     }
 }
 
@@ -447,10 +465,7 @@ static void oneLoop()
     }
     else if (timefunc->next != NULL)
     {
-        struct timeval now;
-        double late;
-        gettimeofday(&now, NULL);
-        late = timefunc->next->tgo - EPOCHDT(&now); /* ms late */
+        double late = remainingTimerNode(timefunc->next); /* ms late */
         if (late < 0)
             late = 0;
         late /= 1000.0; /* secs late */
--- a/eventloop.c
+++ b/eventloop.c
@@ -314,6 +314,15 @@ int remainingTimer(int timer_id)
     return it == NULL ? -1 : remainingTimerNode(it);
 }
 
+/* Returns the timer's remaining value in nanoseconds left until the timeout.
+ * If the timer not exists, the returned value will be -1.
+ */
+int64_t nsecsRemainingTimer(int timer_id)
+{
+    TF *it = findTimer(timer_id);
+    return it == NULL ? -1 : remainingTimerNode(it) * 1000000;
+}
+
 /* add a new work procedure, fp, to be called with ud when nothing else to do.
  * return unique id for use with rmWorkProc().
  */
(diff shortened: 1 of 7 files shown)


knro:
Bravo! I agree, the old way was not precise. This is why you see in some drivers there try to use other methods to measure the actual elapsed time. However, I'm not sure if this new change would have some side effects. Take for example TimerHit. Let's say for a driver this is set at 1000ms or 1 second. It gets called, and the function itself takes about 300ms to execute. Then after 1000ms, it gets called again. However, with your change, it would be called in 700ms, correct? 

That is, the *actual* frequency is 700ms instead of 1000ms.

TallFurryMan:
You need to protect that data structure pointer from being dereferenced in a disconnected signal I think. Also, periodic timers require all timer functions to enforce their deadlines. You can help devs by measuring execution time and issuing a warning if that is exceeded (which also becomes part of the excess). That's obviously mitigated when you require a timer func to retrig itself.

pawel-soja:
> However, I'm not sure if this new change would have some side effects. Take for example TimerHit. Let's say for a driver this is set at 1000ms or 1 second. It gets called, and the function itself takes about 300ms to execute. Then after 1000ms, it gets called again. However, with your change, it would be called in 700ms, correct?

Yes, is correct.
The changes don't affect the current driver implementation - they don't use addPeriodTimer - just for clarity.

> That is, the _actual_ frequency is 700ms instead of 1000ms.

No, it isn't correct.
It is true that 700ms will elapse from the end of the function handling to the next call. So the function calls will be executed every 1000ms - still at 1Hz.

```cpp
timer.callOnTimeout([&](){
    fprintf(stderr, "%.3f ms\n", elapsedTimer.nsecsElapsed() / 1000000.0);
    usleep(100 * 1000);
});

timer.start(1000);
elapsedTimer.start();
```
Output:
```
2021-03-18T10:17:04: Driver const_interval: 1000.984 ms
2021-03-18T10:17:05: Driver const_interval: 2000.986 ms
2021-03-18T10:17:06: Driver const_interval: 3000.996 ms
2021-03-18T10:17:07: Driver const_interval: 4000.616 ms

```

If you want to add function duration - variadic frequency of calls, you can always use singleShot.
```cpp
timer.callOnTimeout([&](){
    fprintf(stderr, "%.3f ms\n", elapsedTimer.nsecsElapsed() / 1000000.0);
    usleep(100 * 1000);
    timer.start(); // added
});

timer.setSingleShot(true); // added
timer.start(1000);
elapsedTimer.start();
```

Output:

```
2021-03-18T10:42:40: Driver variadic_frequency: 1000.406 ms
2021-03-18T10:42:42: Driver variadic_frequency: 2101.599 ms
2021-03-18T10:42:43: Driver variadic_frequency: 3202.824 ms
2021-03-18T10:42:44: Driver variadic_frequency: 4303.993 ms
```

I think TallFurryMan got it and goes one step further.
> You can help devs by measuring execution time and issuing a warning if that is exceeded (which also becomes part of the excess)

I can add something like:
```cpp
node->tgo += node->interval;
if (remainingTimerNode(node) < 0)
{
    DEBUG_MESSAGE("Timer %d is delayed.", node->tid);
    /* The duration of the function call is longer than the timer interval
     * or the single-threaded timer is busy.
     *
     * Perform the appropriate operation:
     * - Increase the interval
     * - Reduce function calling time
     * - Move to a separate thread.
     */
}
```
I think to add it in the next PR. I would like to implement moving timers to other threads.

Let's move on.

> You need to protect that data structure pointer from being dereferenced in a disconnected signal I think
Did you mean that?
```cpp
TimerPrivate::~TimerPrivate()
{
    stop(); // added
}
```
I agree with that, the timer should not call a function if the timer gets destroyed.

TallFurryMan:
That's fine. One thing I do not see clearly because I read the code quite quickly is how your periodic timer gets reinserted at the right place in your list. In any case this solution is very elegant.

pawel-soja:
Thanks!
Did I explain everything? Do you have any more questions? I will be happy to answer.

knro:
No you answered perfectly fine! I thought this replaced the TimerHit defaultdevice timer, so I was worried a bit, but it's all good now!

PR #1391, opened 2021-03-19
StreamManager - thread pool for uploadStream

pawel-soja:
Hi!

On a poor network connection indiserver <-> KStars, the buffer for frames gets full because sending blocks further processing of frames. I found that manually selecting the FPS limit for the preview is inconvenient.

In this PR, I mainly introduce `INDI::SingleThreadPool`.
```cpp
class SingleThreadPool
{
    DECLARE_PRIVATE(SingleThreadPool)
public:
    SingleThreadPool();
    ~SingleThreadPool();

public:
    /** @brief Reserves a thread and uses it to run functionToRun.
     *  A running function can check the 'isAboutToClose' flag and decide whether to end the work and give the thread. */
    void start(const std::function<void(const std::atomic_bool &isAboutToClose)> &functionToRun);

    /** @brief If thread isn't available at the time of calling, then this function does nothing and returns false.
     *  Otherwise, functionToRun is run immediately using thread and this function returns true. */
    bool tryStart(const std::function<void(const std::atomic_bool &isAboutToClose)> &functionToRun);

public:
    /** @brief Sets the 'isAboutToClose' flag to 'true' and waits for the end of running function. */
    void quit();

protected:
    std::shared_ptr<SingleThreadPoolPrivate> d_ptr;
};
```

Now we can easily send as many frames as possible in a separate thread.
```cpp
previewThreadPool.tryStart(std::bind([this, &previewElapsed](const std::atomic_bool &isAboutToQuit, std::vector<uint8_t> frame){

    INDI_UNUSED(isAboutToQuit);
    previewElapsed.start();
    uploadStream(frame.data(), frame.size());
    StreamTimeNP[0].setValue(previewElapsed.nsecsElapsed() / 1000000000.0);
    StreamTimeNP.apply();

}, std::placeholders::_1, std::move(*sourceBuffer)));
```
Of course, the FPS limit setting still works.

Additionally, the sending time is measured.
**Video Stream Delay** - for RPi4 is 800ms (Compress) / 130ms (Raw)

image

Regards,
Paweł

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -165,6 +165,7 @@ include_directories( ${CMAKE_CURRENT_SOURCE_DIR}/libs)
 include_directories( ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase)
 include_directories( ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property)
 include_directories( ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/timer)
+include_directories( ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/thread)
 include_directories( ${ZLIB_INCLUDE_DIR})
 include_directories( ${CFITSIO_INCLUDE_DIR})
 # 3. Build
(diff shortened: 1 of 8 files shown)


knro:
Thanks! I'll check this out.

Meanwhile, can you check this is not related to your recent changes? https://indilib.org/forum/ccds-dslrs/9151-asi-usb3-camera-retries-3-times-on-exposure-and-fails-git-builds.html#68903

pawel-soja:
Dear @knro, Thank you for showing me the topic on the forum, I don't always go to the forum.
I think the situation was clarified in the forum and it has nothing to do with my PR.
I will be happy to solve problems for the future (I mostly work on code at night).
The ASICamera2 library is not fully written correctly. 
usbfs_memory_mb is not my priority at the moment (I don't have such big cameras), maybe I will propose a good solution in the future.

Now with the streammanager/asi_ccd code almost clean, I can start working on performance - finally :)

knro:
Thanks, btw, anyway to reduce this delay? 130ms is still quite a large delay. would encoding it as jpeg helps? I'll test tonight. This is probably very close to as good as it gets considering the encoding and decoding that must take place.

pawel-soja:
You have to pay attention to the type of compression.
When choosing zlib (level 4), the situation is worsened.
The preparation and sending time over the network has increased from 130ms to 800ms.
```cpp
int ret = compress2(compressedFrame.data(), &compressedBytes, buffer, nbytes, 4);
```
Additional time overhead is associated with conversion to Base64.
To do:
- Compression that takes into account network and cpu/memory load.
- Base64 performance boost.
- Use of binary network traffic (without base64) - I've seen websockets in the code, haven't tested yet

You can figure something out. However I will come back to this issue in the future.
Now I will deal with the ASICamera2Boost library.

PR #1398, opened 2021-03-24
Added PegasusAstro new devices

chrysikos:
Changes to indi_dmfc_focuser driver in order to support  Devices (Pegasus ScopsOAG and FocusCube v2).

### DIFF
--- a/drivers.xml
+++ b/drivers.xml
@@ -359,7 +359,11 @@
             <driver name="Pegasus DMFC">indi_dmfc_focus</driver>
             <version>1.0</version>
         </device>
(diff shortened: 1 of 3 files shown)


knro:
Thank you, I presume this was tested with DMFC as well?

chrysikos:
> Thank you, I presume this was tested with DMFC as well?

Driver is test connected simultaneously with DMFC, FocusCube and ScopsOAG devices. Also, I did some test connecting two devices of the same type. All tests passed successfully.

knro:
@chrysikos Can you please check #1413? It looks like this PR caused a regression.

PR #1399, opened 2021-03-25
Fix warning - Correct use of strncpy.

pawel-soja:


drivers/focuser/steeldrive.cpp
```diff
@@ -987,7 +987,7 @@ bool SteelDrive::startMotion(FocusDirection dir)
 
     // inward  --> decreasing value --> DOWN
     // outward --> increasing value --> UP
-    strncpy(cmd, (dir == FOCUS_INWARD) ? ":F2MDOW0#" : ":F1MUP00#", STEELDRIVE_CMD);
```
knro, line 990: Shouldn't we just changed STEELDRIVE_CMD to 10 , and then change in tty_write to STEELDRIVE_CMD - 1 ? I presume this affects other commands as well?

pawel-soja:
This PR does not change the behavior of the driver, is intended to get rid of the warning when compiling.
I'm currently working on CI for indi / indi-3rdparty (linux / macos).

Regarding your question, the whole source needs to be reviewed, but first I want to clear the code from the warnings.

knro:
Thanks, I'll just add +1 to it and push.

pawel-soja:
Thanks, but you know this is not the correct solution?

```cpp
#include <stdio.h>
#include <string.h>

#define MAXRBUF 8

// #define STEELDRIVE_MAXBUF 16
#define STEELDRIVE_MAXBUF 4

#define STEELDRIVE_CMD    9

int main()
{
    const char *src = "123456789";

    printf("\nincorrect strncpy usage\n");
    {
        char err[MAXRBUF] = "ERROR";
        char dst[STEELDRIVE_MAXBUF] = "";

        strncpy(dst, src, STEELDRIVE_CMD + 1);

        printf("err: '%.*s'\n", int(sizeof(err)), err);
        printf("cmd: '%.*s'\n", int(sizeof(dst)), dst);
    }

    printf("\ncorrect strncpy usage\n");
    {
        char err[MAXRBUF] = "ERROR";
        char dst[STEELDRIVE_MAXBUF] = "";

        strncpy(dst, src, STEELDRIVE_MAXBUF); /* size argument - must protect the target buffer */

        printf("err: '%.*s'\n", int(sizeof(err)), err);
        printf("cmd: '%.*s'\n", int(sizeof(dst)), dst);
    }
    return 0;
}
```
Output:
```
incorrect strncpy usage
err: '56789'
cmd: '1234'

correct strncpy usage
err: 'ERROR'
cmd: '1234'
```

strncpy has never been a successful feature. When reading a buffer as a string you must check that it is zero-terminated.
I've also shown above what the correct use of printf is.

Therefore, safer functions have been developed:
_The strlcpy() and strlcat() functions copy and concatenate strings respectively. They are designed to be safer, more consistent, and less error prone replacements for strncpy(3) and strncat(3)._

knro:
oh thanks for the heads up, I never heard of strlcpy before. I think for now we can initialize cmd to zeros just to be on the safe side up until these function can be refactored in the future.

PR #1404, opened 2021-03-29
Fix warnings - MacOS

pawel-soja:
MacOS tests finally passed.

Changes:
- bsdtar support
- Refreshed implementation of base64 tests
- Replace nullptr with "" for 'me' variable.

Not all tests respect the global variable "me".

Conversion `const char *` to `std::string`
https://github.com/indilib/indi/blob/a2ebd7db5cc484a6f1383d5dc99a3f85afa211dd/libs/indibase/defaultdevice.cpp#L947-L948

https://github.com/indilib/indi/blob/a2ebd7db5cc484a6f1383d5dc99a3f85afa211dd/libs/indibase/defaultdevice.cpp#L1198-L1201

https://github.com/indilib/indi/blob/a2ebd7db5cc484a6f1383d5dc99a3f85afa211dd/indidriver.c#L47

If 'me' is null then we have an error:
```
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
```

### DIFF
--- a/indidriver.c
+++ b/indidriver.c
@@ -44,7 +44,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110 - 1301  USA
 
 static pthread_mutex_t stdout_mutex = PTHREAD_MUTEX_INITIALIZER;
 int verbose;      /* chatty */
-char *me = NULL;  /* a.out name */
+char *me = "";  /* a.out name */
 
 #define MAXRBUF 2048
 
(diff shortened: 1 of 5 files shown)


pawel-soja:
It seemed to me that I already touched upon this issue.
https://github.com/indilib/indi/blob/a2ebd7db5cc484a6f1383d5dc99a3f85afa211dd/indidrivermain.c#L103-L109
As I extend the tests, I will come back to this.

Issue #1412, opened 2021-04-03
Escaping initial Text


knro:
I just ran across an issue whereas the initial text needs to be escaped (It was from DSLR camera, so dynamic). IUFillText escapes name and label, but not initial text

https://github.com/indilib/indi/blob/bd9110b4a35f6848ad7880d7b74cae7c3a79bbfc/indidriver.c#L504

I thought about replacing it with this:

```
if (initialText && initialText[0])
    {
        char *escapedInitialText = malloc(strlen(initialText)*2);
        escapeXML2(escapedInitialText, initialText, strlen(initialText)*2);
        IUSaveText(tp, escapedInitialText);
        free(escapedInitialText);
    }
```

Do you think this is sufficient? is there a better way to do it?

pawel-soja:
All escape should be done when converting to XML and from XML. Unfortunately, this has not been done from the beginning and is now problematic. Now you have to make hotfixes that will solve and create new problems (if you want, I can describe it).

I propose to create a function that will do escape only when required.

For short: 
```cpp
escape(escape("Hel&lo")) == escape("Hel&lo") 
```
and 
```cpp
escape("He&amp;lo") == "He&amp;lo"
```
Then put the function in the IUSaveText function.

I'll be making fixes tonight.

pawel-soja:
My suspicions have been confirmed.
I quickly renamed the button in the example to `"Control Rain <FAIL>"`.
```cpp
bool RainDetector::initProperties()
{
    // Must init parent properties first!
    INDI::DefaultDevice::initProperties();

    IUFillLight(&RainL[0], "Status", "", IPS_IDLE);
    IUFillLightVector(&RainLP, RainL, 1, getDeviceName(), "Rain Alert", "", MAIN_CONTROL_TAB, IPS_IDLE);

    IUFillSwitch(&RainS[0], "On", "", ISS_OFF);
    IUFillSwitch(&RainS[1], "Off", "", ISS_OFF);
    IUFillSwitchVector(&RainSP, RainS, 2, getDeviceName(), "Control Rain <FAIL>", "", MAIN_CONTROL_TAB, IP_RW, ISR_1OFMANY, 0,
                       IPS_IDLE);

    return true;
}
```
```
2021-04-03T19:54:11: Driver ./tutorial_rain: tutorial_rain dispatch error: Property Control Rain <FAIL> is not defined in Rain Detector.
```
The structure holds the text `"Control Rain &lt;FAIL&gt;"`.

[...]

PR #1414, opened 2021-04-05
Fix XML Escape

pawel-soja:
Hi!

I moved the XML build to a separate file, indiuserio.c.
Changes:
- A secure XML leak (https://github.com/indilib/indi/issues/1412)
- Fewer calls to printf
- Fewer intermediary buffers
- Single function implementations

To minimize the number of buffers, and not to repeat the implementation for streams, sockets etc, I created a 'userio' (User Input-Output). Currently, it is an introduction to further development.

With a userio, you can write a universal code (INDI Style):
```cpp
void IUUserIOTextContext(const userio *io, void *user, const ITextVectorProperty *tvp)
{
    for (int i = 0; i < tvp->ntp; i++)
    {
        IText *tp = &tvp->tp[i];
        userio_prints    (io, user, "  <oneText name='");
        userio_xml_escape(io, user, tp->name);
        userio_prints    (io, user, "'>\n");
        userio_prints    (io, user, "      ");
        userio_xml_escape(io, user, tp->text ? tp->text : "");
        userio_prints    (io, user, "\n");
        userio_prints    (io, user, "  </oneText>\n");
    }
}
```

Now, you can define what functions are to be triggered:
```cpp
userio ioBaseClient;
/* ... */
ioBaseClient.write = [](void *user, const void * ptr, size_t count) -> size_t
{
    BaseClient *self = static_cast<BaseClient *>(user);
    return net_write(self->sockfd, ptr, count);
};

ioBaseClient.vprintf = [](void *user, const char * format, va_list ap) -> int
{
    BaseClient *self = static_cast<BaseClient *>(user);
    char message[MAXRBUF];
    vsnprintf(message, MAXRBUF, format, ap);
    return net_write(self->sockfd, message, strlen(message));
};
/* ... */
```

And finally, calling:
```cpp
FILE *fp;
/* ... */

IUUserIONewNumber(userio_file(), fp, nvp); // to file
IUUserIONewNumber(userio_file(), stdout, nvp); // to stdout
IUUserIONewNumber(&ioBaseClient, this, nvp); // send via BaseClient

```

Now I can make a button with any name!

```cpp
IUFillSwitchVector(
    &RainSP, RainS, 2, getDeviceName(),
    "Control \"'& &amp;lt; ", // brutal test
    "", MAIN_CONTROL_TAB, IP_RW, ISR_1OFMANY, 0,
    IPS_IDLE
);
```
image

On days, I'll add the rest.
@knro will you take a look?
You will always catch something ;)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -172,6 +172,8 @@ include_directories( ${CFITSIO_INCLUDE_DIR})
 
 SET(indiclient_C_SRC
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/lilxml.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/libs/userio.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indiuserio.c
     ${CMAKE_CURRENT_SOURCE_DIR}/base64.c)
 
 SET(indiclient_CXX_SRC
@@ -243,6 +245,8 @@ include_directories( ${CFITSIO_INCLUDE_DIR})
 message(STATUS "Building INDI Client with Qt5 support")
 SET(indiclientqt_C_SRC
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/lilxml.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/libs/userio.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indiuserio.c
     ${CMAKE_CURRENT_SOURCE_DIR}/base64.c)
 SET(indiclientqt_CXX_SRC
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/basedevice.cpp
@@ -412,6 +416,8 @@ SET(indidriver_C_SRC
     ${CMAKE_CURRENT_SOURCE_DIR}/indidrivermain.c
     ${CMAKE_CURRENT_SOURCE_DIR}/eventloop.c
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/lilxml.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/libs/userio.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indiuserio.c
     ${CMAKE_CURRENT_SOURCE_DIR}/base64.c
     )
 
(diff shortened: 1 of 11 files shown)


pawel-soja:
With double unescape, it's probably a matter of sequence. I will find it.

knro:
Thank you! This is awesome and would certainly pave the way for a more generic approach. Now gotta put it to the test

pawel-soja:
I got it!
All mechanisms work, so there are correct names underneath. So I started looking in the way it was displayed.

```cpp
/* build widgets for property pp using info in root.
 */
void INDI_P::initGUI()
{
    /* ... */
    labelW->setFrameShape(QFrame::Box);
    labelW->setFrameShadow(QFrame::Sunken);
    labelW->setMargin(2);
    labelW->setFixedWidth(PROPERTY_LABEL_WIDTH * KStars::Instance()->devicePixelRatio());
    labelW->setTextFormat(Qt::RichText);
    labelW->setAlignment(Qt::AlignVCenter | Qt::AlignLeft);
    labelW->setWordWrap(true);
    /* ... */
}
```

https://github.com/KDE/kstars/blob/1c964bf33b51cfe0b5efba3a78d92d4cae8b1293/kstars/indi/indiproperty.cpp#L128

Example:
image

knro:
Right, that what I suspected as well. However, I don't think we should be using PlainText, it should remain as rich text.

pawel-soja:
It has to be mentioned somewhere because when using the <> characters, some of the message may disappear.

@knro Qt version working?

KStars clientmanager.h
```cpp
#if INDI_VERSION_MAJOR >= 1 && INDI_VERSION_MINOR >= 5
        virtual void newUniversalMessage(std::string message) override;
#endif
```
But it is absent in baseclientqt.
I also gets deadlocks on blocking connections.
```
Qt: Dead lock detected while activating a BlockingQueuedConnection: Sender is ClientManager(0x5611aa6c8370), receiver is GUIManager(0x5611abcd30f0)
```

pawel-soja:
So rich text. I will paste links :)
image

pawel-soja:
Hi @knro, I tested. I think it is OK.

knro:
Sending links might actually be a good idea for some items :D but hopefully this won't get abused.

Issue #1422, opened 2021-04-12
possible race conditions from userio


knro:
Config files are getting corrupted since <message> messages are getting written the the config file as it gets saved:

e.g.

```
<INDIDriver>
    <message device="NightCrawler" timestamp="2021-04-12T08:43:21" message="[INFO] Saving device configuration..."/>
    <newSwitchVector device="NightCrawler" name="DEBUG">
        <oneSwitch name="ENABLE">
Off
        </oneSwitch>
        <oneSwitch name="DISABLE">
On
        </oneSwitch>
    </newSwitchVector>

```

It appears there needs to be a mutex to protect against contamination of different streams? @pawel-soja can you please check? This is a critical issue since it renders config files unusable now and corrupted.

pawel-soja:
The appearance of userio should not require mutex.
The "message" element should not be in config? Is there no closures "</INDIDriver>"?

knro:
No message is not part of the config file, this is intended for clients like all other messages. The config has one root node element <INDIDriver>...</INDIDriver>

knro:
Oh I see IUUserIOConfigTag now. How can we redirect  IDUserIOMessage from file io to network io?

pawel-soja:
Yes, my little mistake, give me a minute.

pawel-soja:
Thanks.

knro:
Thanks, it works great now!

PR #1423, opened 2021-04-12
Fix, use IDMessage to send info message.

pawel-soja:
Original implementation by introducing Userio:
```cpp
void IUSaveConfigTag(FILE *fp, int ctag, const char *dev, int silent)
{
    if (!fp)
        return;

    /* Opening tag */
    if (ctag == 0)
    {
        fprintf(fp, "<INDIDriver>\n");
        if (silent != 1)
            IDMessage(dev, "[INFO] Saving device configuration...");
    }
    /* Closing tag */
    else
    {
        fprintf(fp, "</INDIDriver>\n");
        if (silent != 1)
            IDMessage(dev, "[INFO] Device configuration saved.");
    }
}
```

### DIFF
--- a/indidriver.c
+++ b/indidriver.c
@@ -1542,7 +1542,21 @@ void IUSaveConfigTag(FILE *fp, int ctag, const char *dev, int silent)
     if (!fp)
         return;
 
-    IUUserIOConfigTag(userio_file(), fp, ctag, dev, silent);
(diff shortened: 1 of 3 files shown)


pawel-soja:
Separation of the implementation of the information message and writing to the stream.
Fix https://github.com/indilib/indi/issues/1422

PR #1426, opened 2021-04-14
PMC8 V2 Firmware update

karlrees:
Various changes for PMC8 firmware
- Fixes crash due to unexpected version string with v2 firmware
- Changes to be able to fetch model type
- Support for standard serial cable (no longer have to invert DTR)
- Increase stability of wireless connection / auto-reconnect
- Auto-detect if mount is already tracking when first connect
- Other minor changes


knro:
Thanks! Please check the CI errors.

karlrees:
Ooops.  Guess I wasn't compiling with warnings.  Fixed it.

PR #1430, opened 2021-04-18
INDI::BaseClient - Disconnect if sending has failed.

pawel-soja:
@knro I lost one commit. Closes the connection if sending has failed.
Check if you still have SIGPIPE.
I'll do more tests by constantly sending data and closing the connection.

### DIFF
--- a/libs/indibase/baseclient.cpp
+++ b/libs/indibase/baseclient.cpp
@@ -995,6 +995,11 @@ size_t INDI::BaseClient::sendData(const void *data, size_t size)
     }
     while(ret == -1 && (errno == EINTR || errno == EAGAIN || errno == EWOULDBLOCK));
 
+    if (ret < 0)
+    {
+        disconnectServer(-1);
+    }
+
     return std::max(ret, 0);
 }
 


knro:
On raspberry pi, it's now getting stuck when KStars disconnect from INDI server on line baseclient.cpp:311

Maybe throttle this function? add sleep between attempts and a retry count? so it doesn't end up in infinite loop. Unless something else caused this to suddenly fail now, maybe the socket is already closed but not reset?

pawel-soja:
oh I see. Two fixes will be needed here. It will be fixed today.

pawel-soja:
I found another deadlock in KStars
```cpp
void ClientManager::newProperty(INDI::Property *prop)
{
    // Do not emit the signal if the server is disconnected or disconnecting (deadlock between signals)
    if (!isServerConnected())
    {
        IDLog("Received new property %s for disconnected device %s, discarding\n", prop->getName(), prop->getDeviceName());
        return;
    }

    // =================================================================================
    // The above condition is correct if all operations are on one thread.
    // Unfortunately, in the meantime performing this function, disconnection can occur.
    // now or somewhere below isServerConnected() if false
    // =================================================================================

```

pawel-soja:
I found a problem.
If Kstar calls `disconnectServer`, then the `listenINDI->dispatchCommand -> buildprop -> newProperty -> emit newINDIProperty` is blocked because this thread has calls the `disconnectServer` function.
Therefore, there is no synchronization between opening and closing connections and receiving data.

I'll fix it, so that it will be resistant to all use cases.

knro:
Great, thanks for looking into it. Basically deadlock may occur due to use of BlockingQueuedConnection to block the sending thread until the receiving thread in KStars finishes building or property or removing it..etc. Blocking the sending thread is not really optimal, but as you already found out, there is no proper synchronization for new/delete properties unless this blocking method is used.

Without it, BaseDevice properties may become invalid and KStars has no idea about that which causes it to crash when it tries to access such properties.

Issue #1424, opened 2021-04-12
Baseclient SIGPIPE crash when disconnected abnormally


knro:
This is a rather old issue, but it still happens in GIT. When an INDI client server connection is terminated due to server termination, it appears the sockfd is not properly closed/recycled. On next connection, the whole app crashes in SIGPIPE, this is repeatable. Attached in the backtrace.

@pawel-soja I'm going to start investigating this more, but thought you might already have some idea on this?

[sigpipe.txt](https://github.com/indilib/indi/files/6296256/sigpipe.txt)

pawel-soja:
It is easy to solve. I'm on vacation now. I'll have a look at it over the weekend.

TallFurryMan:
I think I had a look to this at some point. It was related to the management of signals through the socket. Basically, let the program manage signals instead of propagating them as real signals. And there was a difference between *nix and Win32 apis there.

pawel-soja:
I see we need to take care of the *nix / Windows and Qt versions. I can separate sockets into a separate class that will be portable. But a small version of the Qt library is made.
It would be a lot easier to switch to Qt completely. - all project (INDI / INDI-3rdParty).
Both ways are OK for me, as I create a similar project to Qt, using a newer approach to slot signals (vtable manipulation). Libraries will be opened in the future. The library is modular and lightweight.

What do you think?

knro:
So I thought about this long ago, and adding Qt dependency is an overkill for INDI. Especially many clients do not use Qt and do not want to use Qt in their systems.

knro:
At any rate, any way to dealing with this now so it handles abnormal disconnects more gracefully? It's important for v1.9.0 INDI release.

pawel-soja:
@knro When do you want to release this version?

When analyzing the code, I can see more socket problems.

I can throw in some hotfixes to fix some problems or upload a completely new implementation consisting of:
- EventDispatcher - you can dynamically add sockfd / timers to listen for events
- EventLoop - starts a loop to handle EventDispatcher
- Thread - thread support - you can use multiple EventDispatchers.
- TcpSocket - as the name suggests.


The network connection operation is as follows:
```cpp
int main(int, char*[])
{
    EventLoop e;

    TcpSocket socket;

    socket.callOnReadyRead([&](){
        char data[1024];
        size_t size = socket.read(data, sizeof(data));

        cout << "Got data " << size << " Bytes" << endl;
        cout << data << endl;
    });

    socket.callOnConnected([&](){
[...]

PR #1432, opened 2021-04-20
INDI::BaseClient - Fix sockets deadlock.

pawel-soja:
https://github.com/indilib/indi/pull/1430 https://github.com/indilib/indi/issues/1424

The function closing the connection is non-blocking.
Closing descriptors only happens on the thread.
The connection opening and closing functions are synchronized - There is no race.
Waiting for a thread to finish with a timeout is only implemented in the destructor.

@knro Can you check it's ok? (on windows)
I found some additional problems but completely unrelated to sockets.

### DIFF
--- a/libs/indibase/baseclient.cpp
+++ b/libs/indibase/baseclient.cpp
@@ -31,6 +31,7 @@
 #include <stdarg.h>
 #include <cstring>
 #include <algorithm>
+#include <chrono>
 #include <functional>
 #include <assert.h>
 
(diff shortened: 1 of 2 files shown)


knro:
Thanks! Just tested this with test_ekos_simulator in KStars, which actually was built to test these cases and it crashed. Attached backtrace while I investigate more.
[backtrace.txt](https://github.com/indilib/indi/files/6350454/backtrace.txt)

knro:
So after removing BlockingQueueConnection, it appears to work OK, but let me perform more testing and let me test other clients.

knro:
All tests passed OK and PHD2 also works OK, so looks it's good to go for now.

pawel-soja:
@knro Is this error easily triggered?
I was able to trigger this error by quickly clicking connect / disconnect.

There are still a few things to do. I did not have time to analyze the lifetime of the objects to see if removing BlockingQueueConnection is safe.

If the current fixes are satisfactory, let me come back to this topic later.
Currently I want to finish CI for a 3rd party.
Not to mention my main target - cameras (in this case, i will buy the same cameras as you have).

knro:
@pawel-soja The fixes I submitted are still pending in this MR: https://invent.kde.org/education/kstars/-/merge_requests/274

Once it's merged, it should be stable as I tested thoroughly yesterday on both PC and Raspberry PI.

PR #1436, opened 2021-04-26
Fix Philips Toucam family cameras (pwc driver) ManualExposure switch (label is "Gain, Automatic" )

d6p:
Works on my setup (raspberry pi4 kernel 5.10.31-vl7+ Kstars 3.5.2 Stable) ...
Please review.

### DIFF
--- a/drivers/video/v4l2driver.cpp
+++ b/drivers/video/v4l2driver.cpp
@@ -1787,7 +1787,7 @@ void V4L2_Driver::updateV4L2Controls()
     {
         defineProperty(&Options[i]);
 
-        if (strcmp(Options[i].label, "Exposure, Auto") == 0 || strcmp(Options[i].label, "Auto Exposure") == 0)
+        if (strcmp(Options[i].label, "Exposure, Auto") == 0 || strcmp(Options[i].label, "Auto Exposure") == 0 || strcmp(Options[i].label, "Gain, Automatic")==0)
         {
             ManualExposureSP = Options + i;
             LOGF_DEBUG("- %s (used for manual/auto exposure control)", Options[i].label);


knro:
How does the gain corresponding to exposure? can you elaborte?

d6p:
in linux kernel pwc driver ( https://github.com/raspberrypi/linux/blob/rpi-5.10.y/drivers/media/usb/pwc/pwc-v4l.c ), 
for webcam models >=700 , i.e. CODEC3 type, you can see the following comment: 
/* CODEC3: both gain and exposure controlled by autogain */
and the device reports only the autogain boolean:

astroberry@astroberry:~ $ v4l2-ctl --all
Driver Info:
        Driver name      : pwc
        Card type        : Philips 740 webcam
        Bus info         : usb-0000:01:00.0-1.3
        Driver version   : 5.10.31
        Capabilities     : 0x85200001
                Video Capture
                Read/Write
                Streaming
                Extended Pix Format
                Device Capabilities
        Device Caps      : 0x05200001
                Video Capture
                Read/Write
                Streaming
                Extended Pix Format
Priority: 2
Video input : 0 (Camera: ok)
Format Video Capture:
        Width/Height      : 640/480
        Pixel Format      : 'YU12' (Planar YUV 4:2:0)
        Field             : None
        Bytes per Line    : 640
        Size Image        : 460800
        Colorspace        : sRGB
        Transfer Function : Default (maps to sRGB)
        YCbCr/HSV Encoding: Default (maps to ITU-R 601)
        Quantization      : Default (maps to Limited Range)
        Flags             :
Streaming Parameters Video Capture:
        Capabilities     : timeperframe
        Frames per second: 15.000 (15/1)
        Read buffers     : 2

User Controls

                     brightness 0x00980900 (int)    : min=0 max=127 step=1 default=89 value=89 flags=slider
                       contrast 0x00980901 (int)    : min=0 max=63 step=1 default=32 value=32 flags=slider
                     saturation 0x00980902 (int)    : min=-100 max=100 step=1 default=0 value=0 flags=slider
        white_balance_automatic 0x0098090c (menu)   : min=0 max=4 default=3 value=3 flags=update
                    red_balance 0x0098090e (int)    : min=0 max=255 step=1 default=132 value=132 flags=slider
                   blue_balance 0x0098090f (int)    : min=0 max=255 step=1 default=66 value=66 flags=slider
                          gamma 0x00980910 (int)    : min=0 max=31 step=1 default=0 value=0 flags=slider
                       exposure 0x00980911 (int)    : min=0 max=255 step=1 default=200 value=200
                 gain_automatic 0x00980912 (bool)   : default=0 value=0 flags=update
                           gain 0x00980913 (int)    : min=0 max=63 step=1 default=42 value=42
         backlight_compensation 0x0098091c (bool)   : default=0 value=0
                  color_effects 0x0098091f (menu)   : min=0 max=1 default=0 value=0
               band_stop_filter 0x00980921 (bool)   : default=0 value=0
                   auto_contour 0x0098f900 (bool)   : default=0 value=0 flags=update
                        contour 0x0098f901 (int)    : min=0 max=63 step=1 default=0 value=0 flags=slider
        dynamic_noise_reduction 0x0098f902 (int)    : min=0 max=3 step=1 default=0 value=0
       auto_white_balance_speed 0x0098f903 (int)    : min=1 max=32 step=1 default=1 value=1
       auto_white_balance_delay 0x0098f904 (int)    : min=0 max=63 step=1 default=0 value=0
             save_user_settings 0x0098f905 (button) : flags=write-only, execute-on-write
          restore_user_settings 0x0098f906 (button) : flags=update, write-only, execute-on-write
       restore_factory_settings 0x0098f907 (button) : flags=update, write-only, execute-on-write

Since Philips webcam were quite popular at their time it is possible many indi users still have those in the closet, so I thought it would be nice if it worked. Don't know if ithe fix may mess with other V4L cameras ? Maybe it would be safer to make 2 passes and only set if no autoexposure control is found ?

knro:
Right, I agree it would be more safer this way.

d6p:
Thinking about it, there is no way to be safe unless we know for sure the device is a pwc CODEC3 type ... and we lose the ability to keep autogain/exposure while stacking, which may have a use case... So forget it, sorry for the inconvenience. Will try to fix the crash instead.

PR #1447, opened 2021-05-05
ScopeDome driver update

jpaana:
Finally long overdue update to the ScopeDome driver I've worked on and off for more than a year now :) Finally got testers for the new Arduino Card controller (I only got the controllers from ScopeDome, but they weren't connected to anything) to get that work completed and fixed last regression from the older USB Card 2.1 path so I've now run this code myself for couple of weeks without issues. Progress got documented mostly in INDI forum thread at https://www.indilib.org/forum/domes/7648-cannot-connect-to-2m-scopedome-arduino.html

Other than that the code got cleaner and now uses the new PropertyWidget types, which simplify the code even further.

Main changes:
- abstract card specific functionality better to support very different controller types
- add support for new ScopeDome Arduino Card controller, via USB and ethernet/HTTP connection
- modernize property handling to use Widgets
- add more and better debug logging
- add option to control home sensor polarity (active high/low)
- remove custom shutter handling on park/unpark to use common code

### DIFF
--- a/drivers/dome/scopedome_dome.cpp
+++ b/drivers/dome/scopedome_dome.cpp
@@ -1,7 +1,7 @@
 /*******************************************************************************
  ScopeDome Dome INDI Driver
 
- Copyright(c) 2017-2019 Jarno Paananen. All rights reserved.
+ Copyright(c) 2017-2021 Jarno Paananen. All rights reserved.
 
  based on:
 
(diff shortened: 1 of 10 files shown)


drivers/dome/scopedome_dome.cpp
```diff
@@ -100,103 +109,64 @@ bool ScopeDome::initProperties()
...
 
-    // Set serial parameters
-    serialConnection->setDefaultBaudRate(Connection::Serial::B_115200);
+    CardTypeSP[0].fill("USB_CARD_21", "USB Card 2.1", ISS_ON);
+    CardTypeSP[1].fill("ARDUINO", "Arduino Card", ISS_OFF);
+    CardTypeSP.fill(getDeviceName(), "CARD_TYPE", "Card type",
+                    CONNECTION_TAB, IP_RW, ISR_1OFMANY, 0, IPS_OK);
+
+    defineProperty(&CardTypeSP);
+    if(!loadConfig(true, CardTypeSP.getName()))
```
knro, line 160: Maybe use IUGetOnSwitchIndex(...) directly before defining CardTypeSP? This way you can avoid loadConfig which can be expensive.

drivers/dome/scopedome_dome.cpp
```diff
@@ -100,103 +109,64 @@ bool ScopeDome::initProperties()
...
+    {
+        // Set serial parameters
+        // This is tricky as Arduino card communicates at 9600 bauds, but
+        // USB Card 2.1 uses 115200 bauds
+        serialConnection->setDefaultBaudRate(Connection::Serial::B_115200);
+    }
+
+    defineProperty(CredentialsTP);
+    loadConfig(true, CredentialsTP.getName());
```
knro, line 169: Maybe IUGetConfigText ?

drivers/dome/scopedome_dome.cpp
```diff
@@ -242,383 +209,423 @@ bool ScopeDome::SetupParms()
...
+
+    sim = isSimulation();
+
+    if (sim)
+    {
+        interface.reset(static_cast<ScopeDomeCard *>(new ScopeDomeSim()));
+    }
+    else
+    {
+        if(CardTypeSP[0].getState() == ISS_ON)
```
knro, line 248: Better use enums instead of CardTypeSP[0]. i.e. CardTypeSP[CARD_USB] ...etc

jpaana: True and with FindOnSwitchIndex much cleaner

drivers/dome/scopedome_dome.cpp
jpaana: Yup, though it was a bit clunky with the new PropertyWidget type, but not too bad.

drivers/dome/scopedome_dome.cpp
jpaana: Same comment here, could be cleaner without repeating the names, opportunity for future usability improvement

Issue #1449, opened 2021-05-06
Update BaseMediator and BaseClient


knro:
Along with the recent changes to Properties, the BaseMediator, BaseDevice, and BaseClient classes need to be updated accordingly. Of the important things that caused issues in the past is use of raw pointers to communicate INDI::Property. We should use shared_ptr so that the object is reserved until all parties stop using it.

@pawel-soja Besides the migration to shared_ptr, do you have any suggestions? Let's try to find a way to deprecate the existing interface while allow existing clients to be build against it without requiring to be re-written again. Might be a good option as well to put the D pointer to use here as well.

pawel-soja:
@knro 
Currently
- [INDI::Property](https://github.com/indilib/indi/blob/master/libs/indibase/property/indiproperty.h)
- [INDI::PropertyBasic template](https://github.com/indilib/indi/blob/master/libs/indibase/property/indipropertybasic.h) See below
- [INDI::PropertyBlob](https://github.com/indilib/indi/blob/master/libs/indibase/property/indipropertyblob.h)
- [INDI::PropertyLight](https://github.com/indilib/indi/blob/master/libs/indibase/property/indipropertylight.h)
- [INDI::PropertyNumber](https://github.com/indilib/indi/blob/master/libs/indibase/property/indipropertynumber.h)
- [INDI::PropertySwitch](https://github.com/indilib/indi/blob/master/libs/indibase/property/indipropertyswitch.h)
- [INDI::PropertyText](https://github.com/indilib/indi/blob/master/libs/indibase/property/indipropertylight.h)

They have a built-in `shared_ptr`.
So the copies of the classes store the same data with a counted reference to the data.
```cpp
class EasySharedVirtualClass
{
public:
    EasySharedVirtualClass() = default;
    virtual ~EasySharedVirtualClass() = default;

private:
    void * d_ptr; // or std::shared_ptr
};
/* ... */
std::cout << "sizeof void *                 " << sizeof(void *) << std::endl;
std::cout << "sizeof EasySharedVirtualClass " << sizeof(EasySharedVirtualClass) << std::endl;
std::cout << "sizeof INDI::Property         " << sizeof(INDI::Property)         << std::endl;
std::cout << "sizeof INDI::PropertySwitch   " << sizeof(INDI::PropertySwitch)   << std::endl;
```
Output:
```
sizeof void *                 8
sizeof EasySharedVirtualClass 16
sizeof INDI::Property         16
sizeof INDI::PropertySwitch   16
[...]

PR #1453, opened 2021-05-11
BaseDevice::getProperty as non-pointer result

pawel-soja:
Hi!

I have an idea of how to get rid of completely raw pointers in the future.

Linked: https://github.com/indilib/indi/issues/1449

### Current implementation
For example:
https://github.com/indilib/indi/blob/07e51770a2d5c1853d931b66656ab3836b0941b5/libs/indibase/basedevice.h#L147-L155

Current usage example:
```cpp
INDI::Property *pointerProperty = device->getProperty("HELLO");
```
If the BaseDevice object is destroyed, the pointer expires.

### Current solution
This problem can be solved already when using the function, e.g. like this:
```cpp
INDI::Property sharedProperty = *device->getProperty("HELLO");
```

### Idiot-proof solution
The current possibility to solve the problem is not ideal because you have to remember this '_Current solution_'.

`INDI::Property` have a built-in `std::shared_ptr` that enables secure data sharing. A nice solution would be for the function to return an object, not a pointer.

```cpp
Property getProperty(const char *name, INDI_PROPERTY_TYPE type = INDI_UNKNOWN) const; 
```

Unfortunately, this breaks the backward compatibility.

However, the conversion to pointer operator can take care of this.
```cpp
Property::operator INDI::Property *()
{
    return
        /* 1 */ isValid() ?
        /* 2 */ THIS :
        /* 3 */ nullptr;
}
```

1. Return `nullptr` if Property is invalid - `type == INDI_UNKNOWN`
    This is very important (backward compatibility), for comparing a pointer to nullptr.
     `INDI::Property() == nullptr` returns `true` and `INDI::PropertySwitch() == nullptr` returns `false`
2. `this` cannot be used. The function returns the object, the operator returns the address to it, and then the object is destroyed. We have a pointer to nowhere. The solution is to create an `Property` object that stores shared data (already exists) and place it in the data.
3. See 1. ;)

### Finally
You can use new ```Property BaseDevice::getProperty(....)``` function!

Old unsafe way:
```cpp
INDI::Property *prop = device->getProperty("UNSAFE");
if (prop != nullptr)
    std::cout << prop->getStateAsString() << std::endl;
```

Ane new safe way:
```cpp
INDI::Property prop = device->getProperty("SAFE");
if (prop != nullptr) // or prop.isValid()
    std::cout << prop.getStateAsString() << std::endl;
```

Or (preferred)
```cpp
auto prop = device->getProperty("SAFE");
if (prop.isValid())
    std::cout << prop.getStateAsString() << std::endl;
```

## Give it a try!
It is really safe even if you don't check if the object exists.
```cpp
#include <indibase.h>
#include <indiproperty.h>
#include <iostream>

int main(int, char*[])
{
    std::cout << INDI::BaseDevice().getProperty("non-exist").getStateAsString() << std::endl; // prints "Alert"
    return 0;
}
```

### What's next
I think I can rewrite most of the functions this way, we can add `deprecated` at a later stage.
Operators will spit out warnings to migrate completely to a safe form of function handling.

The case with virtual functions is a bit worse. Of course, all mechanisms will work, but it will require changes, e.g.
```cpp
class Foo: public INDI::BaseMediator
{
public:
    // void newProperty(INDI::Property *property) override; // old API
    void newProperty(INDI::Property property) override; // new API
```
The transition will be simple, it doesn't require changing the implementation of the function.
Remember `INDI::Property property` still behaves like a pointer.

## To Merge Or Not To Merge
I would definitely ask you to check if everything is ok.
@knro you are a specialist, I'll be glad if you find _(or don't find)_ a problem. ;)

If everything is OK, I will keep developing the concept. You can then think about making changes gradually.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -183,6 +183,7 @@ SET(indiclient_CXX_SRC
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/basedevice.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/baseclient.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indiproperty.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indiproperties.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indipropertybasic.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indipropertytext.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indipropertynumber.cpp
@@ -255,6 +256,7 @@ SET(indiclientqt_CXX_SRC
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/basedevice.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/baseclientqt.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indiproperty.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indiproperties.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indipropertybasic.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indipropertytext.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indipropertynumber.cpp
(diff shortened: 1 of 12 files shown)


knro:
I see... thank you for the detailed explanation. Ok, looks like lot of work has to be done to maintain backward compatibility. I think we have another option as well: Next version of INDI Library can be 2.0.0 instead of 1.9.1 so we can comfortably break backward compatibility and existing clients may keep linking to INDI version 1.9.X.

pawel-soja:
@knro I think I have resolved the getProperty/getProperties issue. I would put further modifications in the new PR.
So now I am waiting for your final tests and merge.

In 2.0.0, I would optimize the IXXX/IXXXVectorProperty structures to remove a lot of C/C++ code, and also changing virtual methods. At the moment, it seems like I can rewrite all non-virtual functions to keep backwards compatibility.

knro:
Thanks, I've tested this with KStar's test_ekos_simulator and it crashes it. Investigating....

pawel-soja:
Current KStars master branch?

knro:
Correct.

pawel-soja:
Can you upload any logs/backtrace?

knro:
Never mind, I just rebuilt and the tests and they went through just fine. Sorry for the alarm.

Issue #1456, opened 2021-05-16
indi_paramount_telescope driver takes too long to report pulseguide move result


interplanetarychris:
**Describe the bug**
DirectGuide requests take too long to respond to the requester, resulting in long-response times (9.1 seconds) for short requests (37ms guide pulse).  This results in inconsistent/stuttered images in the guider, and poor guiding performance.

I've posted details about this problem on the [PHD2 forum](https://groups.google.com/g/open-phd-guiding/c/1ezYaVY_ss0)

It appears that DirectGuide requests are made **synchronously** - although that may or may not be problematic.

**To Reproduce**
Any guiding which requests a guide correction.

**Expected behavior**
I would expect guiding to occur at the frequency of the guide imaging duration, but instead, PHD2 often waits several multiples of the guide-image duration before hearing a response from the driver, and initiating the next image.

**Environment:**
 - OS: Ubuntu 20.04.2 LTS
 - Version: indi_paramount_telescope v1.1

**Log Files**
https://www.dropbox.com/s/l6n42p9wvy26798/indi_paramount_telescope_guiding.zip?dl=0

interplanetarychris:
I compiled and tested a version that sends the pulses asynchronously — but it appeared to have no effect.  PHD2 is still sometimes waiting ~8 seconds for completion status of a 90ms pulse.

Examining the PHD2 response, it looks like the paramount driver needs to more promptly send a "GuideComplete" call in order to provide a IDSetNumber result to PHD2, but this looks like unimplemented functionality in the paramount driver.  I'll examine the "timed guide" code in other drivers to try and figure out an implementation appropriate for the Paramount.

knro:
Thank you for the report. Please check how the other drivers are implementing this functionality as well to see if there is any lag from the driver itself.

interplanetarychris:
Per the PR https://github.com/indilib/indi/pull/1458, I believe I've addressed the bug.

PR #1458, opened 2021-05-18
TheSky Server TCP read fixes

interplanetarychris:
I believe I've fixed the "Guiding problem", Issue #1456, which was primarily caused by the underlying driver having general problems with polling theSkyX TCP server.

It appears that it was listening to the TCP response for TIMEOUT (3 sec) duration, but not triggering a timeout error, as it successfully read bytes into the buffer.  This was for EVERY action which performed a `tty_read_section` in order to determine a command response.  This was noticeable in the background RaDec coordinate polling, which was returning every 3 seconds, even though configured every 1000ms.

I've updated the commanding and readout to insert a '#' character as the stop character, and subsequently re-coded all the result parsing to take this into account.  It seemed that the regex examination of the result was unnecessary, and instead of parsing the error, I opted to return the offending failure string, in full, to the user.  The readout of results is accomplished with error-checked `sscanf` calls.

I've also removed other duplicate code, unused code blocks, and put in additional comments throughout.

It has been _lightly_ tested against my Paramount MX+ (parked in an enclosure during inclement weather), and more thoroughly tested against "Software Bisque Telescope Mount Simulator" and the INDI Camera simulator (for guiding). I also induced various guide errors through PHD2 manually guide pulses.  All but the Pier Side and Slew rates seem to check out, but I don't believe the Bisque simulator delivers meaningful results for those two items.

While the Guiding code is still implemented synchronously — serial implementation of guide moves if RA + DEC requests are received in the same cycle — the response is very quick, about 100-150ms.

There might be some residual error handling to iron out - like if TheSkyX is disconnected out of order of bringing down the connections in PHD2 or INDI, but the functionality is at least the same as the version it's upgrading.

### DIFF
--- a/drivers/telescope/paramount.cpp
+++ b/drivers/telescope/paramount.cpp
@@ -1,9 +1,13 @@
 /*******************************************************************************
  Copyright(c) 2017 Jasem Mutlaq. All rights reserved.
+ 2021 Chris Lewicki. Refactor of TCP connections and error handling  
 
- Driver for using TheSky6 Pro Scripted operations for mounts via the TCP server.
- While this technically can operate any mount connected to the TheSky6 Pro, it is
+ Driver for using TheSkyX Pro Scripted operations for mounts via the TCP server.
+ While this technically can operate any mount connected to the TheSkyX Pro, it is
  intended for Paramount mounts control.
+
+ Ref TheSky Functions: 
+ https://www.bisque.com/wp-content/scripttheskyx/classsky6_r_a_s_c_o_m_tele.html
 
  This library is free software; you can redistribute it and/or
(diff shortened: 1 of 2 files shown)


knro:
Thanks, I'll ask someone to test the nightly once it becomes available tomorrow.

interplanetarychris:
I night-sky tested this last night, and everything appears in order - including Lunar track rate and pier side.  

There could be some improvements in the TCP socket read error handling (completing the read doesn't equivocate to emptying the buffer), but I'll continue to work on that and submit those separately.

knro:
@interplanetarychris Thank you, I just pushed another improvement as well to make the operation asynchronous. Can you please test it?

PR #1460, opened 2021-05-18
Fixing Location issue with Rainbow RST-135

jscheidtmann:
See discussion at https://indilib.org/forum/ekos/9645-rst-135-location-issue.html#71411

### DIFF
--- a/drivers/telescope/lx200driver.cpp
+++ b/drivers/telescope/lx200driver.cpp
@@ -574,7 +574,7 @@ int getSiteLatitude(int fd, int *dd, int *mm, double *ssf)
     DEBUGFDEVICE(lx200Name, DBG_SCOPE, "RES <%s>", read_buffer);
 
     *ssf = 0.0;
-    if (sscanf(read_buffer, "%d%*c%d:%lf", dd, mm, ssf) < 2)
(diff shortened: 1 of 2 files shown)


knro:
Thanks! Was this tested as well in other mounts to verify no regressions? I'll test locally as well.

jimmy-ck:
Jasem, no tests so far on any mounts! Be careful.

jimmy-ck:
Built and tested on RPi4 with RST-135 mount: Location pull from mount seems to work now, but there is another fault I think. Obviously KStars/Ekos writes back into the mount after connection process and overrides location information into the EEPROM of the mount, although KStars/INDI is set to "Mount updates KStars". Seems like updateLocation() is called somewhere after starting Ekos. Simply disconnecting and connecting devices in Ekos does not override location in mount, but a restart of INDI Profile and reconnecting does.
Or is this a correct behaviour which I am misunderstanding?

jscheidtmann:
@jimmy-ck This behaviour re upadteLocation() is a different problem and will not be addressed by this pull-request. I suggest you open an issue. I am closing this pull request, as it seems there's still a problem with reading time from the mount.

PR #1462, opened 2021-05-20
Replace gethostbyname with getaddrinfo

paolostivanin:
As per the man page (`man 3 gethostbyname`), both `gethostbyname` and `gethostbyaddr` are obsolete functions that should be replaced by `getaddrinfo` and `getnameinfo`.

What's been tested:
- [x] compile against latest master :heavy_check_mark: 
- [x] compile against latest stable (1.9.0) :heavy_check_mark: 
- [x] test with hardware (tried with Nikon + ASI, all good. Capturing image, stopping server, restarting server, no issues) :heavy_check_mark: 

I've opened the PR before the "real world scenario test", so I can get a code review in the meanwhile :)

### DIFF
--- a/indiserver.c
+++ b/indiserver.c
@@ -683,31 +683,38 @@ static void startRemoteDvr(DvrInfo *dp)
  */
 static int openINDIServer(char host[], int indi_port)
 {
-    struct sockaddr_in serv_addr;
-    struct hostent *hp;
-    int sockfd;
+    struct addrinfo *result, *ptr;
+    struct addrinfo hints = {};
+    hints.ai_family = AF_UNSPEC;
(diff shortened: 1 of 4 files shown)


paolostivanin:
Here the repos for 15.3 and Tumbleweed, in case any openSUSE users would like to test this: https://build.opensuse.org/package/show/home:polslinux:teststuff/indi-git

knro:
Thank you! Will check it out.

paolostivanin:
I did some testing with real hardware connected to the mini pc, and everything went fine. Next I could try remote testing (e.g. start indi on device and then connect there from another pc using kstars)

knro:
Thank you, I need to run tests as well in different configuration.

PR #1463, opened 2021-05-22
Split spectrograph into receiver and spectrograph classes

iliaplatone:
Here I added RTL-SDR into indi main repository.
I also splitted the spectrograph class into receiver and spectrograph.
Receiver is meant for single spectral lines readout, within a defined bandwidth or sampling rate.
Spectrograph is for multiline spectra sampled reading.


libs/indibase/indireceiver.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+bool Receiver::initProperties()
+{
+    // Receiver Info
+    IUFillNumber(&ReceiverSettingsN[RECEIVER_GAIN], "RECEIVER_ANTENNA", "Antenna", "%16.2f", 1, 4, 1, 1);
+    IUFillNumber(&ReceiverSettingsN[RECEIVER_FREQUENCY], "RECEIVER_ANTENNA", "Antenna", "%16.2f", 1, 4, 1, 1);
+    IUFillNumber(&ReceiverSettingsN[RECEIVER_BANDWIDTH], "RECEIVER_ANTENNA", "Antenna", "%16.2f", 1, 4, 1, 1);
+    IUFillNumber(&ReceiverSettingsN[RECEIVER_BITSPERSAMPLE], "RECEIVER_ANTENNA", "Antenna", "%16.2f", 1, 4, 1, 1);
+    IUFillNumber(&ReceiverSettingsN[RECEIVER_SAMPLERATE], "RECEIVER_ANTENNA", "Antenna", "%16.2f", 1, 4, 1, 1);
+    IUFillNumber(&ReceiverSettingsN[RECEIVER_ANTENNA], "RECEIVER_ANTENNA", "Antenna", "%16.2f", 1, 4, 1, 1);
```
knro, line 61: Correct names and labels

knro (changes requested):
Thanks! I also highly recommend you take a look at the new INDI::Property introduced by @pawel-soja . I think we should start using this for all new drivers/frameworks as we move slowly to migrate the existing system to this.

@pawel-soja I'll work on modifying the tutorials, do you think you can update the documentation so that prospective developers start using the new system?

jpaana:
Even though I have an RTL-SDR stick myself, I'm not overly keen on adding a new mandatory 3rd-party library dependency for a single device that probably has relatively few users and might not be available in all platforms INDI has compiled for before. Therefore I'd vote for leaving it in the 3rd-party repo.

knro:
I agree, we should make this optional. @iliaplatone Can you please remove this hard dependency? If it exists, we can build this, if not it's gracefully skipped.

iliaplatone:
I think we need at least a device driver for the receiver or spectrograph class, the RTL-SDR is used more than other devices included into the main repository in its applications range, which is wide enough also.

knro:
It's not available in Fedora as well AFAIK. Not sure about other platforms. How about BSD? MacOS?

knro:
Ok, it's optional now, but the debian package would still build with it. So that leaves other distros to optional support it as they see fit.

xsnrg:
This is the version that would get pulled in with Fedora. It does seem to be available , I just need to add it to the spec to pull it in.

$ sudo dnf info rtl-sdr-devel.x86_64
Last metadata expiration check: 0:38:24 ago on Mon 24 May 2021 03:32:50 PM CDT.
Available Packages
Name         : rtl-sdr-devel
Version      : 0.6.0
Release      : 8.fc33
Architecture : x86_64
Size         : 12 k
Source       : rtl-sdr-0.6.0-8.fc33.src.rpm
Repository   : fedora
Summary      : Development files for rtl-sdr
URL          : http://sdr.osmocom.org/trac/wiki/rtl-sdr
License      : GPLv2+
Description  : Development files for rtl-sdr.

xsnrg:
The latest changes allowed it to compile without changes to the spec:


Package:  indi-1.9.1.git-20210524210448.fc34
COPR:     xsnrg/libindi-bleeding
Built by: https://api.github.com/users/knro
Status:   success
ID:       02204903

This is a strange place to be in. The build machine has to have the -devel package installed, or it will not build with the hardware support. Therefore, if anyone at all wants to use this or try this, the build machine has to have the -devel package. 

However, if the package is built with support for rtl-sdr, then it has to be a required dependency on anyone's distribution that installs INDI, right? Is something linked if it exists?

knro:
That's because I made it OPTIONAL and not REQUIRED. So if it doesn't find it, it skips the indi_rtlsdr driver completely.

xsnrg:
I guess my point is, for somebody that does not want to compile from source, unless the package is added to the spec, they will never have the option to use it. Am I misunderstanding?

1. build without rtl-sdr-devel
2. build is successful, package is published
3. user installs
In this case, they will not ever have the option to use INDI with rtl-sdr because the package they installed does not have support for it.

1. build with rtl-sdr-devel
2. build is successful, package that pulled in rtl-sdr is published
3. user installs
In this case, if a library is linked into the executable, there is a package dependency on rtl-sdr

Or do I have this all messed up, and now compile time does not change, and at runtime if it sees the library, it loads it if it is installed?

knro:
Correct. This is why I added librtlsdr-dev in Debian control. Since this is a standard Debian package, the dependency is OK in Launchpad. I don't think marking this as a required dependency is good for SOURCE, but for any distro that can do it from the get-go, then sure that would save users headaches in case they want to use this particular driver.

xsnrg:
Thank you. Will add it to the spec file.

PR #1465, opened 2021-05-24
OnStep: V1.10, high precision location, rotator/derotator, swap primary focuser, weather support, misc

james-lan:


drivers/telescope/lx200telescope.cpp
```diff
@@ -1380,28 +1380,54 @@ bool LX200Telescope::sendScopeLocation()
...
-    {
-        snprintf(lat_sexagesimal, MAXINDIFORMAT, "%02d:%02d:%02.1lf", lat_dd, lat_mm, lat_ssf);
-        f_scansexa(lat_sexagesimal, &(LocationNP.np[LOCATION_LATITUDE].value));
+    
+    
+    if (genericCapability & LX200_HAS_ALTERNATE_LOCATION_CMD) {
+        if (getSiteLatitudeAlt(PortFD, &lat_dd, &lat_mm, &lat_ssf, ":GtH#") < 0)
+        {
+            LOG_WARN("Failed to get site latitude from device.");
+            return false;
```
knro, line 1389: I think instead of adding LX200_HAS_ALTERNATE_LOCATION_CMD, wouldn't it be easier if sendScopeLocation/Time are virtual and can be overrided by the child class?

This way you don't need to add LX200_HAS_ALTERNATE_LOCATION_CMD and it would make the base class simpler and less prone to regressions.

knro (changes requested):
Need to implement the OnStep specific location changes in the OnStep class instead of letting parent class take care of it.

james-lan:
Going to virtual had me make a few mistakes, it looks like only the iopteron needed modification. Which was changing from void to book so shouldn't mess anything up. It looks like the failed test is failing for everything. I have left the additional call/wrapper in the lx200driver.

PR #1470, opened 2021-05-26
Manage Configured Features for Outputs

azwing:
When OnStep Auxiliary Features are configured, their names will be displayed in the Output Tab


azwing:
Will make more tests and retry later on

PR #1478, opened 2021-06-02
Fixing location accuracy and date&time for RST-135

jimmy-ck:
fixed:
- location accuracy is now received in arcsec resolution
- date & time & UTC offset are received properly now
- it should not interfere in any ways with classic LX200 mounts
- I renamed a lot of "title" comments because I found them misleading

not fixed:
- location in mount gets overwritten when Ekos is started from scratch

Ongoing discussion in: https://indilib.org/forum/ekos/9645-rst-135-location-issue.html?start=24#71910

### DIFF
--- a/drivers/telescope/lx200driver.cpp
+++ b/drivers/telescope/lx200driver.cpp
@@ -420,18 +420,18 @@ int getCalendarDate(int fd, char *date)
     }
     else
     {
-        LOGF_INFO("Date response length: %d", len);
+        LOGF_WARN("Date response length: %d", len);
         if (len == 12)
         {
             /* Rainbowastro format includes ":GC" so has to skip 3 characters */
             /* format is :GCMM/DD/YY# = 12 characters*/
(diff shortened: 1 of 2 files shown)


TallFurryMan:
I think that your computation of the location, by adding sexagesimal components, has a problem when d is negative. But I don't know if that can ever be the case.

knro:
Thanks! I don't think we should be changing lx200driver at all. Can you please just implement them in the rainbow driver? I did yesterday with sendScopeLoction, I believe they can be done for all the others leaving lx200 part unaffected by this change.

jimmy-ck:
@TallFurryMan Thanks for pointing that out. I tested only for positive coordinates – that’s true. Will test this later this das.

@knro Yes, should be possible. Will do this later this day as well.

jimmy-ck:
I'm closing this pull request and opening a new one with changes only in rainbow.cpp.

PR #1481, opened 2021-06-03
Fixing location accuracy and date&time for RST-135

jimmy-ck:
same as https://github.com/indilib/indi/pull/1478 but al changes are implemented directly into rainbow.cpp.
lx200driver.cpp has been reverted.

fixed:

- location accuracy is now received in arcsec resolution
- date & time & UTC offset are received properly now 
- I renamed a lot of "title" comments because I found them misleading

not fixed:

location in mount gets overwritten when Ekos is started from scratch
Ongoing discussion in: https://indilib.org/forum/ekos/9645-rst-135-location-issue.html

### DIFF
--- a/drivers/telescope/lx200driver.cpp
+++ b/drivers/telescope/lx200driver.cpp
@@ -420,18 +420,18 @@ int getCalendarDate(int fd, char *date)
     }
     else
     {
-        LOGF_INFO("Date response length: %d", len);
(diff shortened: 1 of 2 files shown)


knro:
Regarding the location on startup, INDI::Telescope reads the location from config file. But this happen before connecting to the mount. After connection sendScopeLocation() is called and THIS should actually read the location from the mount itself and send it over to the client.

PR #1487, opened 2021-06-13
In this commit, a crash at jpeg extraction is fixed.

Nekomeshi:
In case of YUV420 JPEG format, bitmap data are handled every 16 lines.
But when the image height is not multiple of 16, such as 1920x1080,
it accesses to out of 1080 line and it causes crashes.
So I added "yl < height" to limit the memory access to up to image height for Y plane.
And  I did the same things for U and V plane.


knro:
Thank you! I presume this was tested for different resolutions? no regressions detected there?

Nekomeshi:
I tested this code in indi_v4l2_ccd indi driver with following resolutions, and I don't see any errors.
These resolutions are what I can set on my webcam.
1920x1080
1280x720
640x480
640x360
320x240
320x180

PR #1491, opened 2021-06-19
fix rainbow astro focuser driver

murveit:
There were two issues fixed. (1) The doc from Rainbow Astro from which the original driver was written (https://github.com/indilib/indi/files/5457979/RSF_focuser_Protocol_200313_EN.pdf) mis-states the response to the focuser-current-position command, at least on the hardware that I've been testing. Instead of returning  :FPsDDDD# where s is the +/- and DDDD is a 4-digit position, it seems to return something like :FP-06.150# in other words a signed float that's between -8 and 8.0. I've modified the driver to accept either the original format or the floating point one.

Also, probably because of the marginal precision (3 decimal places) which exactly matches the possible position used, I observed off-by-one edge case. I corrected this by accepting off-by-one as "reached destination". Otherwise the driver hangs, because it thinks it's done.

I've also made all args required in the sendCommand() method, which simplifies the code, and I believe makes bugs a little less likely in the future.

### DIFF
--- a/drivers/focuser/rainbowRSF.cpp
+++ b/drivers/focuser/rainbowRSF.cpp
@@ -18,6 +18,30 @@
 
 */
 
+/*
+RSF Focuser protocol is documented here:
+https://github.com/indilib/indi/files/5457979/RSF_focuser_Protocol_200313_EN.pdf
+Currently the following commands are implemented.
+
+Command              Send      Receive     Comment
+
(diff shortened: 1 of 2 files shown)


knro:
Thanks! The sendCommand function is versatile and can be adapted to multiple drivers, but I suppose no issues of simplifying it if it works for this driver.

PR #1492, opened 2021-06-19
Fix OTA azimuth calculation for dome slaving

jpaana:
Fix OTA azimuth calculation after change from libnova to libastro as EquatorialToHorizontal expects RA in hours and does the *15 internally for libnova, where as previously RA was stored in degrees in dome class and given to libnova ln_get_hrz_from_equ function directly. This caused double conversion for RA and wrong OTA and dome azimuth result.


knro:
Good catch!! I can't believe I missed this. Thanks!

PR #1495, opened 2021-06-22
PMC8 Driver v0.4 (Improved manual slewing, guiding, and other features)

karlrees:
**Improved Manual Slews**
- Can move in both directions at the same time
- Increased max slew rate from 256x to 833x
- Changed move rate presets to 4x, 8x, 16x, 32x, 64x, 128x, 256x, 512x, and 833x
- Moves ramp the motors up exponentially instead of jumping to new speed all at once
- Added configurable ramp-up/down settings: step interval, base step size, step increase factor

**Improved Guiding Performance**
- Use precision tracking rate for RA guiding
- Separate guide rates (SRF) for RA/DEC
- Sync guide rates with mount
- Removed at least 80-100ms delay when initiating pulses by using previously collected tracking state information rather than collecting at the start of each pulse
- Block unnecessary calls to mount while guiding to improve responsiveness
- Reversed order of DEC commands to eliminate potential 20ms period where we could be guiding in wrong direction

**Other Changes**
- Autodetect tracking changes (e.g. initiated by concurrently connected non-Indi clients)
- Added setting to specify what type of serial cable is being used (default is auto-detect; select "Standard" to improve connection speeds with standard cables).
- Fixed bug where lunar/solar tracking rates were juxtaposed
- Experimental Southern Hemisphere support (need testers)
- Prevent issuing goto command while in middle of pulse (which would stop goto)
- Prevent issuing goto command while already in goto (to avoid problems with tracking resumption in latest firmware)
- Increased maximum precision tracking rate to just under 4x sidereal
- Fixed abort handling during manual moves, parking, and other contexts
- Various refactoring and code cleanup


knro:
That's quite an upgrade! Thank you Karl, the users would love these changes.

karlrees:
Hopefully not so many changes I messed stuff up, but I did test for a good while.

I need to make some changes to the PMC8 driver page on the website, but I don't have access.  Can I get that?

PR #1496, opened 2021-06-22
Pegasus Add new Device  USB Control HUB

chrysikos:


drivers/auxiliary/pegasus_uch.cpp
```diff
@@ -0,0 +1,525 @@
+#include "pegasus_uch.h"
```
knro, line 1: Please include the copyright headers.

drivers/auxiliary/pegasus_uch.cpp
```diff
@@ -0,0 +1,525 @@
...
+#include <memory>
+#include <regex>
+#include <termios.h>
+#include <cstring>
+#include <sys/ioctl.h>
+#include <chrono>
+#include <math.h>
+#include <iomanip>
+
+static std::unique_ptr<PegasusUCH> upb(new PegasusUCH());
```
knro, line 14: This is no longer required.

drivers/auxiliary/pegasus_uch.cpp
```diff
@@ -0,0 +1,525 @@
...
+
+    }
+    return INDI::DefaultDevice::ISNewNumber(dev, name, values, names, n);
+}
+
+
+//////////////////////////////////////////////////////////////////////
+///
+//////////////////////////////////////////////////////////////////////
+bool PegasusUCH::ISNewText(const char * dev, const char * name, char * texts[], char * names[], int n)
```
knro, line 260: No need to implement this function if it is not doing anything. Unless you plan to add stuff here in the future?

drivers/auxiliary/pegasus_uch.h
```diff
@@ -0,0 +1,111 @@
+#pragma once
```
knro, line 1: Copyright headers.

knro (changes requested):
Thank you. Minor changes required. Can you include screenshots for each tab? we need to create device documentation for this driver in INDI Devices hub.

PR #1498, opened 2021-06-25
CGX-L wasn't supported for last align command

florian-steffen:
Due to a buggy comparison, only the CGX mount was identified as supporting the last align command, not the CGX-L. I fixed the bug and refactored the code to add a feature flag in the low-level driver to check if the mount supports home indexes. The INDI driver then enables the last align command if the mount has home indexes.

PS: I followed the PR guide, and used artistic style to format the code. It introduced a lot of noise, I hope it is not too much a inconvenience.


knro:
Thanks a lot!!

PR #1505, opened 2021-06-29
Use getaddrinfo instead of gethostbyname (part 2)

paolostivanin:
Last time I replaced `gethostbyname` with `getaddrinfo` it seems that I forgot a couple of them.
This PR now concludes the porting from the deprecated `gethostbyname` to `getaddrinfo`.

Code has been compiled on both openSUSE Leap 15.3 and openSUSE Tumbleweed for both x86_64 and aarch64: https://build.opensuse.org/package/show/home:polslinux:teststuff/indi-git

I'll test it with my hardware later on.

### DIFF
--- a/libs/indibase/baseclient.cpp
+++ b/libs/indibase/baseclient.cpp
@@ -151,27 +151,31 @@ bool BaseClientPrivate::connect()
     }
 #endif
 
-    struct timeval ts;
+    struct timeval ts = {};
     ts.tv_sec  = timeout_sec;
     ts.tv_usec = timeout_us;
 
-    struct sockaddr_in serv_addr;
-    struct hostent *hp;
-    int ret = 0;
+    struct addrinfo *result = {};
+    struct addrinfo *rp = {};
+    struct addrinfo hints = {};
+    hints.ai_family = AF_UNSPEC;
+    hints.ai_socktype = SOCK_STREAM;
 
     /* lookup host address */
-    hp = gethostbyname(cServer.c_str());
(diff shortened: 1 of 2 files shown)


knro:
Thank you!

knro:
Actually, just tested with KStars and line 226 of baseclient.cpp ::connect **always** fails. Please check what's missing.

knro:
Appears it's always producing Connection Refused error on line 261. This is a major regression. We should revert if no solution is found soon.

paolostivanin:
Sorry, my bad :( I should have used the WIP tag on the PR. Last time I tested the changes on my hardware, but yesterday I couldn't do it.
I'll debug the issue in the next days, thanks Jasem!

paolostivanin:
Hi @knro ! I've finally had some time to test the patch. Equipment used:
- iOptron CEM70
- Nikon D7500

The software used was:
- openSUSE both stable (15.3) and rolling (tumbleweed) on x86_64
- indi 1.9.1 with my patch applied (https://build.opensuse.org/package/show/home:polslinux:teststuff/libindi-git)
- kstars 3.5.3

I was able to connect and disconnect the equipment multiple times, and I never had a failure. I was also able to slew, goto, take photos, etc.

Are you able to tell me what you do to make it fail? Maybe I'm able to reproduce it too.

knro:
On Ubuntu, connection to "localhost" apparently isn't recognized. Users complained that INDI Web Manager suddenly stopped working, I think due to the same as above.

PR #1509, opened 2021-07-07
update to 1.9.2 indi dev version in spec file

xsnrg:


xsnrg:
This changes the version of the RPM package and can be done by anyone any time a release is created to update to the latest (next) dev version

PR #1510, opened 2021-07-10
Support stacking mode on a indi_v4l2_ccd driver with v4l2 device that doesn't have AbsExposure.

Nekomeshi:
Action of stacking mode has been changed on #1325 so that the total storage time becomes close to specified exposure time. But it does not work for a device that does not have AbsExposure command such as HDMI capture board because actual exposure time for one frame is unknown.

In this commit, frame interval is used as the exposure time to support stacking mode.
There are two types of device about how to get frame interval. One is to choose frame rate from discrete values and the other is to get absolute frame rate value. This commit supports  former only because I don't have later device and cannot test it.


knro (changes requested):
Please remove the skywatcher file from the PR

Nekomeshi:
I merged latest master branch and pushed. Please check it.

PR #1511, opened 2021-07-12
Move calling of (dis)connect callback out of lock scope

jpaana:
Move calling of parent->serverConnected/serverDisconnected out of scope of the socket lock to prevent deadlock if the callback in turn calls socket locked functions (possibly from different thread). This happens with PHD2 for example and causes timeout when connecting to devices.


knro:
Thanks, looks good and tested with Ekos without issues as well.

PR #1515, opened 2021-07-21
The driver for the Planewave Delta-T was not working

nicholasdahlke:
The driver for the Planewave Delta-T was not working, it always displayed "Invalid Checksum", even tho the checksum was correct. Changing the type of the checksum from `uint8_t` to `char` fixed everything, and the driver is now working properly


jpaana:
Using plain char as a type for anything else than ASCII text is prone to problems as C standard doesn't specify whether it is signed or unsigned and that depends on the architecture. It's unsigned on ARM and signed on x86/x64. So strongly recommended to use (u)int8_t, (un)signed char or anything that specifies signedness explicitly.

knro:
Good point, int8_t should probably be used.

nicholasdahlke:
Thanks for that suggestion, I have not that much experience with C++, so I thought it was a good solution to my problem, I will try changing the type to int8_t.

PR #1521, opened 2021-07-31
Fix invalid substitution in udev rules file

Xiretza:
udevd complained with the following:

```
Invalid value "/bin/sh -c 'test -f /sys/module/usbcore/parameters/usbfs_memory_mb && test $(cat /sys/module/usbcore/parameters/usbfs_memory_mb) -lt 256 && echo 256 > /sys/module/usbcore/parameters/usbfs_memory_mb'" for RUN (char 76: invalid substitution type), ignoring.
```

The command substitution was interpreted as a udev substitution, it can be escaped with a second dollar sign.


knro:
@xsnrg Can you please double check? The rule is working fine on Ubuntu.

@Xiretza on what distro was this reported?

Xiretza:
> @Xiretza on what distro was this reported?

This was reported on Arch Linux's arch-general mailing list today: https://lists.archlinux.org/pipermail/arch-general/2021-July/049130.html

knro:
This works fine on 20.04 so accepted now, but hopefully @xsnrg can verify this on Fedora as well.

PR #1523, opened 2021-08-10
Update wrong coefficients to actual firmware spec

mworion:
For some reason the actual firmware might change the spec. I always saw wrong measurements un UPBv2.
Please keep in mind that dew_c has another coefficient explicitly mentioned in spec.


knro:
@chrysikos Can you please take a look? should we check for firmware version and then apply the rule per that?

PR #1524, opened 2021-08-13
indicom: fix FreeBSD compilation

andkem:
Adaption of FreeBSD ports tree patch.

Original-Author: Jose Alonso Cardenas Marquez <acm@FreeBSD.org>


knro:
Thanks!

PR #1527, opened 2021-08-27
Update lx200_OnStep.cpp

jamiecflinn:
Fix for flip unstable slewing, MF and target - This is related to 3 issues all stemming from one area of code - slew state was correctly being set, however it was immediately set to tracking - a side effect of this was target of slew was showing incorrectly AND flip was not succeeding/ending prematurely or having align start while the mount was actually slewing - I have run this through a solid test of east and west targets and parks - could not get get to MF as storms have blown in but this issues were manifest by basic slew so there should be no issue - This is my first push and using Github Desktop - let me know if all is ok or I have killed something

### DIFF
--- a/drivers/telescope/lx200_OnStep.cpp
+++ b/drivers/telescope/lx200_OnStep.cpp
@@ -1845,7 +1845,12 @@ bool LX200_OnStep::ReadScopeStatus()
             }
             if (strstr(OSStat, "n") && !strstr(OSStat, "N"))
             {
-                IUSaveText(&OnstepStat[1], "Slewing");
+                IUSaveText(&OnstepStat[1], "Slewing"); //Onstep 3.x
+                TrackState = SCOPE_SLEWING;
+            }
+            if (!strstr(OSStat, "n") && !strstr(OSStat, "N"))
+            {
+                IUSaveText(&OnstepStat[1], "Slewing"); //OnStep 4.x
                 TrackState = SCOPE_SLEWING;
             }
             if (strstr(OSStat, "N") && !strstr(OSStat, "n"))
@@ -1911,7 +1916,13 @@ bool LX200_OnStep::ReadScopeStatus()
                 }
                 else if (strstr(OSStat, "n") && !strstr(OSStat, "N")) 
                 {
-                    //set from state above 
+                    //set from state above  - OnStep 3.x
+                    TrackState = SCOPE_SLEWING; 
+                    IUSaveText(&OnstepStat[1], "Slewing");
+                }
+                else if (!strstr(OSStat, "n") && !strstr(OSStat, "N")) 
+                {
+                    //set from state above  - OnStep 4.x
                     TrackState = SCOPE_SLEWING; 
                     IUSaveText(&OnstepStat[1], "Slewing");
                 }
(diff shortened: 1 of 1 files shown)


azwing:
Personally no issue about the code but would suggest to wait until Jame's work is ready since ti has changes about same issue.

jamiecflinn:
Made a change in the code that was the subject of this request - found that v 4.x of OnStep changes the pattern of markins for what is considered SLEWING - so the new version now has a second check (!n and !N) - all else tested with my 3.16 branch and Khalid tested on 4.x - all seems good - New to github here so does the NEW commit need to be re-issued a pull request or can you see this code in what you access...let me know

if (strstr(OSStat, "n") && !strstr(OSStat, "N"))
            {
                IUSaveText(&OnstepStat[1], "Slewing"); //Onstep 3.x
                TrackState = SCOPE_SLEWING;
            }
            if (!strstr(OSStat, "n") && !strstr(OSStat, "N"))
            {
                IUSaveText(&OnstepStat[1], "Slewing"); //OnStep 4.x
                TrackState = SCOPE_SLEWING;
            }

jamiecflinn:
Final update via Khalid - some control characters (windows vs linux) and he moved the comments to better locations - core is the same - this has been built and verified to operate same as prevoius

kbahey:
I did test this code with OnStep 4.x. 
Before this patch, the following were broken, even for those who do not use meridian flips:

- Right after a slew is started, but before it ends, KStars would popup a notification that mount has reached target.
- Also after a slew is started, but before it ends, the "Eq. Coordinates" in INDI would change to yellow then immediately back to green, despite the mount is still moving. 
- The Ekos tab shows the name of the previous star as a target, not the last one. 

After this patch is applied, OnStep is back to normal with Ekos and INDI. 

So +1, it is urgently needed. We also need a package as soon as it gets in.

azwing:
OK for implementation. Thanks Khalid for testing. I just emerge from a brief but heavy down due to the second vaccine shot :-)  so sorry for the late answer.

kbahey:
@azwing 
This bug has messed up a lot of things, and we need a solution to go into the packages very soon.

The github issue says "pending reviewer". You are one of them. Can you mark it as reviewed?

@james-lan Can you please approve this? It has been tested on OnStep 3.x and 4.x, so good to go.

azwing (approved):

azwing:
> Thank you! I've asked @james-lan and @azwing to take a look to confirm this is good to go.

Jasem,

It seems James is not available for some time (I have no news since some time) . Can we go ahead without his approval?

james-lan:
This is fine since it addresses one case in v3/v4 reporting it's fine. 

I did test it earlier, (simulation cc'd & actual OnStep board, running v3, v4, v5 (didn't try X) in eq and altaz) and it does *not* solve the issue which has been quite annoying, which Khalid refers to. (Reproduced on the 2nd or 3rd ish go-to with master + this) 

As far as meridian flips, I did discover an issue with altaz mounts trying to flip.  altaz being commanded from ekos if on the HA > ... is checked. V3 and v5 altaz also have this. (Something I had not seen before.) 

At the moment unchecking that box seems to be the only fix, the driver isn't reporting something (pier side capability I suspect) correctly to ekos.

So based on what I've seen, commit this, as it resolves one big, but it is nothing special that will resolve the big bug. 

If I get a chance while I have some time (limited for a bit) I'll update the own-goto branch (which tries to and does per tests resolve the big bug) and send a pull request for that and see about fixing the extra meridian flips on altaz. It will overwrite this, because I changed the structure setting track state fair bit.

PR #1529, opened 2021-09-02
Script: Developer build script for UNIX users

pedramardakani:
This script is meant to build INDI inside computer RAM shared memory,
i.e. /dev/shm/indi-build, to reduce wear and tear stress on HDD/SDD.

It also aborts if there is less than 1GB space left on their RAM,
since INDI build directory could take 300MB memory.

### DIFF
--- a/developer-build.bash
+++ b/developer-build.bash
@@ -1,0 +1,76 @@
+#!/bin/bash
+
+# Developer Build Script
+#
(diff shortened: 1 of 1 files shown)


jochym:
Just tested it. It is a nice and simple tool. Speeds up the compilation, saves write cycles on SSD and is completely nonintrusive. I vote for merging.

PR #1532, opened 2021-09-10
Fix compilation on MacOS

pchev:
Compilation fail on Mac with the following error, this is fixed by the proposed change.

/Users/pch/src/indi-mac/craft/download/git/libs/indiserver/drivers/rotator/deepskydad_fr1.cpp:280:9: error: call to 'abs' is ambiguous
    if (std::abs(motorPositionDouble - GotoRotatorN[0].value) > 0.01 || GotoRotatorNP.s != motionState)
        ^~~~~~~~
/Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/usr/include/stdlib.h:132:6: note: candidate function
int      abs(int) __pure2;
         ^
/Library/Developer/CommandLineTools/usr/include/c++/v1/stdlib.h:111:44: note: candidate function
inline _LIBCPP_INLINE_VISIBILITY long      abs(     long __x) _NOEXCEPT {return  labs(__x);}
                                           ^
/Library/Developer/CommandLineTools/usr/include/c++/v1/stdlib.h:113:44: note: candidate function
inline _LIBCPP_INLINE_VISIBILITY long long abs(long long __x) _NOEXCEPT {return llabs(__x);}
                                           ^
1 error generated.


knro:
What's wrong with std::abs? maybe we're missing some include?

pchev:
From the messages it find only integer std::abs(), I don't know enough about clang to be sure why.

But I remark that fabs() is used everywhere except this one:

drivers/focuser/deepskydad_af1.cpp:        if (fabs(lastPos - FocusAbsPosN[0].value) > 5)
drivers/focuser/deepskydad_af3.cpp:        if (fabs(lastPos - FocusAbsPosN[0].value) > 5)
drivers/focuser/deepskydad_af3.cpp:        if (fabs(lastTemperature - TemperatureN[0].value) >= 0.1 ) //more accurate update
drivers/focuser/deepskydad_af2.cpp:        if (fabs(lastPos - FocusAbsPosN[0].value) > 5)
drivers/focuser/deepskydad_af2.cpp:        if (fabs(lastTemperature - TemperatureN[0].value) >= 0.5)
drivers/rotator/deepskydad_fr1.cpp:    if (std::abs(motorPositionDouble - GotoRotatorN[0].value) > 0.01 || GotoRotatorNP.s != motionState)

knro:
I changed the rest to std::abs, and they compile fine on MacOS here.

knro:
Please double check.

pchev:
Yes, probably something with my install. 
I see there is other std::abs() for double values in other driver that compile fine.
I check the difference in the include, there is probably a dumb .h somewhere in my system.

PR #1533, opened 2021-09-11
Developer build script

pedramardakani:
1. Update the script so it can take a directory other than the default `/dev/shm/indi-build`.
2. Update the script so it will check if the target build directory sits in `/dev/shm` and abort if memory is less than 1Gb.
2. Update the README.md file with instructions and explanations on using the `developer-build.bash` script.
3. Update the README.md file  pre-requisits section with packages in sync with the indilib website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,32 @@ Learn more about INDI:
 On Debian/Ubuntu:
 
 ```bash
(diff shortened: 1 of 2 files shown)


knro:
Bravo! I'm going to start using this more often!

PR #1537, opened 2021-09-22
Script: Developer build update

pedramardakani:
This version of the bash script takes three arguments:
1. `-j` will specify number of `x` jobs in `make -jx`, uses the maximum cores available by default.
2. `-o` will specify path to the target build directory
3. `-f` will skip checking for free memory available

The new `-j` argument is necessary because the `make -j` will freeze virtual machines. To determine the os type, I have used the output from `uname` function. The output starts with `Linux` for GNU/Linux systems, and with `Darwin` for macOS. The script will use `-j1` for other types of systems by default, unless specified otherwise by the user.

The new `-f` argument is necessary because there might be enough free RAM in `docker` containers, but the `/dev/shm/` directory can be limited. Mine was limited to a mere `64MB`. I used `--shm-size=1g` to increase the `/dev/shm/` size in my container. However, the script should have anticipated that and warned me *before* throwing the error complaining *no more space left on disk*.

Also, checking for free space shouldn't be limited only to RAM and `/dev/shm/`. So, the free space available in the specified target build will be determined using the `df` command, and warn the user unless the `-f` option is passed.

The new `-o` option is there so the user can pass the target directory without worrying about argument position in command line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ sudo make install
 
 ## Build indi-core (script)
 
-You can use the `developer-build.bash` script for faster building.
+**Alternatively**, you can use the `developer-build.bash` script for faster build and less stress on your SSD or HDD.
 
 ```bash
 cd ~/Projects/indi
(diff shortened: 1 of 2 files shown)


knro:
This looks good, thanks for improving the script! @xsnrg Any comments from your side?

xsnrg:
looks good to me, and still works well with defaults.

PR #1538, opened 2021-09-25
Pipes should be closed with pclose(), not fclose()

jpaana:
GCC 11 warns about this with -Wmismatched-dealloc.


knro:
Good catch!

PR #1539, opened 2021-09-28
Force using hour angle when mount is slewing

nouspiro:
Found problem that dome doesn't rotate until meridian flip is complete.
This is because mount report old pier side when it start meridian flip.
Before 6550c8b04 it used HA to figure pier side when taking this goto
path.


knro:
@fcasarramona Any issues with the MR?

PR #1543, opened 2021-10-06
OnStep: Update to 1.12

james-lan:
This includes 
*Speedups on detection for optional things (focusers for example) which cut initialization time down a fair bit. (Something like 15+ seconds sometimes down to 3 seconds for V3, Can still hangup on a few things from the LX200 drivers, like the initial ack for 5 sec, but after that takes 3 sec for all versions)
*All updates to position and tracking setup are done by OnStep, which should resolve the long standing slew done immediately issue. 
* Human readable TMC Driver messages
* PEC and PierSide capabilities not added to the driver unless detected. (Prevents some Meridian Flips on mounts that don't need them.) 
* Detects OnStep or OnStepX version (doesn't do much with it.) 
* Behind the scenes: command strings handled in a manner which should prevent some future issues that have happened due to short strings/string copies

NOTE @knro : Before merging this please check it, as this adds functions to indicom allowing <1 sec timeouts (per the timeval type (already used with seconds in indicom)) and changes some variable scopes in inditelescope.h (no changes to inditelescope.cpp though) It shouldn't break anything and hasn't in my testing, but I don't have a lot of other hardware to test with.


azwing:
no issues observed during my tests. Thnak you for the work

PR #1546, opened 2021-10-14
usbfs_memory_mb_fix: exit with TRUE even if the memory is already set

xsnrg:
… at or above 256M to prevent logging an error.

Tested by resetting memory below 256M, plugging and unplugging a USB device repeatedly. Then leaving at 256M, and plugging and unplugging repeatedly. The error is gone and the memory is still set correctly.


xsnrg:
with a nod to @Safari77

PR #1549, opened 2021-10-28
LX200: Fix setSiteName/getSiteName crash, solution 1

james-lan:
For alternate, Solution 2, see: https://indilib.org/forum/development/1406-driver-onstep-lx200-like-for-indi.html?start=732#76974


knro:
I submitted a different fix here: https://github.com/indilib/indi/commit/4018919a2298cc683092588e3a9acf68fbd082d0

Please test and let me know if it works OK.

PR #1550, opened 2021-10-30
LX200 SiteName Fix V2

james-lan:
This should fix it for all the places that IUSaveText wasn't used in relation to SiteName, and also fix the pulsar2 driver to use IUSaveText. (I can't test that one.) 

I am away from home, but I am not 100% able to test on OnStep.

### DIFF
--- a/drivers/telescope/lx200telescope.cpp
+++ b/drivers/telescope/lx200telescope.cpp
@@ -121,7 +121,7 @@ bool LX200Telescope::initProperties()
 
     char siteName[64] = {"NA"};
     IUGetConfigText(getDeviceName(), "Site Name", "Name", siteName, 64);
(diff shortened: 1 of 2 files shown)


drivers/telescope/lx200telescope.cpp
```diff
@@ -120,9 +121,10 @@ bool LX200Telescope::initProperties()
 
     char siteName[64] = {"NA"};
     IUGetConfigText(getDeviceName(), "Site Name", "Name", siteName, 64);
-    IUFillText(&SiteNameT[0], "Name", "Name", siteName);
+    IUFillText(&SiteNameT[0], "SiteName", "SiteName", siteName);
```
knro, line 124: I don't think we need to change this. It would break users existing config files. It also doesn't really add any benefit, "Name" works OK.

knro (approved):

PR #1551, opened 2021-11-02
Fix FocusLynx GetConfig for focusers without HomeOnStart

DanielVanNoord:
The END from devices that did not support HomeOnStart would contain a '\10' or other characters from the last response on the end, causing them to fail the strcmp check for a match with END. When they failed the driver would timeout trying to read the next line (which would never come). Other end checks (~line 1062 as an example) would set the last byte to '\0' so they would pass. In this case the response may be used for processing so instead of setting the last character as a '\0' this checks the first 3 characters as a response key will never start with END .

This was tested on Ubuntu 20.04 with a ThirdLynx and FocusLynx controller. Both controllers were tested with firmware from before and after the addition of HomeOnStart being an option.

### DIFF
--- a/drivers/focuser/focuslynxbase.cpp
+++ b/drivers/focuser/focuslynxbase.cpp
@@ -1029,7 +1029,7 @@ bool FocusLynxBase::getFocusConfig()
         return false;
     }
     // If END, then ignore
-    else if (strcmp(response, "END"))
+    else if (strncmp(response, "END", 3))
     {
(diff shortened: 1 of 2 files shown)


knro:
I swear we had a PR before for this exact issue!

DanielVanNoord:
I did not see one so I wrote this pull request. I replicated the issue on both the current release and the master branch.

PR #1552, opened 2021-11-03
Updates and cleanup for the OpenWeatherMap driver

jpaana:
Updates and cleanup for the OpenWeatherMap driver:
- use new property api
- report curl errors
- fix labels and reset values for rain & snow
- clean up json parsing by moving repeated code to lambda function
- run astyle for header


knro:
Wow thanks!! That's a *lot* cleaner!

PR #1558, opened 2021-11-08
Improve lynx sync and homing behavior

### LINKED ISSUE #1554: FocusLynx Syncing current issues and proposed improvement
With further testing on the FocusLynx driver I discovered odd behavior involving syncing. For many device types the driver forces the focuser to sync before it can be moved. Although this can be turned off (seemingly only for a single session do to an issue that I believe occurs in checkIfAbsoluteFocuser()) with the Sync Mandatory option I am unsure if this requirement it is needed. All Optec Lynx devices are absolute, they operate in steps and retain the position across power cycles and connections.

I suspect this issue arose over differences between Optec FocusLynx Homing and Syncing focusers. Homing focusers automatically move on powerup to calibrate the range and position. These focusers cannot sync. Syncing focusers allow the user to sync the position. However, once synced the focuser will retain the position unless the user manually disengages the focuser and manually moves it. At that point it is recommended that the user resync the focuser so the physical position matches the controllers position but this is considered optional.

Currently the “Absolute” check in checkIfAbsoluteFocuser() is checking for a mix of Homing and Syncing absolute focusers. This can be updated to be either Homing or Syncing with both being absolute and only Syncing supporting optional Sync commands.  

Proposed changes:
1.	Update list of homing vs syncing devices types to use the Device ID code for homing focusers. This can be read on connection to determine if the controller supports homing or syncing for the current focuser.
2.	Make syncing optional by default or remove the sync mandatory functions completely. As all focusers are absolute syncing should be optional for all that support it.
3.	Disable syncing for homing focusers. Currently allowing it should be harmless, the controller will return an error if these are synced and ignore the request. But is seems good to remove something that has no function.

I can work on developing and testing a pull request if you like the proposed changes. But I wanted some feedback to make sure that I am on the right track for how indi uses focusers. Do you have any thoughts?

DanielVanNoord:
Requesting review for the following changes to the Lynx driver

1. Removes the required sync before move from all Lynx focusers as they are all absolute
2. Use the controllers returned device ID to determine support for Home vs Sync
3. Disable syncing for homing focusers

This should fix #1554 

Please let me know if there are any required or requested changes, I am learning the layout and style for INDI drivers.

So far I have tested this on a FocusLynx hub with a DirectSync and TCF-S 2 as well as a ThirdLynx and a TCF-LSI.

### DIFF
--- a/drivers/focuser/focuslynxbase.cpp
+++ b/drivers/focuser/focuslynxbase.cpp
@@ -75,8 +75,7 @@ FocusLynxBase::FocusLynxBase()
                       FOCUSER_CAN_REVERSE  |
                       FOCUSER_HAS_BACKLASH);
 
-    isAbsolute = false;
-    isSynced   = false;
+    canHome = false;
     isHoming   = false;
 
     simStatus[STATUS_MOVING]   = ISS_OFF;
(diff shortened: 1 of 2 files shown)


knro:
Thank you for fixing these pending issues, the driver is certainly a lot simpler now.

Issue #1564, opened 2021-11-10
Add Star Alignment option to Rainbow Astro RST-135(E) mount driver


wjdrijfhout:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%24160%20Funded-%2300A156.svg" />](https://issuehunt.io/r/indilib/indi/issues/1564)

The RST-135(E) mount offers an option to build a star alignment model in-mount, based on 5 or more star alignment points. With a proper in-mount model, it is possible to use this mount for decent exposures of unguided imaging. The mount model also enhances GoTo results, which can be important for small/faint targets.
However, with the current driver, the 'sync' function only sends an RA/DEC error to the mount, and it is not interpreted as a star alignment point. If you want to build a mount model, you have to do that manually. 

What I would like to have is a checkbox  in the driver to turn on 'star alignment on sync'. When ticked, a regular sync command from Ekos would then be interpreted in the mount as star alignment point. As soon as the mount has received at least 5 star alignment points, it automatically calculates this model. If such an option is available, the existing mount model functionality in Ekos can be used to just define five (or more) random alignment points, press 'run', and the model builds itself in the mount. Once the model is built, the check-box can be unchecked again if desired.

The alternative is either a fully manual method (GoTo star, center star using handpad, sync in handpad, repeat five times) or a semi-automated method (GoTo star, 'Sync & Slew to target' in ekos, sync in handpad, repeat five times). These are both annoying and time-consuming options.

I am in contact with BJ from Rainbow Astro and he is willing to share the protocol for star alignment to be included in the INDI driver. The ASCOM driver has a similar feature built in already.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $160.00)

- [<img src='https://avatars.githubusercontent.com/u/63748804?v=4' alt='gordoabc' width=24 height=24> gordoabc](https://issuehunt.io/u/gordoabc) ($20.00)
- [<img src='https://avatars.githubusercontent.com/u/38950558?v=4' alt='wjdrijfhout' width=24 height=24> wjdrijfhout](https://issuehunt.io/u/wjdrijfhout) ($100.00)
- $40.00 have been anonymously funded.
### Submitted pull Requests
- [#1566 Save Alignment Data to mount before sync](https://issuehunt.io/r/indilib/indi/pull/1566)
[...]

PR #1566, opened 2021-11-14
Save Alignment Data to mount before sync

sywong2000:
Save Alignment Data to mount before sync.
Users can checked the "Save Star Alignment to mount before sync". This would send the star alignment command before every sync operations. Then use the "Mount Model Tool", choose the "Sync" as the platesolve operation. Then at each sync, the alignment data will save to the mount.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#1564: Add Star Alignment option to Rainbow Astro RST-135(E) mount driver](https://issuehunt.io/repos/58881442/issues/1564)
---
</details>

### DIFF
--- a/drivers/telescope/rainbow.cpp
+++ b/drivers/telescope/rainbow.cpp
@@ -88,9 +88,9 @@ bool Rainbow::initProperties()
     IUFillNumberVector(&GuideRateNP, GuideRateN, 1, getDeviceName(), "GUIDE_RATE", "Guiding Rate", MOTION_TAB, IP_RW, 0,
                        IPS_IDLE);
 
-    // Test Function to send the StarAlign Command to the mount
-    IUFillSwitch(&SaveAlignAfterSlewS[0], "SEND_ALIGN_SLEW_COMPLETED", "Send Align After Slew Completed", ISS_OFF);
-    IUFillSwitchVector(&SaveAlignAfterSlewSP, SaveAlignAfterSlewS, 1, getDeviceName(), "SEND_ALIGN_SLEW_COMPLETED", "Send Align Command After Slew Completed",
+    // Align Star Before Sync
+    IUFillSwitch(&SaveAlignBeforeSyncS[0], "SEND_ALIGN_BEFORE_SYNC", "Save Star Alignment to mount before sync", ISS_OFF);
+    IUFillSwitchVector(&SaveAlignBeforeSyncSP, SaveAlignBeforeSyncS, 1, getDeviceName(), "SEND_ALIGN_BEFORE_SYNC", "Save Star Alignment to mount before sync",
                        ALIGNMENT_TAB, IP_RW, ISR_NOFMANY, 60, IPS_IDLE);
 
     setDriverInterface(getDriverInterface() | GUIDER_INTERFACE);
@@ -120,7 +120,7 @@ bool Rainbow::updateProperties()
         defineProperty(&GuideWENP);
         defineProperty(&GuideRateNP);
 
-        defineProperty(&SaveAlignAfterSlewSP);
+        defineProperty(&SaveAlignBeforeSyncSP);
 
     }
     else
@@ -131,7 +131,7 @@ bool Rainbow::updateProperties()
         deleteProperty(GuideNSNP.name);
         deleteProperty(GuideWENP.name);
         deleteProperty(GuideRateNP.name);
-        deleteProperty(SaveAlignAfterSlewSP.name);
+        deleteProperty(SaveAlignBeforeSyncSP.name);
     }
 
     return true;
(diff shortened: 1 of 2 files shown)


drivers/telescope/rainbow.cpp
```diff
@@ -230,6 +241,17 @@ bool Rainbow::ISNewSwitch(const char *dev, const char *name, ISState *states, ch
             IDSetSwitch(&HomeSP, nullptr);
             return true;
         }
+        else if (!strcmp(SaveAlignBeforeSyncSP.name, name)) {
+            if (SaveAlignBeforeSyncS[0].s == ISS_ON)
```
knro, line 245: No need to do this, just do IUUpdateSwitch(...) then SaveAlignBeforeSyncSP.s = IPS_OK, then setIDSetSwitch(&Save...)

drivers/telescope/rainbow.h
```diff
@@ -160,6 +176,10 @@ class Rainbow : public INDI::Telescope, public INDI::GuiderInterface
         const std::string getSlewErrorString(uint8_t code);
         uint8_t m_SlewErrorCode {0};
 
+        ISwitchVectorProperty SaveAlignBeforeSyncSP;
+        ISwitch SaveAlignBeforeSyncS[1];
```
knro, line 180: I think it's better to make this SaveAlignBeforeSyncS[2] and use INDI_ENABLED and INDI_DISABLED switches. I understand you want to use it as a toggle (which works), but using two switches (Enabled and Disabled) is more *clear* to the end user.

Finally, try to add this switch to the saveConfig(...).

drivers/telescope/rainbow.cpp
```diff
@@ -922,8 +947,17 @@ bool Rainbow::Abort()
...
-    char cmd[DRIVER_LEN] = {0};
 
+    if (SaveAlignBeforeSyncS[0].s == ISS_ON)
+    {
+           StarAlign(ra, dec);
+    }
+    else
+    {
+        LOG_DEBUG("SaveAlignBeforeSyncS is OFF. Not doing anything.");
+    }
```
knro, line 958: so starAlign and regular sync can be run after each other? or ONLY one should be used at a time?

knro (changes requested):
Few changes and clarifications.

drivers/telescope/rainbow.cpp
sywong2000: Thanks I have made the changes to my latest push

drivers/telescope/rainbow.h
sywong2000: Yes definitely better to align to the common UX, I've updated this also

drivers/telescope/rainbow.cpp
sywong2000: yes you're definitely correct, Chris also mentioned this so I changed the command and give the status printout updated a little bit

knro (commented):
THanks, looks good but I see many indention and style changes. Are you using **astyle** from INDI?

sywong2000:
I used Qt Creator and I must have hit the Ctrl-I and changed the indents before I commit.. sorry about that.... meanwhile Chris has submitted a version which is very similar. If mine having too much code change it's okay to reject my version and use his version instead. Thank you.

knro:
I don't see any other version submitted?

sywong2000:
Sorry I don't have access to my latest source now, I'm going to submit another version with the astyle later tonight (I'm in GMT +8:00) I'll re-request the review once I've pushed the latest version. Thanks.

jimmy-ck:
@sywong2000: Sorry, I was very busy in the last couple of days. If needed I can create a new PR with my solution. How shall we proceed?

sywong2000:
I just revert a few versions and commit again a more condensed code hopefully this is much better. Thank you.

sywong2000:
> @sywong2000: Sorry, I was very busy in the last couple of days. If needed I can create a new PR with my solution. How shall we proceed?

@jimmy-ck sure, sorry I just noticed this and I has commit another version to revert the unnecessary indents. Please feel free to go ahead if you want to create another PR and overwrite my changes. Thank you.

jimmy-ck:
@sywong2000 Your latest commit looks good. No need for anything else. :)

jimmy-ck (approved):
Reviewed (not tested) & looks good.

knro:
Thanks! Can this be tested on the actual mount?

sywong2000:
Yes I tested it with my RST-135E before I pushed the code.

knro:
Thanks, let's ask end users now to test from the nightly PPA.

jimmy-ck:
Oh, I think we messed something up here. :(
I completely overlooked that the new property does not get initialised and updated, therefore is not visible in INDI Control Panel. @sywong2000 Did you really compile and test the latest commit on your mount? This not not work. I'll commit a fix.

sywong2000:
@jimmy-ck My fault - I'm terribly sorry I must have messed up the versions when I fix the code indent problem - thanks for catching this and applying the fix.

PR #1570, opened 2021-11-15
More Lynx fixes - Step Size and Devices Types

DanielVanNoord:
This fixes an incorrect command doc that listed the step size conversion as 10000 instead of the correct 1000. Because this is informational only (not used within the controller for anything other then display) changing it to the correct value should have no consequences.

The second more major fix is a cleanup of the device types. This makes a few device type names generic, removes unused device types and merges a few duplicates. The exact details are in the commit with additional notes in the code.


knro:
Great work, thanks!

PR #1574, opened 2021-11-21
CCD drivers: snooping for standard epoch positions and issues with fast exposure mode fixed

stefansimg:
-  CCD drivers now also snoop for standard epoch positions (indi property EQUATORIAL_COORD)
- CCD drivers: conversion of epoch of day position (indi property EQUATORIAL_EOD_COORD) to standard epoch position (if no standard epoch position given) and calculation of ALT/AZ coordinates moved to addFITSKeywords such that they are updated during fast exposure
- CCD drivers: backup of exposure start time and duration before a new exposure is started in fast exposure mode. (Previously the fits keyword contained the exposure information of the next frame in fast exposure mode.)

### DIFF
--- a/libs/indibase/indiccd.cpp
+++ b/libs/indibase/indiccd.cpp
@@ -91,6 +91,7 @@ CCD::CCD()
     pierSide        = -1;
     J2000RA         = std::numeric_limits<double>::quiet_NaN();
     J2000DE         = std::numeric_limits<double>::quiet_NaN();
+    J2000Valid      = false;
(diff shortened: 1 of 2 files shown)


knro:
Thanks, but aren't all driver exposing EQUATORIAL_EOD_COORDS? perhaps maybe synscan isn't ??

stefansimg:
Not all drivers use indilib ... Since EQUATORIAL_COORDS is a standard property it should also be supported.

E.g. I developed a mount / telescope controller (which is not published yet) that uses property.

Issue #1580, opened 2021-12-23
auxiliary/pegasus_upb strange behaviour with USB_PORT_CONTROL switch vector


gnthibault:
**Describe the bug**
I am currently developing a small software to drive a telescope setup, and the UPBV2 is a central piece of my setup.
Unfortunately, I recently came across a weird behaviour with regard to the switchvector called "USB_PORT_CONTROL"

My current strategy for every switchvector, as a client, is to send the vector to the indiserver, and then wait for the corresponding light to be set to Ok (or eventually idle, but that's another story). This is somehow an acknowledgment mechanism, and the light is set to Ok, independantly of wether any or none of the switch value have actually been changed.

Anyway, this acknowledgment strategy is working pretty ok on most of the drivers I have been playing with recently, also, it is working on the "POWER_CONTROL" and "POWER_ON_BOOT" switchvector. However, when it comes to "USB_PORT_CONTROL", I noticed that the driver is never acknowledging the switch reception in the case that the sent switchvector has no difference with the current switch values.

Second issue:
In addition, I noticed that the switchvector named "USB_HUB_CONTROL" always exhibit an "Alert" value when used. I suppose there is an issue somewhere with this feature inside the indi driver, but I am not competent enough to debug this.

Is there anyway for someone to crosscheck if the behaviour can be reproduced, and if it is expected ?

Thank you in advance for your help

**To Reproduce**
Exact steps to reproduce the behavior.
1. Run indi_pegasus_upb driver
2. Set USB_PORT_CONTROL switch vector, with some values for PORT_1, PORT_2, PORT3, PORT_4
3. Try to reproduce the step 2, i.e with the exact same switch values
3. You should notice that no acknowledgement is never sent for the second call

**Expected behavior**
I expect the driver to send a "USB_PORT_CONTROL" switchvector with a light set to "Ok" independantly of wether the latest client command did change or didn't

**Screenshots**

**Desktop (please complete the following information):**
 - OS:  Linux rockpro64 5.15.7-1-MANJARO-ARM #1 SMP PREEMPT Wed Dec 8 13:17:00 UTC 2021 aarch64 GNU/Linux
 - Version Hash ce7416338a3f99888b21c2f284aa2033718c5e6f

**Log Files**
Edit removed useless logs

knro:
[...]

PR #1581, opened 2021-12-26
Trying to fix https://github.com/indilib/indi/issues/1580

gnthibault:
Trying to fix https://github.com/indilib/indi/issues/1580 I think there was some kind of copy/paste mistake, but I am not 100% sure. However, I tested this solution on my local install, and the behaviour is now the expected one.

### DIFF
--- a/drivers/auxiliary/pegasus_upb.cpp
+++ b/drivers/auxiliary/pegasus_upb.cpp
@@ -714,7 +714,8 @@ bool PegasusUPB::ISNewSwitch(const char * dev, const char * name, ISState * stat
                 USBControlV2SP.s = IPS_ALERT;
             }
 
-            IDSetSwitch(&USBControlSP, nullptr);
+            IDSetSwitch(&USBControlV2SP, nullptr);
+
             return true;
         }
 
--- a/drivers/auxiliary/pegasus_upb.cpp
+++ b/drivers/auxiliary/pegasus_upb.cpp
@@ -703,7 +703,7 @@ bool PegasusUPB::ISNewSwitch(const char * dev, const char * name, ISState * stat
             }
 
             // All is OK
-            if (rc[1] && rc[2] && rc[3] && rc[4] && rc[5])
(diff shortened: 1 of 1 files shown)


gnthibault:
Ok nope actually, problem is not solved, there is something else going on I think

gnthibault:
Ok I narrowed it down to that line:
https://github.com/indilib/indi/blob/ce7416338a3f99888b21c2f284aa2033718c5e6f/drivers/auxiliary/pegasus_upb.cpp#L701

Il looks like this check s=consistently returns False for rc[0] i.e the first usb port. I think I need to get feedback from someone from pegasus on this, it looks too specific

gnthibault:
Ok, going through the documentation, I finally saw that:

> USB
> Enable or disable the USB hub. This can only control ports 1-5 since port 6 is always on.
> 

From
https://www.indilib.org/devices/auxiliary/pegasus-ultimate-power-box.html

I think index 0 in that case relates to port 6 in the documentation

gnthibault:
Ok I think there is an additional error in the driver. In case you have a UPBV2, the usb from 1 to 6 can be enabled/disabled. However, one of the 
`setUSBPortEnabled(i, USBControlV2S[i].s == ISS_ON)`
Always return False (i.e, value 1), when i == 0
It is not clear for me why it is so ? However I absolutely need any possible call to setSwitch, to return a status of either Idle or Ok, to cope with the overall semantic of the drivers. I think there might be some semantic mismatch with what is done in upbv1 ? Maybe someone from Pegasus can help here.

gnthibault:
This might help, especially for the expected return values:
[Ultimate_Powerbox_v2_Serial_Command_Table.pdf](https://github.com/indilib/indi/files/7777500/Ultimate_Powerbox_v2_Serial_Command_Table.pdf)

gnthibault:
Ok, after a lot of tinkering, and updating my (now) rusty C++14 knowledges, there was an issue with implicit initialization of array. Commit to come soon. There were 2 bugs in a row, and I was kind of disturbed by this ^^

PR #1589, opened 2022-01-10
Add timeouts to rainbow focuser

murveit:
Increases the tolerance for reaching the focuser position target after waiting for a number of seconds. If the focuser has been busy for 60 seconds then finish no matter what.

### DIFF
--- a/drivers/focuser/rainbowRSF.h
+++ b/drivers/focuser/rainbowRSF.h
@@ -78,7 +78,8 @@ class RainbowRSF : public INDI::Focuser
         const static uint32_t homePosition { 8000 };
 
         // Timer used to timeout when waiting for requested movement to complete.
-        std::unique_ptr<INDI::ElapsedTimer> m_MovementTimer;
+        INDI::ElapsedTimer m_MovementTimer;
+        bool m_MovementTimerActive { false };
 
         /////////////////////////////////////////////////////////////////////////////
         /// Static Helper Values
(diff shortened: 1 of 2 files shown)


drivers/focuser/rainbowRSF.h
```diff
@@ -76,6 +76,8 @@ class RainbowRSF : public INDI::Focuser
 
         const static uint32_t homePosition { 8000 };
 
+        time_t timeoutStartSeconds = 0;
```
knro, line 79: Use INDI::ElapsedTimer, it's simpler to use and standard.

knro (changes requested):

drivers/focuser/rainbowRSF.h
murveit: Done. Thanks for the suggestion.

drivers/focuser/rainbowRSF.h
```diff
@@ -76,6 +77,8 @@ class RainbowRSF : public INDI::Focuser
 
         const static uint32_t homePosition { 8000 };
 
+        // Timer used to timeout when waiting for requested movement to complete.
+        std::unique_ptr<INDI::ElapsedTimer> m_MovementTimer;
```
knro, line 81: There is no need to use unique_tr, you can just declare it as INDI::ElapsedTimer m_MovementTimer. The class internally manages its own smart pointer.

knro (approved):

PR #1593, opened 2022-01-13
fix for bug 1359

alteholz:
macros like __BYTE_ORDER__ do not really work with gcc  (especially not on PPC64)

With this patch, the execution time increases about 25% on PPC64:
1: [ RUN      ] CORE_BASE64.Test_from64tobits_fast_time
1: [       OK ] CORE_BASE64.Test_from64tobits_fast_time (169 ms)
1: [ RUN      ] CORE_BASE64.Test_from64tobits_fast_with_bug_time
1: [       OK ] CORE_BASE64.Test_from64tobits_fast_with_bug_time (139 ms)

Whereas on amd64 the execution time increases about 10%:
1: [ RUN      ] CORE_BASE64.Test_from64tobits_fast_time
1: [       OK ] CORE_BASE64.Test_from64tobits_fast_time (79 ms)
1: [ RUN      ] CORE_BASE64.Test_from64tobits_fast_with_bug_time
1: [       OK ] CORE_BASE64.Test_from64tobits_fast_with_bug_time (72 ms)

(the second function from64tobits_fast_with_bug() is only needed to compare performance)

### DIFF
--- a/base64.c
+++ b/base64.c
@@ -40,6 +40,11 @@
 #include "base64_luts.h"
 #include <stdio.h>
 
+#include <byteswap.h>
+#include <arpa/inet.h>
+#define  IS_BIG_ENDIAN     (1 == htons(1))
+#define  IS_LITTLE_ENDIAN  (!IS_BIG_ENDIAN)
+
 /* convert inlen raw bytes at in to base64 string (NUL-terminated) at out. 
  * out size should be at least 4*inlen/3 + 4.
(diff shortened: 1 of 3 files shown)


knro:
@alteholz Looks like byteswap might not be available on MacOS?

alteholz:
@knro I copied the stuff from byteswap.h, so base64.c should compile on macos now.

Issue #1592, opened 2022-01-12
Celestron GPS driver isn't disabling tracking while slewing, causes trouble with CPC mounts.


bobrich:
**Describe the bug**
I have a Celestron CPC 1100 and find that if I slew the scope (typ using kstars mount control) with tracking enabled, the tracking will get interrupted for 5-10 seconds and throw off alignment.  

Celestron doc here suggests tracking should be disabled while slewing:

https://s3.amazonaws.com/celestron-site-support-files/support_files/1154108406_nexstarcommprot.pdf

> **Note:** in most configurations, issuing the slew commands will override (or conflict with) the tracking mode of the
> mount. Hence it is always best to disable tracking first using the **Tracking Commands**, issue the slew command, then
> re-enable tracking.

FWIW this appears to be largely addressed in 3rdparty indi_celestron_aux.  It looks like it uses TrackState = SCOPE_SLEWING plus some additional info (ScopeStatus and axis state) and doesn't appear to be affected by this issue.   Tracking isn't impacted by manual slewing.  

**To Reproduce**

1. Configure KStars to control a CPC1100 mount with indi_celestron_gps.
2. Align the scope with the hand controller
3. Use KStars mount control to slew left, right, up and down
4. Observe that tracking is partially or completely stalled for 5-10 seconds

**Expected behavior**
Slewing does not interfere with tracking.

PR #1595, opened 2022-01-15
Added manualSlewing flag to sustain tracking suppression during manual slew

bobrich:
For issue #1592 

May not be the most elegant approach but the Celestron GPS mount implementation needs to suppress tracking for both Goto and manual slewing activity.  The current mechanism relies on the mount to maintain the is_slewing state, but it only does this for Goto, not manual.

I can't find a specific command in the Nextar to check for motor activity so I just created a flag to track manual slew state changes.  This, in combination with the change from 221c10c22cd0d43a57e41375dee00246c31c2111 generally get it working.

There's still a small delay based on the loop tempo, but I couldn't get SetTrackingState(true) to work if I called it immediately after the move command stopped.

### DIFF
--- a/drivers/telescope/celestrongps.cpp
+++ b/drivers/telescope/celestrongps.cpp
@@ -653,6 +653,8 @@ bool CelestronGPS::MoveNS(INDI_DIR_NS dir, TelescopeMotionCommand command)
     switch (command)
     {
         case MOTION_START:
+            TrackState = SCOPE_SLEWING;
+            manualSlewing = true;
             if (driver.start_motion(move, rate) == false)
             {
                 LOG_ERROR("Error setting N/S motion direction.");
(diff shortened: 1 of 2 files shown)


knro:
Thanks, so should tracking be explicitly called off on manual slew and GOTO? i.e. setTrackEnabled(false) ?

bobrich:
My understanding is quite limited, based solely on the notes in the Celestron doc I linked in the issue.  To me it seems like a bug in their protocol that we need to work around.  I don't know if stopping tracking before the slew is necessary, I just think that when the Goto and manual slew commands come to the end of their action it also kills the tracking so we need to explicitly restart it.

I had to add another commit b/c my simplistic state tracking didn't work with multi-axis slewing.  I would greatly prefer a command to get current slewing state from the mount that also includes manual slewing, but I have been going back and forth with their support and haven't found one yet.  Thanks for taking a look, happy to test.

knro:
I already [added setTrackEnabled](https://github.com/indilib/indi/blob/5ed21bda64b6e04d43290efa2a606d261aaf0b6f/drivers/telescope/celestrongps.cpp#L825) explicitly after GOTO is done. Maybe a similar thing can be stopped for manual motion after Stop is executed? Perhaps try with that?

bobrich:
Sure, I'll tinker with that some more.  I've tried it once by just adding to the stop flow in MoveNS/EW and didn't have any effect, but I'm sure I'm just missing something.  Thanks!

PR #1596, opened 2022-01-16
Send command at start to flush serial line for Pulsar/Rigel dome

nouspiro:
When first started rigel dome would crash on unhandled exception thrown
from std::stoi() as it receive "E,1" because drive unit had some garbage
in input buffer.


knro:
Would `tcflush(PortFD, TCIOFLUSH);` work?

nouspiro:
I don't think so. This is about to flush any garbage that is received by controller on other side.

PR #1602, opened 2022-01-23
Pyindi - run actual tests

abecadel:
run actual pyindi tests, not just import the pyindi lib to verify that python binding is working


knro:
Looks like first actual test failed: https://github.com/indilib/indi/runs/4914088595?check_suite_focus=true

Any ideas why?

abecadel:
yes, we still have to fix https://github.com/indilib/pyindi-client/issues/4

PR #1609, opened 2022-01-29
Add and apply Meade Protocol spec definitions

d33psky:
Add Meade Protocol spec definitions, and apply:
Swap :Sh and :So
Fix :Sd, setCommandXYZ, :SC, :SG, :Sz, :SM, :SN, :SO, :SP, :ST, DEBUGFDEVICE logging
Apply astyle (I hope it automatically uses what's in the repo)
Document a few 10Micron extensions


knro:
Thanks! We need some tests to confirm this works OK on LX200. Any feedback?

PR #1620, opened 2022-02-08
Updated version of the AstroPhysics driver--for testing

murveit:
I've tested this A-P driver with Ekos automation, and now its ready for further testing with other versions of the controller and other users. 

The two main files are copies of the old driver with modifications.

The code is still rough, and will be updated before the March Indi release, but would like to submit to facilitate testing.

### DIFF
--- a/drivers.xml
+++ b/drivers.xml
@@ -53,7 +53,7 @@
             <driver name="AstroPhysics">indi_lx200ap</driver>
             <version>1.0</version>
         </device>
-        <device label="AstroPhysics" manufacturer="Astrophysics">
+        <device label="AstroPhysics Testing" manufacturer="Astrophysics">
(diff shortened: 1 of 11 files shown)


drivers.xml
```diff
@@ -53,6 +53,10 @@
             <driver name="AstroPhysics">indi_lx200ap</driver>
             <version>1.0</version>
         </device>
+        <device label="AstroPhysics" manufacturer="Astrophysics">
```
knro, line 56: Label should be Astrophysics Testing, not just Astrophysics. Label is the final driver name.

knro (changes requested):
Simple change required.

drivers.xml
murveit: Done.

PR #1622, opened 2022-02-11
Fix Celestron focuser crash after calibration

lglina:
The Celestron SCT Focuser driver may crash after focuser calibration has completed. In `TimerHit()`, `text` for an `IText` is set to the address of a stack char array. When calibration is complete and we call `IUSaveText()`, we call `realloc()` on this (now out of scope) stack memory. Correct behaviour is to use `IUSaveText()`, which will `strcpy()` to heap alloc'd memory, with a `realloc` if necessary (to create or resize the string).

We should also initialise all member variables in the constructor to avoid branch on uninitialised, which otherwise happens when we, e.g., check `calibrateInProgress` before setting it, and `backlashmove` before setting it. Running under Linux with `valgrind` now shows no memory errors.

### DIFF
--- a/drivers/focuser/celestron.cpp
+++ b/drivers/focuser/celestron.cpp
@@ -33,7 +33,12 @@
 
 static std::unique_ptr<CelestronSCT> celestronSCT(new CelestronSCT());
 
-CelestronSCT::CelestronSCT()
+CelestronSCT::CelestronSCT() :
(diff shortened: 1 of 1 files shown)


knro:
Thank you for catching this error. Indeed all text operations have to use IUSaveText.

PR #1632, opened 2022-02-28
Add CEM70(G)-EC identifier and a deprecation note

spacekitteh:


knro:
Thanks! Should we perhaps support the newer command set now?

spacekitteh:
Probably - do NSEW map directly to DEC+-/RA+-?

PR #1634, opened 2022-03-02
Add Driver for Esatto with Arco (goto, abort, sync, cal)

ChristianHen:
Add new Driver for Esatto-Arco unit based on the existing Esatto / Sestosenso-Driver

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1061,6 +1061,16 @@ add_executable(indi_sestosenso2_focus ${sesto2_SRC})
 target_link_libraries(indi_sestosenso2_focus indidriver)
 install(TARGETS indi_sestosenso2_focus RUNTIME DESTINATION bin)
 
+################ Esatto with Arco Focuser ################
+SET(esarc_SRC
+    ${CMAKE_CURRENT_SOURCE_DIR}/drivers/focuser/esattoarco.cpp)
+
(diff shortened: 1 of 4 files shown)


knro:
Thanks, I think you should move this to rotator directly unless it's BOTH a focuser + rotator? Isn't this for the Arco 2" and 3" ?

ChristianHen:
Ist is both, Focuser and Rotator.

knro:
Please check CI build failure. Also, where all functionalities tested? everything in the INDI control panel is confirmed to function correctly?

ChristianHen:
I changed one line in the .cpp ( if(!busy==0) --> if(!(busy==0) ) then the MacOs Workflow ran successfully in my repo.

PyIndi and Docker failed. Docker-Workflow claims that the repo-name must be lower case. I dont' know what to do about these two workflows.

ChristianHen:
I created a second github-account with a username that is completely lower case. There the docker-workflow runs successfully.
So dokker, macOS and Linux are ok.

knro:
I see that Arco can be used separately... so this driver required a combination of Essato + Arco together I presume?

ChristianHen:
Correct

spacekitteh:
> Really.. okay thanks for clarifying this.. so Arco can never be used standalone??

Yeah, it's extreme vendor lock-in.

PR #1642, opened 2022-03-13
Change driver name. Allow ParkTo & UnparkFrom to be visible before co…

murveit:
…nnecting.


knro:
@murveit Can you please check?

PR #1665, opened 2022-06-12
Update dmfc.cpp

sonny486:
Adds Pegasus Astro Prodigy focuser to DMFC code.  Prodigy command code is the same as the DMFC Code.
Tested on my Prodigy Microfocuser and works well.
@knro


knro:
Perhaps add an additional alias to drivers.xml as well?

sonny486:
Done, however I am unsure of versioning rules.

PR #1667, opened 2022-06-12
Update drivers.xml

sonny486:
@knro - I am unsure about the rules for versioning.


sonny486:
@knro Do I also need to add to indidrivers.xml on kstars/data?  I can do that too if needed.

knro:
We sync KStars indidrivers.xml from time to time, so we'll do that.

PR #1668, opened 2022-06-15
Fast blob transfer

pludov:
This MR adds support for local connection and fast memory buffer exchange in the indi protocol.

Working that way, data for BLOB (fits, stream, ...) needs no more being copied/base64 converted. The same memory is directly shared by driver to the client. This is an obvious win for CPU usage and latency, especially on low-end HW (rpi).

This works only for client/server located on the same unix/macos host. In that case, BLOB are written into buffers (shm or memfd) that are then exchanged by reference and shared & mmaped in the client. This is very lightweight compared to the existing base64 transfer

For remote connection, TCP is still supported for remote clients, unchanged. However, the shared buffer are used between driver and server, to eliminate handling there. In that case, the server handle the base64 encoding on a dedicated work thread.

Client that attempts to connect to localhost will be redirected to the local socket of the unix domain to take advantage. It is possible to target a specific unix socket path by using the syntax: `localhost:/path/to/socket` (an arg to indiserver is available to decide the path it listen on)

For client, since the existing semantic allows them to modify the blob data and that is not compatible with the new mechanism (blob are received as readonly), I added a new function for the client to explicitely allow readonly blob data. This removes one more copy of the data: 
```
    camera_client->enableDirectBlobAccess(MYCCD, nullptr);
```

There are further optimisations possible to avoid more memory copies, on the driver side (like producing the camera frame directly in the memory buffer instead of copying).

The MR also adds :
   * a convertion of indiserver to cpp with stl
   * usage of libev for optimized event loop in indiserver
   * some integration/non regression test for indiserver

Feedback are welcome, especially for MacOS, since I don't have access to that system...


indiserver.cpp
```diff
@@ -0,0 +1,3703 @@
...
+
+    char logname[128];
+    snprintf(logname, 128, LOGNAME, getlogin());
+    fprintf(stderr, "switching stderr to %s", logname);
+    freopen(logname, "w", stderr);
+
+    fifo = new Fifo();
+    fifo->name = FIFONAME;
+    verbose   = 1;
+    ac        = 0;
```
jpaana, line 835: Any particular reason why argument parsing is disabled in this mode?

indiserver.cpp
```diff
@@ -0,0 +1,3703 @@
...
+                log(fmt("name: %s - dp->dev[0]: %s\n", tName, dp->dev.empty() ? "" : dp->dev.begin()->c_str()));
+
+                /* If device name is given, check against it before shutting down */
+                //if (tName[0] && strcmp(dp->dev[0], tName))
+                if (tName[0] && !dp->isHandlingDevice(tName))
+                    continue;
+                if (verbose)
+                    log(fmt("FIFO: Shutting down driver: %s\n", tDriver));
+
+                //                    for (i = 0; i < dp->ndev; i++)
```
jpaana, line 1672: Some problem with this commented code?

knro:
@pludov This is outstanding work! As a drop-in replacement, everything worked 100% from first run when running everything locally. Have to test this using different scenarios, but it's looking good so far! There are lots of clang code warnings that would probably need to be addressed.

Looks like all the BLOB work is managed by baseclient which would make clients transitioning to this very painless indeed. 

Optimizations:
1. Pre-allocate FITS buffer?
2. Can GammaLut16 be optimized further? We really need it as sending 16bit frames would be too dark at the client side. Also, very few clients, if any, have support for RGB48 (16bit per channel). The LUT loop is probably a prime target for CPU vector instructions? Maybe this is already done by GCC at -O3?

knro:
@xsnrg What needs to be done to make this compiles on Fedora? I see libev was added already.

knro:
Minor issues:
+ Add copyright header to each new file.
+ Add your own copyright info to each header you modified

indiserver.cpp
```diff
@@ -0,0 +1,3703 @@
...
+#define FIFONAME "/tmp/indiserverFIFO"
+#endif
+
+#define STRINGIFY_TOK(x) #x
+#define TO_STRING(x) STRINGIFY_TOK(x)
+
+static ev::default_loop loop;
+
+template<class M>
+class ConcurrentSet {
```
knro, line 119: Might be a good idea to break this down into multiple files

indiserver.cpp
```diff
@@ -0,0 +1,3703 @@
...
+    SerializedMsg(Msg * parent);
+    virtual ~SerializedMsg();
+
+    // Calling requestContent will start production
+    // Return true if some content is available
+    bool requestContent(const MsgChunckIterator & position);
+
+    // Return true if some content is available
+    // It is possible to have 0 to send, meaning end was actually reached
+    bool getContent(MsgChunckIterator & position, void * & data, ssize_t & nsend, std::vector<int> & sharedBuffers);
```
knro, line 352: Signature doesn't match function call at line 2777

knro:
A few fprintf(stderr) messages need to be removed. e.g.

fprintf(stderr, "parseAttachedBlobs\n");

or 

2022-06-16T18:48:53: Driver indi_qhy_ccd: Received 1 fds
2022-06-16T18:48:53: Found one fd !

knro:
Chained servers tested and work exactly as expected, great work!

indiserver.cpp
pludov: It is already the case on master branch...

pludov:
> @xsnrg What needs to be done to make this compiles on Fedora? I see libev was added already.

It's working on my repo. Probably just need a restart of the "docker" action to build with libev installed

lboclboc (approved):

lboclboc:
Really good PR, ive been wanting this function also.

knro:
@xsnrg @jochym Can you please check why it's failing? the docker containers can be rebuilt?

knro:
@pludov Thank you! The PR is merged and I added support to KStars (just one line) so this would give users ability to test now and report back any issues.

I think we may have a few pending issues:
+ BaseClient Qt support.
+ Documentation: Need to explain the changes in detail and what clients need to do in order to support this. Not all clients use BaseClient so they need to understand how to implement this.

pludov:
> * Optimization of FITS: I will do this, found one small trick that could help reduce amount of memory reallocation. I couldn't figure out a way to access fptr internal buffers to directly write there.

I think it would be enough to pass allocate the ptr passed to fits_create_memfile  using IDSharedBlobAlloc /  IDSharedBlobFree, and use a rough estimation of the size of the fits for the initial size (pixel storage + enough for worst case fits header, like 128K).

jpaana:
I now had time to test this and it seems to have a regression that affects my startup scripts. I have a shell script that uses indi_setprop to connect devices and force load their configs and neither of these happen now, ie. devices are started, but not connected or their configs loaded.

The script consists of lines like:
indi_setprop "ScopeDome Dome.CONNECTION.CONNECT=On"
indi_setprop "OpenWeatherMap.CONNECTION.CONNECT=On"
indi_setprop "Allsky.CONNECTION.CONNECT=On"
indi_setprop "USB_Dewpoint.CONNECTION.CONNECT=On"
sleep 3
indi_setprop "ScopeDome Dome.CONFIG_PROCESS.CONFIG_LOAD=On"
indi_setprop "OpenWeatherMap.CONFIG_PROCESS.CONFIG_LOAD=On"
indi_setprop "Allsky.CONFIG_PROCESS.CONFIG_LOAD=On"
indi_setprop "USB_Dewpoint.CONFIG_PROCESS.CONFIG_LOAD=On"

ie. there are multiple requests for multiple devices in succession, but sleep before their configs are loaded to account for the connection time and this has worked fine until today.

Indiserver now gives different output when the commands are issued:

2022-06-18T15:19:09: Client 30: new arrival from 127.0.0.1:38390 - welcome!
2022-06-18T15:19:09: Client 30: write: Broken pipe

before they were like:

2022-06-18T15:20:17: Client 39: new arrival from 127.0.0.1:38404 - welcome!
2022-06-18T15:20:19: Client 39: read EOF
2022-06-18T15:20:19: Client 39: shut down complete - bye!

Connecting the devices from KStars works fine, but I start a subset of my devices 24/7 on boot for monitoring purposes (I have Cacti graphics temperatures, humidities, etc.) so I don't want to start it just for this.

It looks a bit like the property write gets lost if the client immediately disconnects like indi_setprop does, so it might affect other uses as well.

knro:
fits_create_memfile manages the memory buffer and its size. How do we pass this to IDSharedBobAlloc which is in INDI server? The fits_create_memfile is in INDI::CCD class at the driver side.

pludov:
> fits_create_memfile manages the memory buffer and its size. How do we pass this to IDSharedBobAlloc which is in INDI server? The fits_create_memfile is in INDI::CCD class at the driver side.

Something like that:

```c++
            //  Now we have to send fits format data to the client
            memsize = 128000 + nelements * depth(byte_type);
            memptr  = IDSharedBlobAlloc(memsize);
            if (!memptr)
            {
                LOGF_ERROR("Error: failed to allocate memory: %lu", memsize);
                return false;
            }

            fits_create_memfile(&fptr, &memptr, &memsize, 2880, IDSharedBlobRealloc, &status);

            // Later:  free memptr exclusively using IDSharedBlobFree
```

pludov:
> It looks a bit like the property write gets lost if the client immediately disconnects like indi_setprop does, so it might affect other uses as well.

It's quite possible... I'll look into it.

knro:
> > fits_create_memfile manages the memory buffer and its size. How do we pass this to IDSharedBobAlloc which is in INDI server? The fits_create_memfile is in INDI::CCD class at the driver side.
> 
> Something like that:
> 
> ```c++
>             //  Now we have to send fits format data to the client
>             memsize = 128000 + nelements * depth(byte_type);
>             memptr  = IDSharedBlobAlloc(memsize);
>             if (!memptr)
>             {
>                 LOGF_ERROR("Error: failed to allocate memory: %lu", memsize);
>                 return false;
>             }
> 
>             fits_create_memfile(&fptr, &memptr, &memsize, 2880, IDSharedBlobRealloc, &status);
> 
>             // Later:  free memptr exclusively using IDSharedBlobFree
> ```

Thanks! It's done in 3d63e3be5008605ce04621fb796f1a9b32c88a86

So now we have RawFrame in CCDChip. This is what we pass to camera SDKs to copy image data unto. We have two primary transfer types:
1. FITS: We write to FITS memory file using fits_write_img and we can the RawFrame. We flush then flish then upload the BLOB. This is not using the Shared BLOB.
2. Native: Use whatever in RawFrame directly (e.g. JPG). Is it worth using IDSharedBlobAlloc for this as well? We also can perform software binning in CCDChip::binFrame which allocates another BinFrame to perform the binning , then we simply swap pointers in the function. I'm not sure how to deal with this, they must all be using IDSharedBlob? is it OK to swap pointers?

jpaana:
> > It looks a bit like the property write gets lost if the client immediately disconnects like indi_setprop does, so it might affect other uses as well.
> 
> It's quite possible... I'll look into it.

Adding one second delay before calling shutdown in setINDIproperty.c:378 works around this issue. The pattern seems to be:
- indi_setprop sends getProperties message
- indiserver starts sending properties
- indi_setprop finds the property it's looking for, sends the message to set that one property and immediately calls shutdown() on the socket
- indiserver gets error from the socket write as it's still writing rest of the properties, drops the client and also the message to set that one property from that client

Adding the small delay before call to shutdown() allows the server to finish writing properties and handle the set property message before connection is closed.

pludov:
> > > It looks a bit like the property write gets lost if the client immediately disconnects like indi_setprop does, so it might affect other uses as well.

You were right. I reproduced and fixed it here: https://github.com/indilib/indi/pull/1669

rlancaste:
I think this might be having issues on MacOS.  This pull request added a new dependency libev and I created a craft recipe for that.  It now builds successfully on my computer and it built successfully on the kde binary server.  https://binary-factory.kde.org/view/MacOS/job/KStars_Nightly_macos/

But, once you start it up and connect to the camera (I tried it with ZWO and the Simulator CCD), when you try to take an exposure, KStars says "[ERROR] FITS Error: could not allocate memory".  This is definitely INDI and not KStars since I can use INDI built a few days ago and it works well with old and new versions of kstars, but if I try INDI built today with old or new kstars it does not work.

rlancaste:
Any ideas what might be wrong?

jpaana:
Noticed another issue after #1669 that sockets don't get freed properly after disconnect. As said I have cron job querying temperatures and other info from indi using indi_getprop and it makes indiserver eventually run out of file handles:

2022-06-20T14:05:19: Client 1023: new arrival from 127.0.0.1:43258 - welcome!
2022-06-20T14:05:19: Client 1023: write: Broken pipe
2022-06-20T14:05:19: Client 1023: read EOF
2022-06-20T14:05:19: Client 1023: shut down complete - bye!
2022-06-20T14:10:02: accept: Too many open files
2022-06-20T14:10:02: good bye

Looking at /proc/<pid of indiserver>/fd shows accumulation of one socket per indi_getprop run and netstat shows sockets in CLOSE_WAIT or FIN_WAIT state:

tcp        0      0 localhost:7624          localhost:43756         CLOSE_WAIT

pludov:
> Any ideas what might be wrong?

Macos support was only tested through automated test, so it may have flaws...  do you have the indiserver logs ? Also can you build indi with the INDI_BUILD_INTEGTESTS option ?

pludov:
> Noticed another issue after #1669 that sockets don't get freed properly after disconnect.

Should be fixed there : https://github.com/indilib/indi/pull/1672

jpaana:
> > Noticed another issue after #1669 that sockets don't get freed properly after disconnect.
> 
> Should be fixed there : #1672

Indeed it is, thanks!

rlancaste:
I just built it with that option turned on.  I can easily make a log.  would you like me to attach a log here?

rlancaste:
Is this what you need?

2022-06-21T16:50:26: Driver indi_simulator_ccd: read message CCD Simulator   '[ERROR] FITS Error: could not allocate memory'
2022-06-21T16:50:26: Driver indi_simulator_ccd: read message CCD Simulator   '[ERROR] FITS Error: could not allocate memory'
2022-06-21T16:50:26: Client 11: queuing <message device='CCD Simulator' name=''>
2022-06-21T16:50:26: Client 12: queuing <message device='CCD Simulator' name=''>
2022-06-21T16:50:26: Client 13: queuing <message device='CCD Simulator' name=''>
2022-06-21T16:50:26: Driver indi_simulator_ccd: 
2022-06-21T16:50:26: Driver indi_simulator_ccd: FITSIO status = 113: could not allocate memory
2022-06-21T16:50:26: Driver indi_simulator_ccd: Failed to reallocate memory (mem_truncate)
2022-06-21T16:50:26: Client 11: sending msg nq 1:
<message device="CCD Simulator" timestamp="2022-06-21T16:50:26" message="[ERROR] FITS Error: could not allocate memory"/>

2022-06-21T16:50:26: Client 12: sending msg nq 1:
<message device="CCD Simulator" timestamp="2022-06-21T16:50:26" message="[ERROR] FITS Error: could not allocate memory"/>

2022-06-21T16:50:26: Client 13: sending msg nq 1:
<message device="CCD Simulator" timestamp="2022-06-21T16:50:26" message="[ERROR] FITS Error: could not allocate memory"/>

rlancaste:
[logfile.txt](https://github.com/indilib/indi/files/8950904/logfile.txt)

pludov:
I'll check that , thx

Le mar. 21 juin 2022 à 18:56, rlancasteAstro ***@***.***> a
écrit :

pludov:
> Thanks! It's done in [3d63e3b](https://github.com/indilib/indi/commit/3d63e3be5008605ce04621fb796f1a9b32c88a86)
> 
> So now we have RawFrame in CCDChip. This is what we pass to camera SDKs to copy image data unto. We have two primary transfer types:
> 
>     1. FITS: We write to FITS memory file using fits_write_img and we can the RawFrame. We flush the FITS file then upload the BLOB. This is not using the Shared BLOB.
> 
>     2. Native: Use whatever in RawFrame directly (e.g. JPG). Is it worth using IDSharedBlobAlloc for this as well? We also can perform software binning in CCDChip::binFrame which allocates another BinFrame to perform the binning , then we simply swap pointers in the function. I'm not sure how to deal with this, they must all be using IDSharedBlob? is it OK to swap pointers?

@knro : The memory of the fits - to be sent to client shared - must be allocated for every distinct image, then passed to newBlob. It must not be reused / updated after newBlob was called. Otherwise, client will see changing data (crash garanteed) ! I think this change may also have introduced the problem seen by @rlancaste  (+ maybe a macos specific bug in the reallocation code that should not be triggered if the blob is allocated at the right size anyway )

So it's important the fits memory is allocated through IDSharedBlobAlloc  just before the call to fits_create_memfile
```
            //  Now we have to send fits format data to the client
            memsize = 128000 + nelements * depth(byte_type);
            memptr  = IDSharedBlobAlloc(memsize);
            if (!memptr)
            {
                LOGF_ERROR("Error: failed to allocate memory: %lu", memsize);
                return false;
            }

            fits_create_memfile(&fptr, &memptr, &memsize, 2880, IDSharedBlobRealloc, &status);
```

knro:
Thanks, is there any explanation as why I'm not getting any crashes with KStars now on Linux? I even tried to stress test it but it works OK.

pludov:
If the fits are of the same size, with the same header keys, probably only the pixels will be mixed between two consecutive frames... not obvious
Also maybe your client reads frames before new one are processed

I am tempted to remap a buffer readonly wjen it has been shared so this problem will not go unoticed. ...

knro:
When binning, I am clearing the buffer. But also I think we need to do that for subframing. Or we can simply allocate for each FITS separately as you suggested. I just though we save on memory allocation speed, but maybe it's north worth it.

pludov:
> When binning, I am clearing the buffer. But also I think we need to do that for subframing. Or we can simply allocate for each FITS separately as you suggested. I just though we save on memory allocation speed, but maybe it's north worth it.
> 
> EDIT: I submitted the necessary changes as you suggested.

I'll provide a patch to use blob in fits where possible. The cost of allocation is probably worse it when having shared blob (the memory will be allocated anyway...)

knro:
I think fits_create_memfile is failing on MacOS:

2022-06-22T17:33:02: Driver indi_simulator_ccd: FITSIO status = 113: could not allocate memory
2022-06-22T17:33:02: Driver indi_simulator_ccd: Failed to reallocate memory (mem_truncate)

mem_truncate compatible with sharedblob? it seems it does not support shrinking? not sure if this is why?

knro:
int ret = ftruncate(sb->fd, reallocated);

could this be where it's failing? since I'm just seeing truncate error?

knro:
Yes I can confirm this as well:
```
jasem@Jasems-Mac-M1-4 ~ % ipcs -M
IPC status from <running system> as of Wed Jun 22 20:45:47 +03 2022
shminfo:
        shmmax: 4194304 (max shared memory segment size)
        shmmin:       1 (min shared memory segment size)
        shmmni:      32 (max number of shared memory identifiers)
        shmseg:       8 (max shared memory segments per process)
        shmall:    1024 (max amount of shared memory in pages)
```

So we avoid using sharedBLOB completely on MacOS?

rlancaste:
Please let me know if there is anything I can test further or if you need anything else.

knro:
@pludov Shall we disable shared BLOBs for MacOS? limit it to Linux?

rlancaste:
Ok, I don't pretend to know what is best, or the limitations of various methods, but is there not an alternative method to do this that will work for MacOS such as fifo files, memory mapped files, sockets, or even just splitting files for transfer so they don't exceed that 4 MB limit you mentioned?

pludov:
Memory mapped files would technically be an option, but they would involve
diskio - as afaik there is no ramdisk on macos - so not better than
previous situation.

There is also the option to raise the 4m limit through sysctl, but it seems
to be very complicated since catalina...

Le jeu. 23 juin 2022 à 21:23, rlancasteAstro ***@***.***> a
écrit :

pludov:
How to raise that limit is explained here - not for newbies! :
https://developer.apple.com/forums/thread/669625

rlancaste:
Yeah, I don't recommend we try raising that limit, that is not user friendly.  But if there is a way to maybe use a fifo file or memory mapped file, I think that would be better

sonny486:
I can confirm the memory leak on UBUNTU 20.04 running indi-server.  Any capture results in an increase of memory on HTOP until it reaches max and crashes.

pludov:
> I can confirm the memory leak on UBUNTU 20.04 running indi-server. Any capture results in an increase of memory on HTOP until it reaches max and crashes.

I think I reproduced it : This is a distinct problem indeed, specific to tcp connection handling of blob.  Can you confirm client connects to indiserver through tcp in your case ? using "localhost" as address should  transparently route to unix domain and avoid that bug.

knro:
I can't reproduce this. In 87e842046bd0311162a3192d3a13b3e558673a6d I free the shared BLOB on closing FITS file. Does this help? But I cannot see CCD simulator nor KStars memory leaking.

sonny486:
> > I can confirm the memory leak on UBUNTU 20.04 running indi-server. Any capture results in an increase of memory on HTOP until it reaches max and crashes.
> 
> I think I reproduced it : This is a distinct problem indeed, specific to tcp connection handling of blob. Can you confirm client connects to indiserver through tcp in your case ? using "localhost" as address should transparently route to unix domain and avoid that bug.

Yes I am running remote indi-server on ubuntu.  I will check @knro update to see if that helps, can't test tonight though will try in the early AM local Texas time.

sonny486:
I don't know if I see that on my side.  

I did do a free -s 1 and it appears the buff/cache is the memory allocation that is increasing.

sonny486:
lsof: WARNING: can't stat() fuse.gvfsd-fuse file system /run/user/1000/gvfs
      Output information may be incomplete.
COMMAND    PID      USER   FD      TYPE             DEVICE SIZE/OFF     NODE NAME
indiserve 1623 telescope  cwd       DIR              259,2     4096  6029314 /home/telescope
indiserve 1623 telescope  rtd       DIR              259,2     4096        2 /
indiserve 1623 telescope  txt       REG              259,2  1798600 33819672 /usr/bin/indiserver
indiserve 1623 telescope  DEL       REG                0,1                 9 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1              2053 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1                 8 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1              2052 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1                 7 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1              1029 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1              1028 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1              3079 /memfd:shm_anon
indiserve 1623 telescope  mem       REG              259,2  1369384 33820064 /usr/lib/x86_64-linux-gnu/libm-2.31.so
indiserve 1623 telescope  mem       REG              259,2  2029592 33820050 /usr/lib/x86_64-linux-gnu/libc-2.31.so
indiserve 1623 telescope  mem       REG              259,2   104984 33818753 /usr/lib/x86_64-linux-gnu/libgcc_s.so.1
indiserve 1623 telescope  mem       REG              259,2  1956992 33821941 /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.28
indiserve 1623 telescope  mem       REG              259,2    71680 33821220 /usr/lib/x86_64-linux-gnu/libev.so.4.0.0
indiserve 1623 telescope  mem       REG              259,2   157224 33820131 /usr/lib/x86_64-linux-gnu/libpthread-2.31.so
indiserve 1623 telescope  mem       REG              259,2   191504 33818285 /usr/lib/x86_64-linux-gnu/ld-2.31.so
indiserve 1623 telescope    0r      CHR                1,3      0t0        5 /dev/null
indiserve 1623 telescope    1w      REG              259,2     4014  4194338 /tmp/indiserver.log
indiserve 1623 telescope    2w      REG              259,2     4014  4194338 /tmp/indiserver.log
indiserve 1623 telescope    3u  a_inode               0,14        0    14444 [eventpoll]
indiserve 1623 telescope    4u  a_inode               0,14        0    14444 [eventfd]
indiserve 1623 telescope    5u     IPv4              41961      0t0      TCP *:7624 (LISTEN)
indiserve 1623 telescope    6u     unix 0xffff9c23032d8880      0t0    41962 @/tmp/indiserver type=STREAM
indiserve 1623 telescope    7r     FIFO              259,2      0t0  4194337 /tmp/indiFIFO
indiserve 1623 telescope    8u     IPv4              41969      0t0      TCP telescope2.local:7624->Telescope-Desktop.local:62517 (ESTABLISHED)
indiserve 1623 telescope    9u     unix 0xffff9c23032d9980      0t0    41964 type=STREAM
indiserve 1623 telescope   10r     FIFO               0,13      0t0    41965 pipe
indiserve 1623 telescope   11u     unix 0xffff9c23032d8440      0t0    41967 type=STREAM
indiserve 1623 telescope   12r     FIFO               0,13      0t0    41968 pipe
indiserve 1623 telescope   13u     IPv4              41978      0t0      TCP telescope2.local:7624->Telescope-Desktop.local:62518 (ESTABLISHED)
root@telescope2:/home/telescope#

sonny486:
lsof: WARNING: can't stat() fuse.gvfsd-fuse file system /run/user/1000/gvfs
      Output information may be incomplete.
COMMAND    PID      USER   FD      TYPE             DEVICE SIZE/OFF     NODE NAME
indiserve 1623 telescope  cwd       DIR              259,2     4096  6029314 /home/telescope
indiserve 1623 telescope  rtd       DIR              259,2     4096        2 /
indiserve 1623 telescope  txt       REG              259,2  1798600 33819672 /usr/bin/indiserver
indiserve 1623 telescope  DEL       REG                0,1              2055 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1              3082 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1              1030 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1              3081 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1              3080 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1              2054 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1                10 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1                 9 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1              2053 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1                 8 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1              2052 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1                 7 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1              1029 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1              1028 /memfd:shm_anon
indiserve 1623 telescope  DEL       REG                0,1              3079 /memfd:shm_anon
indiserve 1623 telescope  mem       REG              259,2  1369384 33820064 /usr/lib/x86_64-linux-gnu/libm-2.31.so
indiserve 1623 telescope  mem       REG              259,2  2029592 33820050 /usr/lib/x86_64-linux-gnu/libc-2.31.so
indiserve 1623 telescope  mem       REG              259,2   104984 33818753 /usr/lib/x86_64-linux-gnu/libgcc_s.so.1
indiserve 1623 telescope  mem       REG              259,2  1956992 33821941 /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.28
indiserve 1623 telescope  mem       REG              259,2    71680 33821220 /usr/lib/x86_64-linux-gnu/libev.so.4.0.0
indiserve 1623 telescope  mem       REG              259,2   157224 33820131 /usr/lib/x86_64-linux-gnu/libpthread-2.31.so
indiserve 1623 telescope  mem       REG              259,2   191504 33818285 /usr/lib/x86_64-linux-gnu/ld-2.31.so
indiserve 1623 telescope    0r      CHR                1,3      0t0        5 /dev/null
indiserve 1623 telescope    1w      REG              259,2     4413  4194338 /tmp/indiserver.log
indiserve 1623 telescope    2w      REG              259,2     4413  4194338 /tmp/indiserver.log
indiserve 1623 telescope    3u  a_inode               0,14        0    14444 [eventpoll]
indiserve 1623 telescope    4u  a_inode               0,14        0    14444 [eventfd]
indiserve 1623 telescope    5u     IPv4              41961      0t0      TCP *:7624 (LISTEN)
indiserve 1623 telescope    6u     unix 0xffff9c23032d8880      0t0    41962 @/tmp/indiserver type=STREAM
indiserve 1623 telescope    7r     FIFO              259,2      0t0  4194337 /tmp/indiFIFO
indiserve 1623 telescope    8u     IPv4              41969      0t0      TCP telescope2.local:7624->Telescope-Desktop.local:62517 (ESTABLISHED)
indiserve 1623 telescope    9u     unix 0xffff9c23032d9980      0t0    41964 type=STREAM
indiserve 1623 telescope   10r     FIFO               0,13      0t0    41965 pipe
indiserve 1623 telescope   11u     unix 0xffff9c23032d8440      0t0    41967 type=STREAM
indiserve 1623 telescope   12r     FIFO               0,13      0t0    41968 pipe
indiserve 1623 telescope   13u     IPv4              41978      0t0      TCP telescope2.local:7624->Telescope-Desktop.local:62518 (ESTABLISHED)
root@telescope2:/home/telescope#

pludov:
That confirms you hit the bug I found, i am fixing it

Le ven. 24 juin 2022 à 17:41, Sonny Cavazos ***@***.***> a
écrit :

sonny486:
Glad I could be of some help!

Awesome work by the way.

sonny486:
Will have to wait for commit.  Will keep an eye on it.

sonny486:
I will build your fork and check it out.

pludov:
> Can GammaLut16 be optimized further? We really need it as sending 16bit frames would be too dark at the client side. Also, very few clients, if any, have support for RGB48 (16bit per channel). The LUT loop is probably a prime target for CPU vector instructions? Maybe this is already done by GCC at -O3?

I've checked the speed of Lut16 and some variation around it... I've used a test program  (-O3) to test the approach, and run against random 4Mb buffers to get timings. The program is single threaded.

On good x86, the LUT table implementation is fine. The table fits in cache and the convertion occurs at a speed of ~ 1.8G sample/seconds. I doubt it can be made very more efficient (maybe using AVX2 that has a dedicated instruction for parallel LUT...).

lut using uint8_t : rate: 1814.676194 Mb/s
lut using uint16_t : rate: 1723.580739 Mb/s
lut on 11bit using uint8_t : rate: 1805.665275 Mb/s
naive float arithmetic: rate: 86.721807 Mb/s

(11 bits means only the 11 upper bits are used - precision is obviously lost - it should be possible to have a non linear LUT)

On smaller hardware (a RPI 2), things are very different. The L1 cache of the CPU is probably the limiting factor here:

lut using uint8_t : rate: 8.791209 Mb/s
lut using uint16_t : rate: 5.882353 Mb/s
lut on 11bit using uint8_t : rate: 28.444444 Mb/s
naive float arithmetic - rate: 0.490497 Mb/s

I submitted a PR for using a vector of uint8_t, which multiply by two the effectiveness of the cache: https://github.com/indilib/indi/pull/1680

knro:
How about using NEON/SIMB? any CPU-agnostic libraries can do this? or perhaps some libraries that can implement AVX/SIMD depending on the underlying CPU architecture?

pludov:
Only avx 2 has an interesting hw instruction for lookup (available on not
too old amd and intel). Neon doesn't.

Anyway, best is probably to not to gamma in server/driver. If you seek high
fps perfs, your connection will be local anyway, so passing 8 or 16bits
costs the same. Then client can look at opengl for more optimized rendering

Le dim. 26 juin 2022 à 16:14, Jasem Mutlaq ***@***.***> a
écrit :

knro:
@pludov Any ideas what's causing this test to fail? https://github.com/indilib/indi/runs/7307903893?check_suite_focus=true

@eric-vickery made some changes to the Docker builds as well recently

TallFurryMan:
It seems indiserver, which is an external process, might not have the time to consider the driver stopped per the test request, and returns 0 as exit code. Because that's not part of the verification, I suggest the test disregard the exit code. Eventually another test should verify the conditions leading to the unexpected exit code, but it isn't really useful to check that in the situation considered.

pludov:
Hello !

The verification of the exit code is there to ensure indiserver did not terminate by a signal (like sigsegv...).

However, the code of the test itself is reporting 0 instead of the signal. I've opened a PR here to fix that: https://github.com/indilib/indi/pull/1699

Once merged, we'll now what the signal is, but it's probably nothing good (an uncatched sigpipe ?  sigsegv ? ,...  )

pludov:
I reproduced the failure :-) It's an indiserver segmentation fault, that can occur when doing base64 encoding (it's a race condition)

The fix is here: https://github.com/indilib/indi/pull/1700

pludov:
There is another way of getting shm on MacOS (posix vs sys5) that we can
test ( in shm_creat_fd.c).

Le mer. 22 juin 2022 à 19:47, Jasem Mutlaq ***@***.***> a
écrit :

PR #1669, opened 2022-06-19
Fix for indi_setprop

pludov:
This MR fixes half closed client beeing dropped by indiserver


jpaana (approved):
Thanks, fixes the issue I had with startup scripts

PR #1671, opened 2022-06-20
Prevent rewrite of indicom.h defines

iliaplatone:
This PR prevents re-declaration of physical constants in indicom.h


iliaplatone:
There's OpenVLBI which has an INDI client which uses these declarations. These are declared in part in OpenVLBI

knro:
Shouldn't this change be in OpenVLBI then?

iliaplatone:
No because these are used in OpenVLBI routines. indicom.h offer these for photometers or receivers when OpenVLBI is not used.

PR #1672, opened 2022-06-20
Fix socket leak

pludov:
Fix indiserver keeping connections open up to a crash.


knro:
@jpaana Do you know if we need to re-build the dockers? I am not involved in this part so not sure how to proceed.

PR #1673, opened 2022-06-22
More AP changes: UDP, Home, Encoder Status, version major/minor.

murveit:


drivers/telescope/lx200ap_v2.cpp
```diff
@@ -392,11 +406,41 @@ bool LX200AstroPhysicsV2::getPECState(const char *statusString)
...
+                char *start2 = end1 + 1;
+                if (*start2 != '\0')
+                {
+                    if (strlen(start2) < sizeof(minorVersion))
+                        strncpy(minorVersion, start2, sizeof(minorVersion));
+                }
+            }
+        }
+    }
+}
```
knro, line 435: seems like a good use of Regex in warranted here. What's the format?

PR #1674, opened 2022-06-24
Fix memory leak when converting blob to base64

pludov:
The size passed to munmap didn't match the size from mmap... Not all mapping were removed


sonny486:
Compiled and testing.  Looks like the leak is fixed!  Great work!

PR #1675, opened 2022-06-24
Adding some basic PEC status and control to lx200gemini

amendolajamie:
I've added some basic PEC status and control to the lx200gemini driver.


knro:
Thanks! I think one this that is missing is you need to use PEC capability from [INDI::Telescope](https://indilib.org/api/classINDI_1_1Telescope.html) so setTelescopeCapability needs to include TELESCOPE_HAS_PEC and then you need to call setPECState(....) from INDI::Telescope when PEC state changes.

amendolajamie:
Thanks! I'll make that change.

Jamie

PR #1676, opened 2022-06-24
Write fits to shm

pludov:
Allocate memory for fits in shareable buffer to avoid copying it at the end

### DIFF
--- a/drivers/ccd/ccd_simulator.cpp
+++ b/drivers/ccd/ccd_simulator.cpp
@@ -1491,11 +1491,11 @@ void * CCDSim::streamVideo()
     return nullptr;
 }
 
-void CCDSim::addFITSKeywords(INDI::CCDChip *targetChip)
+void CCDSim::addFITSKeywords(INDI::CCDChip *targetChip, INDI::CCDChip::FitsFile * targetFile)
 {
-    INDI::CCD::addFITSKeywords(targetChip);
+    INDI::CCD::addFITSKeywords(targetChip, targetFile);
 
-    auto fptr = *targetChip->fitsFilePointer();
+    auto fptr = targetFile->fitsFile();
 
     int status = 0;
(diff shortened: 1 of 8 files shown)


knro:
is there a reason not to use the existing functions?

pludov:
> is there a reason not to use the existing functions?

The memory buffer must not be reused after newBlob gets called, so I wanted it to have a scope coherent with that (not exposed in long living structure)

knro:
I just pushed changes in e74f9e1ae5c701bc645d81c85e94fa9da9c3c357 that I think should be sufficient while keeping the same structure. 

Btw, for non-FITS files and for video streaming where we use something like this:

`uploadFile(targetChip, targetChip->getFrameBuffer(), targetChip->getFrameBufferSize(), sendImage, saveImage);`

Should the internal buffer also be allocated as shared blob?

pludov:
I had a look at your change. I still think having these stored in ccdchip is misleading since these properties just live during uploadFile. 
What about a specific object for FITS content with its own lifecycle, implementing a RAI pattern ? I updated this PR with a possible way of using RAI here, to make the lifecycle obvious. However, it has impact up to the drivers.

By the way, the existing code has a possible leak in openFitsFile, in the unlikely failure of fits_open_file, the memory is not freed.

Allocating the framebuffer into a shared blob would be good for perf, when possible. It would work the same, the framebuffer must be reallocated before every new capture. I also think dedicated object and RAI here would help keeping things clear.

knro:
Ok, we can have it as its own object then. FITSFile would perhaps be a better class name and it probably needs to go to its own header and implementation files.

Regarding preallocation, I found the 65536 is too much. In multiple tests with different cameras, I found that 2880 * 3 is more than enough to add, so uint32_t size = (2880 * 3) + nelements * (targetChip->getBPP() / 8) would cover 99% of the cases here.

knro:
Ok I made more changes so this PR is no longer compatible. I'll try to keep the FITS File handling outside.

PR #1677, opened 2022-06-24
Disable broken shm support for macos

pludov:
Disable support for shared memory protocol on MACOS by default, since the current mechanism is very limited there (4Mb max) and no evident alternatives seem to be available

It can still be explicitly enabled with -DINDI_SHARED_MEMORY=ON

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,6 +91,11 @@ OPTION (INDI_BUILD_UNITTESTS "Build INDI tests" OFF)
 OPTION (INDI_BUILD_INTEGTESTS "Build INDI integration tests" OFF)
 OPTION (INDI_BUILD_WEBSOCKET "Build INDI with Websocket support" OFF)
 OPTION (INDI_FAST_BLOB "Build INDI with Fast BLOB support" ON)
+if (UNIX AND NOT APPLE)
+    OPTION(INDI_SHARED_MEMORY "Build INDI with support for UNIX protocol with shared memory" ON)
+elseif (UNIX)
+    OPTION(INDI_SHARED_MEMORY "Build INDI with support for UNIX protocol with shared memory (require shm specific settings)" OFF)
+endif()
 OPTION (INDI_CALCULATE_MINMAX "Calculate and store image minimum and maximum values in FITS header" OFF)
 
 CHECK_FUNCTION_EXISTS(mremap HAVE_MREMAP)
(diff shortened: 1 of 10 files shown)


rlancaste:
So I have been away from the computer for the last few days and I noticed this fix was merged the other day.  I just recompiled and tested it and found that when I tried to take an image, CCD Simulator crashed.  Here is the log.  I'm not sure if it was caused by this or some other change, but as you know before this change, I wasn't able to take an image either, but it didn't crash CCD simulator.

2022-06-30T05:24:09: Driver indi_simulator_ccd: sendmsg: Bad file descriptor
2022-06-30T05:24:09: Driver indi_simulator_ccd: read EOF
<delProperty device="CCD Simulator"/>
2022-06-30T05:24:09: Client 10: queuing <delProperty device='CCD Simulator' name=''>
2022-06-30T05:24:09: Client 11: queuing <delProperty device='CCD Simulator' name=''>
2022-06-30T05:24:09: Client 12: queuing <delProperty device='CCD Simulator' name=''>
2022-06-30T05:24:09: Driver indi_simulator_ccd: restart #0

pludov:
Hi

Sorry for that,  i'll check

Le jeu. 30 juin 2022 à 07:31, rlancasteAstro ***@***.***> a
écrit :

knro:
Might be fixed in 99ff3d2c9950574db3420bbe641fea8f95ca8e26

Checking now

knro:
It's not, still crashes

```
2022-06-30T06:04:30: Driver ./indi_simulator_ccd: /GSC/bin/regions.bin: No such file or directory
2022-06-30T06:04:31: Driver ./indi_simulator_ccd: sendmsg: Bad file descriptor
2022-06-30T06:04:31: Driver ./indi_simulator_ccd: read EOF
<delProperty device="CCD Simulator"/>
2022-06-30T06:04:31: Driver ./indi_simulator_ccd: restart #0
2022-06-30T06:04:31: Driver ./indi_simulator_ccd: pid=75554 rfd=7 wfd=7 efd=13
```

Perhaps we're missing in some place ENABLE_INDI_SHARED_MEMORY check?

rlancaste:
btw, you can ignore the line that says "/GSC/bin/regions.bin: No such file or directory."  That is just because GSCDAT was not set prior to running the simulator CCD. it isn't part of the problem

pludov:
@knro : I think adding one ifdef in is_unix_io would fix that to always returns 0 on macos would fix that.  I am not able to test myself ATM...

knro:
@pludov Thanks!! that did it. No more crashes on MacOS now. @rlancaste please double check on your end.

PR #1690, opened 2022-07-06
Fix connection error for Virtuoso mount

Nekomeshi:
TalkwithAxis() function does not follow Skywatcher's communication protocol.
Skywatcher's protocol allows extra responses before leading character "=" and "!". But current TalkwithAxis() does not consider the extra responses.

Virtuoso mount mirrors operation command before sending correct response. It causes connection error.
In this commit, it is fixed.


knro:
Thank you, I can confirm it works with AZ-GTi as well without any issues.

PR #1692, opened 2022-07-10
Fix issue in AstroPhysics getApMountFeatures

murveit:
Small fix to the AP v2 driver, suggested my Mike Hanson


knro:
@xsnrg Any hopes of figuring out the Fedora error?

PR #1694, opened 2022-07-12
Fix Fedora build

TallFurryMan:


knro:
Still failing?

TallFurryMan:
From the current state of Fedora, `gcc 12.1.1-2` is supposed to include the fix to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105562. The latest gcc is `gcc 12.1.1-3` from one week ago (12.1.1-2 is quite recent too).

So I suggest to simply rebuild the docker containers.

This MR only contains a change to the `docker` workflow, so that build containers are regenerated when `.github/docker/` files are altered.

It is to be noted that the current Ubuntu build actually uses a Debian build (`debian:latest`) and should probably be changed to `ubuntu:latest`, but that goes directly to Ubuntu 22. To be decided.

knro:
Shall we proceed with merging this PR now?

TallFurryMan:
Let me add a schedule to the docker rebuild procedure. Once every week, with a `master` rebuild trigger.

TallFurryMan:
From the doc, I have no idea whether `continue-on-error` will allow the indi package to be reused by other builds.

The action used in indi 3rd party is to download the last successful one, indeed the last *fully* successful one. So to bypass the Fedora issue, it might be easier to simply remove the "fedora" build from the matrix.

I'll do that now.

PR #1696, opened 2022-07-13
Fix typo in cron schedule for docker build containers.

TallFurryMan:
This MR fixes an unfortunate array hyphen missing from the cron schedule rebuilding docker containers, as more than one schedule can be set up.


TallFurryMan:
I also added a rebuild of those containers when the workflow stating how to rebuild them is modified.

knro:
is this change also required for indi-3rdparty repo?

PR #1709, opened 2022-07-28
fix periodic error simulation

murveit:
Periodic error simulation seems broken. Fix.


knro:
@spacekitteh Can you please take a look? does this produce nice PE curve for you?

spacekitteh:
I haven't tested it, but the maths looks correct. It indeed should be seconds.

PR #1728, opened 2022-09-10
Fix locale_compat.h for newest MSVC

DarthGandalf:
https://devblogs.microsoft.com/cppblog/msvc-preprocessor-progress-towards-conformance/


knro:
Thanks. Is this backward compatible with prior versions of MSVC?

DarthGandalf:
Yeah. Works with 2017, I didn't test anything earlier.

PR #1730, opened 2022-09-12
PegasusAstro ProdigyMF Focuser Initial commit, Fix ScopsOAG Backlash

chrysikos:


drivers/focuser/pegasus_prodigyMF.cpp
```diff
@@ -0,0 +1,552 @@
...
+}
+
+bool PegasusProdigyMF::ISNewSwitch(const char *dev, const char *name, ISState *states, char *names[], int n)
+{
+    if (dev != nullptr && strcmp(dev, getDeviceName()) == 0)
+    {
+
+    }
+    return INDI::Focuser::ISNewSwitch(dev, name, states, names, n);
+}
```
knro, line 233: No need to override if not in use.

drivers/focuser/pegasus_prodigyMF.cpp
```diff
@@ -0,0 +1,552 @@
...
+    // Reverse
+    if ((rc = tty_write(PortFD, cmd, strlen(cmd), &nbytes_written)) != TTY_OK)
+    {
+        char errstr[MAXRBUF];
+        tty_error_msg(rc, errstr, MAXRBUF);
+        LOGF_ERROR("Reverse error: %s.", errstr);
+        return false;
+    }
+
+    this->ignoreResponse();
```
knro, line 459: is there a reason for this-> ?

knro (changes requested):

drivers/focuser/pegasus_prodigyMF.cpp
evansgl: This is the initial driver to help out customers use the driver to focus. The switch is required (to be implemented in next revision) as the device has a USB and Power Hub.

PR #1732, opened 2022-09-14
feat(): Adding wanderer cover v3

grm:


grm:
Through github ? or indi website ?

grm:
I added two relevant screenshots to indilib ;)

knro:
I'll check, you can also share it here. Each product has some details about it, I was trying to pull this from the website, but it would be nice if you have post a short paragraph about its feature..etc like we do for other drivers.

knro:
@grm is the Mini version also supported using the same drivers?

PR #1734, opened 2022-09-17
Fix memory leaks in indiserver

nadvornik:
Valgrind showed some memory leaks in indiserver.
This PR should fix them.


knro:
@pludov Can you please check if this is OK?

pludov:
This looks good to me !

PR #1738, opened 2022-10-05
Feature: INDI::Property onUpdate event

pawel-soja:
Hi!

Event service for INDI::Property
Example:
```cpp
mRainSwitch[0].fill("On", "");
mRainSwitch[1].fill("Off", "");
mRainSwitch.fill(getDeviceName(), "Control Rain", "", MAIN_CONTROL_TAB, IP_RW, ISR_1OFMANY, 0, IPS_IDLE);

mRainSwitch.onUpdate([this]() // new method
{
    if (mRainSwitch[0].getState() == ISS_ON)
    {
        mRainLight[0].setState(IPS_ALERT);
        mRainLight.setState(IPS_ALERT);
        mRainLight.apply("Alert! Alert! Rain detected!");
    }
    else
    {
        mRainLight[0].setState(IPS_IDLE);
        mRainLight.setState(IPS_OK);
        mRainLight.apply("Rain threat passed. The skies are clear.");
    }
    mRainSwitch.setState(IPS_OK);
    mRainSwitch.apply();
});
```
There are 2 events to choose from:
- onUpdate - updates values appropriate for the type (Switch / Number / Text / Blob) and then performs the function
- onNewValues - turns off the automatic updating of values and performs the function passing new values in the form of a map - currently implemented only for Switch

The changes are backwards compatible.
This will allow you to get rid of completely in the future:
- DefaultDevice :: ISNewSwitch
- DefaultDevice :: ISNewNumber
- DefaultDevice :: ISNewText
- DefaultDevice :: ISNewBLOB

I updated the "tutorial_five" example, which now shows how to use onUpdate.
If the feature is approved, I'd like to use it to simplify indi and indi-3rdparty.
Of course, the new indi-3rdparty will lose compatibility with the older indi libraries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2143,6 +2143,7 @@ if (INDI_BUILD_DRIVERS OR INDI_BUILD_CLIENT OR INDI_BUILD_QT5_CLIENT)
         ${CMAKE_CURRENT_SOURCE_DIR}/indidevapi.h
         ${CMAKE_CURRENT_SOURCE_DIR}/base64.h
         ${CMAKE_CURRENT_SOURCE_DIR}/libs/lilxml.h
+        ${CMAKE_CURRENT_SOURCE_DIR}/libs/indililxml.h
         ${CMAKE_CURRENT_SOURCE_DIR}/libs/libastro.h
         ${CMAKE_CURRENT_SOURCE_DIR}/libs/indicom.h
         ${CMAKE_CURRENT_SOURCE_DIR}/libs/json.h
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -210,6 +210,7 @@ SET(indiclient_CXX_SRC
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/libastro.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/basedevice.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/baseclient.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/watchdeviceproperty.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/sharedblob_parse.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indiproperty.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indiproperties.cpp
@@ -469,6 +470,7 @@ SET(indidriver_CXX_SRC
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/libastro.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/basedevice.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/defaultdevice.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/watchdeviceproperty.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/sharedblob_parse.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indiproperty.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indiproperties.cpp
--- a/drivers/agent/agent_imager.cpp
+++ b/drivers/agent/agent_imager.cpp
@@ -436,8 +436,8 @@ bool Imager::ISSnoopDevice(XMLEle *root)
 bool Imager::Connect()
 {
     setServer("localhost", 7624); // TODO configuration options
-    watchDevice(controlledCCD);
-    watchDevice(controlledFilterWheel);
+    BaseClient::watchDevice(controlledCCD);
+    BaseClient::watchDevice(controlledFilterWheel);
     connectServer();
     setBLOBMode(B_ALSO, controlledCCD, nullptr);
 
--- a/examples/tutorial_five/dome.cpp
+++ b/examples/tutorial_five/dome.cpp
@@ -24,6 +24,7 @@
 */
 
 #include "dome.h"
+#include <inditimer.h>
 
 #include <memory>
 #include <cstring>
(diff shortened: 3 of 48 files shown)


knro:
Thanks Pawel! So this is similar to ISNewSwitch for example, except perhaps it can be caught for property changes that happen programatically (i.e. not from client side)?

pawel-soja:
Yes, it is similar to ISNewSwitch / ISNewNumber ... or more precisely, the new way is based on it. If you do not overwrite the function when inheriting and the function in the base class is called (DefaultDevice::ISNewNumber), the events will be fired.

see defaultdevice.cpp
```cpp
bool DefaultDevice::ISNewNumber(const char *dev, const char *name, double values[], char *names[], int n)
{
    D_PTR(DefaultDevice);

    INDI::PropertyNumber property = getProperty(name, INDI_NUMBER); // "pretty" cast to typed property

    if (!property.isValid())
        return false;

    property.update(values, names, n); // property update and event propagate onUpdate/onNewValue

    for (Connection::Interface *oneConnection : d->connections)
        oneConnection->ISNewNumber(dev, name, values, names, n);

    return false;
}
```

From the client's side, I can also write the mechanism.
It would look something like this (on the example of tutorial_six):
```cpp
MyClient::MyClient()
{
    ccd_simulator = nullptr;

    // watch the device and call if is available
    watchDevice(MYCCD, [this](INDI::BaseDevice *dp)
    {
        ccd_simulator = dp; // save device

        // get typed property CCD_TEMPERATURE
        INDI::PropertyNumber ccdTemperature = dp->getProperty("CCD_TEMPERATURE");

        // set callback
        ccdTemperature.onUpdate([this, ccdTemperature]()
        {
            IDLog("Receving new CCD Temperature: %g C\n", ccdTemperature[0].getValue());

            if (ccdTemperature[0].getValue() == -20)
            {
                IDLog("CCD temperature reached desired value!\n");
                takeExposure();
            }
        });
    });
}
```

this approach will reduce the risk of leaks, and the INDI::Property class family will take care of data integrity.

I can write it in stages and divide it into simple pull requests, I can also continue a branch and upload a massive pull request. What do you think?

knro:
Great! Yes, I can see now how in clients it can be indeed quite useful. Please continue with the branch, don't mind a large PR!

pawel-soja:
## What's new?
Most of the changes in the code are related to the increase in readability, the "onUpdate" mechanism itself does not contain many changes.

## Friendly Xml class
- INDI::LilXmlParser - read from xml buffer or file
- INDI::LilXmlDocument - takes care of freeing up memory for documents
- INDI::LilXmlElement - element with attributes, ability to search for nested elements
- INDI::LilXmlAttribute - element attribute, has a name and value, and can be checked if it is correct
- INDI::LilXmlValue - enables automatic casting of the read value *
    - operator const char *() const { return toCString(); }
    - operator double  () const { return toDouble();  }
    - operator int     () const { return toInt();     }
    - operator ISState () const { return toISState(); }
    - operator IPState () const { return toIPState(); }
    - operator IPerm   () const { return toIPerm();   }
    - operator size_t  () const { return toInt();     }

now reading the content of the xml element, converting to the ISState type and setting it like this
```cpp
widget.setState(element.context());
```
instead
```cpp
ISState state;
crackISState(pcdataXMLEle(ep), &state);

widget.setState(state);
```

An example of parsing a document from a buffer:
```cpp
INDI::LilXmlParser xmlParser;

auto documents = xmlParser.parseChunk(buffer, n); // from buffer

for (const auto &document: documents)
{
    INDI::LilXmlElement root = document.root();
    
    std::cout << root.tagName() << std::endl;
    std::cout << root.getAttribute("device").isValid()  << std::endl;
    std::cout << root.getAttribute("device") << std::endl;

    for (const auto &element: root.getElementsByTagName("defNumber"))
    {
        std::cout << element.getAttribute("name") << std::endl;
        std::cout << element.getAttribute("label") << std::endl;
        std::cout << element.getAttribute("format") << std::endl;
    }
}
```

## INDI::BaseClient::watchDevice
In INDI::BaseClient there is a similar watchDevice method with the same name, which additionally takes a function that will be called if the device comes up.
```cpp
/** @brief Add a device to the watch list.
 *
 *  A client may select to receive notifications of only a specific device or a set of devices.
 *  If the client encounters any of the devices set via this function, it will create a corresponding
 *  INDI::BaseDevice object to handle them. If no devices are watched, then all devices owned by INDI server
 *  will be created and handled.
 */
void watchDevice(const char *deviceName);
void watchDevice(const char *deviceName, const std::function<void(INDI::BaseDevice *)> &callback); // new
```

## INDI::BaseDevice::watchProperty
To be consistent, a similar method appeared in INDI::BaseDevice. callback will be called if property with the given name is available.

```cpp
void watchProperty(const std::string &name, const std::function<void(INDI::Property)> &callback); // new
```
A specific INDI::PropertyXXX type can be specified in the callback function by specifying its type:
```cpp
// auto cast to INDI::PropertyBlob
dp->watchProperty("CCD1", [](INDI::PropertyBlob property)
{
    /* ... */
});

// auto cast to INDI::PropertyNumber
dp->watchProperty("CCD_TEMPERATURE", [](INDI::PropertyNumber property)
{
    /* ... */
});
```
## Finally, using the new mechanisms, an example of "tutorial_six" might look like this:
```cpp
MyClient::MyClient()
{
    // wait for the availability of the device
    watchDevice(MYCCD, [this](INDI::BaseDevice *dp)
    {
        mCcdSimulator = dp; // save device

        // wait for the availability of the "CONNECTION" property
        dp->watchProperty("CONNECTION", [this](INDI::Property)
        {
            IDLog("Connecting to INDI Driver...\n");
            connectDevice(MYCCD);
        });

        // wait for the availability of the "CCD_TEMPERATURE" property
        dp->watchProperty("CCD_TEMPERATURE", [this](INDI::PropertyNumber property)
        {
            if (mCcdSimulator->isConnected())
            {
                IDLog("CCD is connected.\n");
                setTemperature(-20);
            }

            // call a handler function if property changes
            property.onUpdate([property, this]()
            {
                IDLog("Receving new CCD Temperature: %g C\n", property[0].getValue());
                if (property[0].getValue() == -20)
                {
                    IDLog("CCD temperature reached desired value!\n");
                    takeExposure(1);
                }
            });
        });

        // wait for the availability of the "CCD1" property
        dp->watchProperty("CCD1", [](INDI::PropertyBlob property)
        {
            // call a handler function if property changes
            property.onUpdate([property]()
            {
                // Save FITS file to disk
                std::ofstream myfile;

                myfile.open("ccd_simulator.fits", std::ios::out | std::ios::binary);

                myfile.write(static_cast<char *>(property[0].getBlob()), property[0].getBlobLen());

                myfile.close();

                IDLog("Received image, saved as mCcdSimulator.fits\n");
            });
        });
    });
}
```

## What's next?
- I am thinking of putting empty functions in INDI::BaseMediator so that I don't need to implement them when inheriting. Everything will still be backward compatible, but if you don't want to use them, all or part of them.
- analyzing the entire project to switch to INDI::PropertyXXX - I think to introduce changes gradually so as not to resolve merge conflicts - this will allow you to get rid of or tag "deprecated" a large number of potentially unsafe functions.
- I have prepared a new version of the "tutorial_six" example. From what I can see, most of them could be refreshed and therefore I propose to make a separate pull request. Please advise me whether to do so and whether to update something else?

knro:
Fantastic changes @pawel-soja ! Can you please also submit a (PR to update the documentation)[https://github.com/indilib/docs]?

pawel-soja:
Sure, I can take care of this. But to fix all the low-level calls, there is still some work on the main classes (BaseClient, BaseDevice, DefaultDevice).

I noticed that (client side)
https://github.com/indilib/indi/blob/59616f6f9d0474392c0c763cfe4b5249e134e3bc/examples/tutorial_six/tutorial_client.cpp#L57
```cpp
// or new callback version
watchDevice(MYCCD, [](INDI::BaseDevice *device) {
    /* ... */
});
```

is an analogy of this (driver side)
https://github.com/indilib/indi/blob/59616f6f9d0474392c0c763cfe4b5249e134e3bc/examples/tutorial_five/dome.cpp#L94
I am thinking of standardizing it so that the observation of another device on the driver side looks the same:
```cpp
// driver side - new version of snoop
watchDevice("Rain Detector", [](INDIBaseDevice *device) {
    /* ... */
});
```

pawel-soja:
i think i made all the important changes. Anything else to correct? If not, please accept and I will start the next stages of library optimization.

knro:
Absolutely! Thanks, hopefully we can adopt these new features in the drivers with time. Great work as always Pawel!

knro:
@pawel-soja Driver version and interface are all corrupted. Looking into it....

knro:
@pawel-soja Looks like indi_strlcpy is buggy, it cuts off 1 bytes. setText("5") is not copied at all.

knro:
Ok, found other issues with clients as well. Reverting until more testing is complete.

pawel-soja:
@knro thanks for catching the problem, I will start analyzing what is happening.

pawel-soja:
strlcpy is correct. The incorrect use of strncpy is now exposed.
The function will ensure that the string will always be terminated with \0 not exceeding the declared buffer size.

pawel-soja:
I added a branch with the `setText` patch https://github.com/indilib/indi/tree/refactor/property-callback
@knro can you briefly say what else is breaking down?

PR #1741, opened 2022-10-15
Reopen - Refactoring & Feature INDI::Property onUpdate event

pawel-soja:
Pull request is a continuation of https://github.com/indilib/indi/pull/1738.
Contains many changes to the base class of the library.
The assumption is to maintain backward compatibility and introduce modern mechanisms, minimizing the amount of code and errors when writing drivers and clients.
The proposed changes may take over memory management, reference counting, etc.

This pull request affects all drivers and clients. I try not to introduce degressions, but something may fall apart.
If possible, please do tests and report any problems.
I am asking for short descriptions of what does not work.
Personally, I will try to find the problem and solve it.
Thank you very much.


knro:
baseclientqt needs to be updated as well.

pawel-soja:
Sure, I can also refresh this library with new features.
Please just tell me is there any bug? I tried to be compatible with the old solution, hence my question.
I will also enable qt client building, because I can see that it is not built in Github Actions.

pawel-soja:
ohh i see a bug.... I need to enable building in Actions ;)

pawel-soja:
I wanted to implement `shared blobs` in `BaseClientQt` as well, but I have doubts about compiling on windows systems. I will rewrite the implementation into a separate class used only for `BaseClient`. I can take care of this in the future after adding the Windows build for Github Actions.

Currently, I separated the common code for `BaseClient` and` BaseClientQt` into the `AbstractBaseClient` class.

knro:
Great progress! Do you think it's time to use shared_ptr for BaseDevice as well? We should keep BaseDevice pointer and associated signals as-is to maintain backward compatibility but since BaseDevice is a shared pointer, we ought to use shared_ptr to protect it against some race conditions upon client termination for example.

pawel-soja:
BaseDevice itself includes `std::shared_ptr`, and it was once my idea not to use raw pointers (similar to `INDI::Property` / `INDI::PropertyXXX`).
An example is the new method:
https://github.com/indilib/indi/blob/118308fa92b3f16b013d5c1399a234ca83e0f488/libs/indibase/abstractbaseclient.h#L101

Usage:
https://github.com/indilib/indi/blob/118308fa92b3f16b013d5c1399a234ca83e0f488/examples/tutorial_six/tutorial_client.h#L54
https://github.com/indilib/indi/blob/118308fa92b3f16b013d5c1399a234ca83e0f488/examples/tutorial_six/tutorial_client.cpp#L68-L74

I'm thinking of getting rid of the raw pointers and replacing that with a simple `INDI::BaseDevice`
https://github.com/indilib/indi/blob/118308fa92b3f16b013d5c1399a234ca83e0f488/libs/indibase/abstractbaseclient.h#L125-L131
But there will be some compatibility struggle again. I don't want to make these changes to this pull request because too many changes will come up and it will be hard to catch bugs.

Currently, I would like to test what is already there and if it's okay, then accept this pull request.

I have concerns about Windows, I am currently trying to run tests on this system, but I have a problem with libnova (I do not work on Windows on a daily basis and is a stranger to me).

I can dry fix bugs, but I'm not sure what's not working. If I can, please help with testing this pull request.

I am very happy that you like the changes :)

pawel-soja:
@knro, I think I was in a hurry. Which elements should build on Windows? Is it officially supported? When editing the code, I see inserts indicating that it is compiled on Windows, I also remember a year ago there was a problem, but now I see that there are elements that prevent compilation in a simple way.
`INDI_BUILD_CLIENT` and `INDI_BUILD_QT5_CLIENT` should build on Windows?

If so, libnova is not needed:
```cmake
SET(indiclient_CXX_SRC
    ${CMAKE_CURRENT_SOURCE_DIR}/libs/libastro.cpp # <========== why?
    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/basedevice.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/baseclient.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/abstractbaseclient.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/watchdeviceproperty.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/sharedblob_parse.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indiproperty.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indiproperties.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indipropertybasic.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indipropertytext.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indipropertynumber.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indipropertyswitch.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indipropertylight.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indipropertyblob.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/property/indipropertyview_client.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/libs/indibase/indistandardproperty.cpp)
```
https://github.com/indilib/indi/blob/6bae059fd241fd24b1acd0d47bdf6726d6f0b2b6/CMakeLists.txt#L210

knro:
Windows is supported 100% on the client-side. For the driver side, it was never supported before due to INDI server using UNIX only technologies. @pludov updated the INDI Server but it still relies on POSIX features to work. You can get simple drivers to perhaps work, but a lot relies of Linux/BSD/MacOS code.

I'd say include WIndows support for client side only.

pawel-soja:
I wrote and separated a library for network connections. I think it's worth a little work on the main `CMakeLists.txt` and separate the sources into separate libraries. This will allow you to control dependencies (less spaghetti in the code) and speed up the compilation process.

The interface of the new `TcpSocket` class is based on similar methods as `QTcpSocket`, with the difference that it creates its own thread for event processing. Signals & Slots are replaced with `onXXX` callbacks.

```cpp
class TcpSocket
{
    /* ... */

    public:
        TcpSocket();
        virtual ~TcpSocket();

    public:
        void connectToHost(const std::string &hostName, uint16_t port);
        void disconnectFromHost();

    public:
        int write(const char *data, int size);
        int write(const std::string &data);

    public:
        SocketError error() const;
        std::string errorString() const;

    public:
        void onConnected(const std::function<void()> &callback);
        void onDisconnected(const std::function<void()> &callback);
        void onData(const std::function<void(const char *, size_t)> &callback);
        void onErrorOccurred(const std::function<void(SocketError)> &callback);

    public:
        bool waitForDisconnected(int timeout = 2000) const;
        bool waitForConnected(int timeout = 2000) const;

    public:
        int *socketDescriptor() const;

    /* ... */
};
```

knro:
Great idea to split the monolithic CMakeLists.txt into smaller parts!

knro:
This is now marked for view.. shall I take a deep dive now and check with existing clients?

pawel-soja:
Yes, I would like to ask for a check. I don't know as much as you, what client applications are used.

knro:
Looks like this affect 3rd party drivers loading external XML properties like EQMod Mount. Can you please check? I'll investigate as well.

pawel-soja:
Sure, I'm starting to look for the problem.

pawel-soja:
Found the missing buttons problem.
I misused `dynamic` in `INDI::Property` to store free memory information for `INDI::PropertyXXX`
I did not pay attention to:
https://github.com/indilib/indi/blob/05d9db02b3d2ebfb7fc58160ebdc9d4a8f87fe0c/libs/indibase/defaultdevice.cpp#L677
@knro hope this was the bug.

By the way, I started making fixes for the current compilation problem for the repository. I undo the changes, I did not expect so many errors.

knro:
@pawel-soja I just tested today again. This time, there was an issue with BLOB loading in KStars. The FITS file cannot be read with this error:
```
FITSIO status = 107: tried to move past end of file
ffomem could not interpret primary array header of file: (ffomem)
```
Looks like BLOB are not transferred correctly.

pawel-soja:
I caught that yesterday too. I tested all the time for AF_UNIX. There is indeed a problem for a regular connection. I'll take care of it.

pawel-soja:
Of course, there was a bug parsing an xml blob (jumping scrolled image). However, it should not cause the error you mention. I tested the AF_INET / AF_UNIX connection to streaming and loading / reading fits with KStars and now I see no problem.
I will test the behavior for several cameras at once.

knro:
Right, several cases needs to be tested:

1. Test with PHD2
2. Test with remote client
3. Test with chained INDI servers

pawel-soja:
Tested, it works for me (thanks for the hint, something new for me - chained servers).
Note, the current libraries are version 1.9.9, this branch is based on 1.9.8 so it's worth cleaning the system ;)

knro:
All testing looks good here. It's ready to be merged and exposed to more users. Thanks a lot!!

PR #1743, opened 2022-10-17
Enable INDI_BUILD_QT5_CLIENT

pawel-soja:
This pull request enables INDI_BUILD_QT5_CLIENT that is building the `INDI::BaseClientQt` class.


knro:
Thank you!! This should have been enabled long ago and I just didn't know how to do it!

PR #1758, opened 2022-11-10
Refactor - split libraries

pawel-soja:
### indicore
Contains self-dependent source files used by drivers and clients. They can be compiled without additional dependencies.

### indiproperty
A set of propert management classes for C++. The library requires an indicore

The `indicore`, `indiproperty`, `sockets` libraries are built as `OBJECT` and compiled into drivers and clients. In short, no static or shared library is created after compilation.
Separation helps keep code encapsulated and improves readability and prevents clutter.


pawel-soja:
I am closing this pull request.
After changes in `BaseMediator` and `Property`, another attempt to refactor the library structure continues in https://github.com/indilib/indi/pull/1767

PR #1761, opened 2022-11-12
Shared pointers for INDI::BaseDevice

knro:
This would ensure that references to INDi::BaseDevice are safe in memory when shared among clients.

### DIFF
--- a/libs/indibase/abstractbaseclient.h
+++ b/libs/indibase/abstractbaseclient.h
@@ -125,10 +125,20 @@ class AbstractBaseClient : public INDI::BaseMediator
         /** @param deviceName Name of device to search for in the list of devices owned by INDI server,
          *  @returns If \e deviceName exists, it returns an instance of the device. Otherwise, it returns NULL.
          */
+        [[deprecated("Use getSharedDevice(const char *, std::shared_ptr<INDI::BaseDevice> &).")]]
         INDI::BaseDevice *getDevice(const char *deviceName);
 
+        /** @param deviceName Name of device to search for in the list of devices owned by INDI server,
+         *  @param dp shared pointer to device, if found.
+         *  @returns If \e deviceName exists, it returns true, otherwise false.
+         */
+        bool getSharedDevice(const char *deviceName, std::shared_ptr<BaseDevice> &dp);
+
         /** @returns Returns a vector of all devices created in the client. */
+        [[deprecated("Use getSharedDevices().")]]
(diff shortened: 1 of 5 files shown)


knro:
Looks like all the deprecated warnings I added causing CI failures.

libs/indibase/abstractbaseclient.h
```diff
@@ -125,11 +125,21 @@ class AbstractBaseClient : public INDI::BaseMediator
...
          *  @returns If \e deviceName exists, it returns an instance of the device. Otherwise, it returns NULL.
          */
+        [[deprecated("Use getSharedDevice(const char *, std::shared_ptr<INDI::BaseDevice> &).")]]
         INDI::BaseDevice *getDevice(const char *deviceName);
 
+        /** @param deviceName Name of device to search for in the list of devices owned by INDI server,
+         *  @param dp shared pointer to device, if found.
+         *  @returns If \e deviceName exists, it returns true, otherwise false.
+         */
+        bool getSharedDevice(const char *deviceName, std::shared_ptr<BaseDevice> &dp);
```
pawel-soja, line 135: why not simply:
```cpp
std::shared_ptr<BaseDevice> getSharedDevice(const char *deviceName);
```
then easy to test and use:
```cpp
if (auto device = getSharedDevice("..."))
{
  /* ... */
}
```

pawel-soja:
I suggest to use like with `INDI::Property` to pass `INDI::BaseDevice` class as value. This class has a shared_ptr in it so the data will be shared.
Additionally, it will continue to refer to the method completely safely. (https://github.com/indilib/indi/pull/1453).

```cpp
std::cout << client.getDevice("non-exist").getProperty("non-exist").getStateAsString() << std::endl; // prints "Alert"
```

knro:
@pawel-soja This is what I started with, but faced an issue and had to explicitly used shared_ptr<...>. If you can get it to work then that would be great!

pawel-soja:
@knro  Of course! Please roll back changes on master.
I will take care of the reimplementation of classes/methods and take into account the additional fixes you have made.

knro:
@pawel-soja Done!

knro:
@pawel-soja Just found another important regression. When INDI server is terminated, the client is not informed. It appears serverDisconnected is not triggered.

pawel-soja:
@knro Wow! The implementation of the onErrorOccurred event handler is missing.
Fixed: https://github.com/indilib/indi/pull/1763

PR #1764, opened 2022-11-14
Built-in shared pointers for INDI::BaseDevice

pawel-soja:
Referring to https://github.com/indilib/indi/pull/1761. This pull request aims to get rid of the use of raw pointers and use the `shared_ptr` which is in the `INDI::BaseDevice` class to store the data.

### BaseClient
Previously:
https://github.com/indilib/indi/blob/6b5ac8408ef021eb832479d4392aa97b66687c78/libs/indibase/abstractbaseclient.h#L128-L131

Now:
https://github.com/indilib/indi/blob/a53ae5bd45aa4c37361e5d50a28cb73b724f54c5/libs/indibase/abstractbaseclient.h#L128-L131

### BaseMediator
The virtual methods in `INDI::BaseMediator` require a bit of talking.
https://github.com/indilib/indi/blob/6b5ac8408ef021eb832479d4392aa97b66687c78/libs/indibase/indibase.h#L88-L103

This class is no longer needed after the introduction of `watchDevice` and `watchProperty`  https://github.com/indilib/indi/pull/1738
There is also passing `BaseDevice` by the pointer, but it will not be a problem to convert it to a value (built-in shared_ptr).

### DIFF
--- a/libs/indibase/abstractbaseclient.cpp
+++ b/libs/indibase/abstractbaseclient.cpp
@@ -72,7 +72,7 @@ int AbstractBaseClientPrivate::dispatchCommand(const LilXmlElement &root, char *
     // Ignore echoed newXXX
     if (root.tagName().find("new") == 0)
     {
-        return 0;       
+        return 0;
     }
 
     if (root.tagName() == "pingRequest")
@@ -114,7 +114,8 @@ int AbstractBaseClientPrivate::dispatchCommand(const LilXmlElement &root, char *
         return 0;
     }
 
-    return watchDevice.processXml(root, errmsg, [this]() { // create new device if nessesery
+    return watchDevice.processXml(root, errmsg, [this]  // create new device if nessesery
+    {
         BaseDevice *device = new BaseDevice();
         device->setMediator(parent);
         return device;
@@ -405,7 +406,7 @@ std::vector<BaseDevice> AbstractBaseClient::getDevices() const
 bool AbstractBaseClient::getDevices(std::vector<BaseDevice> &deviceList, uint16_t driverInterface )
 {
     D_PTR(AbstractBaseClient);
-    for (auto &it: d->watchDevice)
+    for (auto &it : d->watchDevice)
     {
         if (it.second.device.getDriverInterface() & driverInterface)
             deviceList.push_back(it.second.device);
--- a/libs/indibase/abstractbaseclient.cpp
+++ b/libs/indibase/abstractbaseclient.cpp
@@ -22,6 +22,9 @@
 #include "abstractbaseclient.h"
 #include "abstractbaseclient_p.h"
 
+#include "basedevice.h"
+#include "basedevice_p.h"
+
 #include "indiuserio.h"
 #include "locale_compat.h"
 #include "indistandardproperty.h"
@@ -122,7 +125,7 @@ int AbstractBaseClientPrivate::deleteDevice(const char *devName, char *errmsg)
 {
     if (auto device = watchDevice.getDeviceByName(devName))
     {
-        parent->removeDevice(device);
+        device.d_ptr->mediateRemoveDevice();
         watchDevice.deleteDevice(device);
         return 0;
     }
@@ -157,7 +160,9 @@ int AbstractBaseClientPrivate::delPropertyCmd(const LilXmlElement &root, char *e
     if (auto property = dp.getProperty(propertyName))
     {
         if (sConnected)
-            parent->removeProperty(property);
+        {
+            dp.d_ptr->mediateRemoveProperty(property);
+        }
         return dp.removeProperty(propertyName, errmsg);
     }
 
--- a/libs/indibase/abstractbaseclient.cpp
+++ b/libs/indibase/abstractbaseclient.cpp
@@ -116,7 +116,7 @@ int AbstractBaseClientPrivate::dispatchCommand(const LilXmlElement &root, char *
 
     return watchDevice.processXml(root, errmsg, [this]  // create new device if nessesery
     {
-        BaseDevice device;
+        ParentDevice device(ParentDevice::Valid);
         device.setMediator(parent);
         return device;
     });
--- a/libs/indibase/abstractbaseclient.cpp
+++ b/libs/indibase/abstractbaseclient.cpp
@@ -614,45 +614,6 @@ void AbstractBaseClient::newUniversalMessage(std::string message)
     IDLog("%s\n", message.c_str());
 }
 
-void AbstractBaseClient::newDevice(INDI::BaseDevice *)
-{ }
-
-void AbstractBaseClient::removeDevice(INDI::BaseDevice *)
-{ }
-
-void AbstractBaseClient::newProperty(INDI::Property *)
-{ }
-
-void AbstractBaseClient::removeProperty(INDI::Property *)
-{ }
-
-void AbstractBaseClient::newBLOB(IBLOB *)
-{ }
-
-void AbstractBaseClient::newSwitch(ISwitchVectorProperty *)
-{ }
-
-void AbstractBaseClient::newNumber(INumberVectorProperty *)
-{ }
-
-void AbstractBaseClient::newText(ITextVectorProperty *)
-{ }
-
-void AbstractBaseClient::newLight(ILightVectorProperty *)
-{ }
-
-void AbstractBaseClient::newMessage(INDI::BaseDevice *, int)
-{ }
-
-void AbstractBaseClient::serverConnected()
-{ }
-
-// avoid calling pure virtual method from destructor
-void AbstractBaseClient::serverDisconnected(int exit_code)
-{
-    INDI_UNUSED(exit_code);
-}
-
 }
 
 #if defined(_MSC_VER)
--- a/libs/indibase/abstractbaseclient.cpp
+++ b/libs/indibase/abstractbaseclient.cpp
@@ -126,7 +126,7 @@ int AbstractBaseClientPrivate::deleteDevice(const char *devName, char *errmsg)
 {
     if (auto device = watchDevice.getDeviceByName(devName))
     {
-        device.d_ptr->mediateRemoveDevice(device);
+        device.detach();
         watchDevice.deleteDevice(device);
         return 0;
     }
--- a/libs/indibase/abstractbaseclient.cpp
+++ b/libs/indibase/abstractbaseclient.cpp
@@ -138,27 +138,27 @@ int AbstractBaseClientPrivate::deleteDevice(const char *devName, char *errmsg)
 int AbstractBaseClientPrivate::delPropertyCmd(const LilXmlElement &root, char *errmsg)
 {
     /* dig out device and optional property name */
-    BaseDevice *dp = watchDevice.getDeviceByName(root.getAttribute("device"));
+    BaseDevice dp = watchDevice.getDeviceByName(root.getAttribute("device"));
 
-    if (dp == nullptr)
+    if (!dp.isValid())
         return INDI_DEVICE_NOT_FOUND;
 
-    dp->checkMessage(root.handle());
+    dp.checkMessage(root.handle());
 
     const auto propertyName = root.getAttribute("name");
 
     // Delete the whole device if propertyName does not exists
     if (!propertyName.isValid())
     {
-        return deleteDevice(dp->getDeviceName(), errmsg);
+        return deleteDevice(dp.getDeviceName(), errmsg);
     }
 
     // Delete property if it exists
-    if (auto property = dp->getProperty(propertyName))
+    if (auto property = dp.getProperty(propertyName))
     {
         if (sConnected)
             parent->removeProperty(property);
-        return dp->removeProperty(propertyName, errmsg);
+        return dp.removeProperty(propertyName, errmsg);
     }
 
     // Silently ignore B_ONLY clients.
@@ -175,11 +175,11 @@ int AbstractBaseClientPrivate::messageCmd(const LilXmlElement &root, char *errms
 {
     INDI_UNUSED(errmsg);
 
-    BaseDevice *dp = watchDevice.getDeviceByName(root.getAttribute("device"));
+    BaseDevice dp = watchDevice.getDeviceByName(root.getAttribute("device"));
 
-    if (dp)
+    if (dp.isValid())
     {
-        dp->checkMessage(root.handle());
+        dp.checkMessage(root.handle());
         return 0;
     }
 
@@ -249,44 +249,43 @@ void AbstractBaseClientPrivate::userIoGetProperties()
 
 void AbstractBaseClientPrivate::setDriverConnection(bool status, const char *deviceName)
 {
-    BaseDevice *drv                 = parent->getDevice(deviceName);
-    ISwitchVectorProperty *drv_connection = nullptr;
+    BaseDevice drv = parent->getDevice(deviceName);
 
-    if (drv == nullptr)
+    if (!drv.isValid())
     {
         IDLog("BaseClientQt: Error. Unable to find driver %s\n", deviceName);
         return;
     }
 
-    drv_connection = drv->getSwitch(SP::CONNECTION);
+    auto drv_connection = drv.getSwitch(SP::CONNECTION);
 
-    if (drv_connection == nullptr)
+    if (!drv_connection.isValid())
         return;
 
     // If we need to connect
     if (status)
     {
         // If there is no need to do anything, i.e. already connected.
-        if (drv_connection->sp[0].s == ISS_ON)
+        if (drv_connection[0].getState() == ISS_ON)
             return;
 
-        IUResetSwitch(drv_connection);
-        drv_connection->s       = IPS_BUSY;
-        drv_connection->sp[0].s = ISS_ON;
-        drv_connection->sp[1].s = ISS_OFF;
+        drv_connection.reset();
+        drv_connection.setState(IPS_BUSY);
+        drv_connection[0].setState(ISS_ON);
+        drv_connection[1].setState(ISS_OFF);
 
         parent->sendNewSwitch(drv_connection);
     }
     else
     {
         // If there is no need to do anything, i.e. already disconnected.
-        if (drv_connection->sp[1].s == ISS_ON)
+        if (drv_connection[1].getState() == ISS_ON)
             return;
 
-        IUResetSwitch(drv_connection);
-        drv_connection->s       = IPS_BUSY;
-        drv_connection->sp[0].s = ISS_OFF;
-        drv_connection->sp[1].s = ISS_ON;
+        drv_connection.reset();
+        drv_connection.setState(IPS_BUSY);
+        drv_connection[0].setState(ISS_OFF);
+        drv_connection[1].setState(ISS_ON);
 
         parent->sendNewSwitch(drv_connection);
     }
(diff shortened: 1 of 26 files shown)


libs/indibase/abstractbaseclient.cpp
```diff
@@ -523,14 +517,9 @@ void AbstractBaseClient::sendNewNumber(INDI::Property pp)
 void AbstractBaseClient::sendNewNumber(const char *deviceName, const char *propertyName, const char *elementName,
                                        double value)
 {
-    INDI::BaseDevice *drv = getDevice(deviceName);
+    auto nvp = getDevice(deviceName).getNumber(propertyName);
```
knro, line 520: Such call does not result in crash is a device does not exist for deviceName?

libs/indibase/defaultdevice.h
```diff
@@ -301,11 +301,6 @@ class DefaultDevice : public BaseDevice
          */
         virtual bool ISSnoopDevice(XMLEle *root);
 
-        /**
-         * @return getInterface Return the interface declared by the driver.
-         */
-        virtual uint16_t getDriverInterface() override;
```
knro, line 307: Why was this removed?

knro (commented):
Clarification required.

knro:
For BaseMediator, we still need this since client rely on it, we can't remove. Furthermore, how do you get notified when a new device is created by BaseClient? watchDevice assumes you already know which device to look for, correct? what if you don't? Even for properties, right now we get notifiied when *any* property is created or delete, or when any message is dispatched..etc.

libs/indibase/abstractbaseclient.cpp
pawel-soja: In case of non-existent `BaseDevice`, `BaseDevice` is returned marked with the method `isValid() == false`
Of course, `BaseDevice` can still be compared to nullptr and has additional operators to allow the use of init-statement in the condition.

### Example:
```cpp
#include <libindi/baseclient.h>
#include <libindi/basedevice.h>
#include <iostream>

int main(int, char *[])
{
    std::cout << "#1 property isValid: "
              << INDI::BaseClient().getDevice("invalid-device").getNumber("invalid-property").isValid()
              << std::endl;
    
    std::cout << "#2 isValid: "
              << INDI::BaseClient().getDevice("invalid-device").isValid()
              << std::endl;

    std::cout << "#3 compare device to nullptr: "
              << (INDI::BaseClient().getDevice("invalid-device") == nullptr)
              << std::endl;

    std::cout << "#4 compare device to not nullptr: "
              << (INDI::BaseClient().getDevice("invalid-device") != nullptr)
              << std::endl;

    if (auto device = INDI::BaseClient().getDevice("invalid-device"))
    {
        std::cout << "#5 device exists: " << device.getDeviceName() << std::endl;
    }
    else
    {
        std::cout << "#5 device not exists" << std::endl;
    }
    return 0;
}
```
### Output:
```
#1 property isValid: 0
#2 isValid: 0
#3 compare device to nullptr: 1
#4 compare device to not nullptr: 0
#5 device not exists
```

libs/indibase/defaultdevice.h
pawel-soja: #### Please look carefully:
Implementation in `DefaultDevice`:
https://github.com/indilib/indi/blob/e425d43bac5f716af11ecab10b2c2588d435f036/libs/indibase/defaultdevice.cpp#L796-L809
https://github.com/indilib/indi/blob/e425d43bac5f716af11ecab10b2c2588d435f036/libs/indibase/defaultdevice.cpp#L899-L904
Implementation in `BaseDevice`:
https://github.com/indilib/indi/blob/e425d43bac5f716af11ecab10b2c2588d435f036/libs/indibase/basedevice.cpp#L898-L908

#### Summary:
The implementations are equivalent.
For optimization, I suggest moving `interfaceDescriptor` to `BaseDevice`, and leaving property in `DefaultDevice` (so that there is still no property in `BaseDevice`)

That's right?

knro: We could since DefaultDevice inherits BaseDevice. However, BaseDevice was originally created as *Device-View* for Clients, so there was some separation in functions. We can consolidate now any duplicate features.

pawel-soja:
> For BaseMediator, we still need this since client rely on it, we can't remove. Furthermore, how do you get notified when a new device is created by BaseClient? watchDevice assumes you already know which device to look for, correct? what if you don't? Even for properties, right now we get notifiied when _any_ property is created or delete, or when any message is dispatched..etc.

Sure, I understand the problem. I am concerned about the need to reimplement clients. Unfortunately, the `deprecated` designation was not created for a virtual function that is overridden by subsequent classes.

By introducing `watchDevice` I wanted to standardize the detection of new `BaseDevice` while also getting rid of `ISSnoopDevice`.

Please see:
https://github.com/indilib/indi/blob/83539e65f16227c05522b627ded90a658a045d5f/examples/tutorial_five/dome.cpp#L151-L175

Now is:
https://github.com/indilib/indi/blob/e425d43bac5f716af11ecab10b2c2588d435f036/examples/tutorial_five/dome.cpp#L84-L112

Using `watchDevice` `watchProperty` `BaseDevice` and `Property/XXX` you can handle everything without additional problematic mechanisms.

When it comes to detecting any `device` or `property`, I can create e.g. `watchAnyDevice` or `watchDevice(BaseDevice::Any, ...)`, it will also allow pre-filtering if needed.

I can also convert `BaseMediator` to the form (as you originally assumed):
```cpp
        /** @brief Emmited when a new device is created from INDI server.
         *  @param dp Pointer to the base device instance
         */
        virtual void newDevice(INDI::BaseDevice *dp) = 0;
        virtual void newDevice(INDI::BaseDevice dp) = 0;

        /** @brief Emmited when a device is deleted from INDI server.
         *  @param dp Pointer to the base device instance.
         */
        virtual void removeDevice(INDI::BaseDevice *dp) = 0;
        virtual void removeDevice(INDI::BaseDevice dp) = 0;
        /* ... */
```

But it is worth considering which direction to go and which mechanism to ultimately support.

You can point me to sources that use `BaseMediator` where you're most concerned about getting rid of memory leaks (raw pointers) and I'll take care of it.

knro:
or perhaps prepare 1.9.9 by adding the new form without raw pointers. Then in 2.0.0 we can get rid of it completely.

pawel-soja:
I suggest that cmake automatically enter values based on a tag or values defined in CMakeLists. Then I can add macros that disable `BaseMediator` depending on version.
https://github.com/indilib/indi/blob/e425d43bac5f716af11ecab10b2c2588d435f036/indiapi.h#L136-L139

Before version 2.0.0, I would like to separate the libraries something like in https://github.com/indilib/indi/pull/1758 and improve the dependencies so that you don't have to additionally link libraries that are required, e.g. by `indiclient` (e.g. zlib) https://github.com/indilib/indi/blob/e425d43bac5f716af11ecab10b2c2588d435f036/examples/tutorial_six/CMakeLists.txt#L4-L7

> or perhaps prepare 1.9.9 by adding the new form suggested above. Then in 2.0.0 we can get rid of it completely.

you mean in version 1 it would be:
```cpp
        virtual void newDevice(INDI::BaseDevice *dp) = 0;
        virtual void newDevice(INDI::BaseDevice dp) = 0;
```
?

When would you like to introduce version 2?
I have plans to improve a few more things that would better prepare version 2.

knro:
Yes, in upcoming v1.9.9, we will have this:

```
        virtual void newDevice(INDI::BaseDevice *dp) = 0;
        virtual void newDevice(INDI::BaseDevice dp) = 0;
```

Then in 2.0.0 we can ask clients to use the new watchXXX system.

pawel-soja:
I understand. Just to clarify, will there be an in-between version? e.g. 1.9.10?

As we are already making such big changes, it would be worth changing a little more :) E.g. arrange the fields in the structures defined in `indiapi` (this will simplify `PropertyView`) or get rid of them and use only `INDI::PropertyXXX` - a lot of work (maybe from version 3.0.0)

Ultimately, I would like PropertyView not to be an overlay on `indiapi` structures and to be more user-friendly. They had memory management for blobs etc.

I would like PropertyXXX to eventually use a pointer to BaseDevice and not have to write explicitly:

https://github.com/indilib/indi/blob/e425d43bac5f716af11ecab10b2c2588d435f036/examples/tutorial_five/raindetector.cpp#L89-L107

I have prepared extension of `sockets` library to simplify indiserver.

Only time is missing and milestones need to be set, so I'm asking what the version numbering will look like.

knro:
We have a bi-monthly release, so 2.0.0 is January 31st

knro:
I updated the milestones. 2.0.0 on January 31st and 2.0.1 on March 30th.

pawel-soja:
@knro I have a question:
https://github.com/indilib/indi/blob/e425d43bac5f716af11ecab10b2c2588d435f036/libs/indibase/indibase.h#L134
why not `std::string` ?:
```cpp
    virtual void newMessage(/* ... */, const std::string &message) ...
```

pawel-soja:
@knro add macro with version below 2?
```cpp
        /* ... */
#if INDI_VERSION_MAJOR < 2
        /** @brief Emmited when a new device is created from INDI server.
         *  @param dp Pointer to the base device instance
         */
        virtual void newDevice(INDI::BaseDevice *dp); // deprecated

        /** @brief Emmited when a device is deleted from INDI server.
         *  @param dp Pointer to the base device instance.
         */
        virtual void removeDevice(INDI::BaseDevice *dp); // deprecated

        /** @brief Emmited when a new property is created for an INDI driver.
         *  @param property Pointer to the Property Container
         */
        virtual void newProperty(INDI::Property *property); // deprecated

        /** @brief Emmited when a property is deleted for an INDI driver.
         *  @param property Pointer to the Property Container to remove.
         */
        virtual void removeProperty(INDI::Property *property); // deprecated

        /** @brief Emmited when a new switch value arrives from INDI server.
         *  @param svp Pointer to a switch vector property.
         */
        virtual void newSwitch(ISwitchVectorProperty *svp); // deprecated

        /** @brief Emmited when a new number value arrives from INDI server.
         *  @param nvp Pointer to a number vector property.
         */
        virtual void newNumber(INumberVectorProperty *nvp); // deprecated

        /** @brief Emmited when a new text value arrives from INDI server.
         *  @param tvp Pointer to a text vector property.
         */
        virtual void newText(ITextVectorProperty *tvp); // deprecated

        /** @brief Emmited when a new light value arrives from INDI server.
         *  @param lvp Pointer to a light vector property.
         */
        virtual void newLight(ILightVectorProperty *lvp); // deprecated

        /** @brief Emmited when a new message arrives from INDI server.
         *  @param dp pointer to the INDI device the message is sent to.
         *  @param messageID ID of the message that can be used to retrieve the message from the device's messageQueue() function.
         */
        virtual void newMessage(INDI::BaseDevice *dp, int messageID); // deprecated
#endif
```

The library already has changes made to compile. Will you take care of other clients so that after January 31st the whole world will still compile? :)

knro:
> @knro I have a question:
> 
> https://github.com/indilib/indi/blob/e425d43bac5f716af11ecab10b2c2588d435f036/libs/indibase/indibase.h#L134
> 
> 
> why not `std::string` ?:
> ```c++
>     virtual void newMessage(/* ... */, const std::string &message) ...
> ```

IIRC, this was causing a client lockup back in the early days when this client was developed. Moving forward, I think that each device should emit its own message (perhaps by having watchMessage for example), and then the client can emit the universal messages not associated with any particular device.

knro:
> @knro add macro with version below 2?
> 
> ```c++
>         /* ... */
> #if INDI_VERSION_MAJOR < 2
>         /** @brief Emmited when a new device is created from INDI server.
>          *  @param dp Pointer to the base device instance
>          */
>         virtual void newDevice(INDI::BaseDevice *dp); // deprecated
> 
>         /** @brief Emmited when a device is deleted from INDI server.
>          *  @param dp Pointer to the base device instance.
>          */
>         virtual void removeDevice(INDI::BaseDevice *dp); // deprecated
> 
>         /** @brief Emmited when a new property is created for an INDI driver.
>          *  @param property Pointer to the Property Container
>          */
>         virtual void newProperty(INDI::Property *property); // deprecated
> 
>         /** @brief Emmited when a property is deleted for an INDI driver.
>          *  @param property Pointer to the Property Container to remove.
>          */
>         virtual void removeProperty(INDI::Property *property); // deprecated
> 
>         /** @brief Emmited when a new switch value arrives from INDI server.
>          *  @param svp Pointer to a switch vector property.
>          */
>         virtual void newSwitch(ISwitchVectorProperty *svp); // deprecated
> 
>         /** @brief Emmited when a new number value arrives from INDI server.
>          *  @param nvp Pointer to a number vector property.
>          */
>         virtual void newNumber(INumberVectorProperty *nvp); // deprecated
> 
>         /** @brief Emmited when a new text value arrives from INDI server.
>          *  @param tvp Pointer to a text vector property.
>          */
>         virtual void newText(ITextVectorProperty *tvp); // deprecated
> 
>         /** @brief Emmited when a new light value arrives from INDI server.
>          *  @param lvp Pointer to a light vector property.
>          */
>         virtual void newLight(ILightVectorProperty *lvp); // deprecated
> 
>         /** @brief Emmited when a new message arrives from INDI server.
>          *  @param dp pointer to the INDI device the message is sent to.
>          *  @param messageID ID of the message that can be used to retrieve the message from the device's messageQueue() function.
>          */
>         virtual void newMessage(INDI::BaseDevice *dp, int messageID); // deprecated
> #endif
> ```
> 
> The library already has changes made to compile. Will you take care of other clients so that after January 31st the whole world will still compile? :)

Yes this looks good, I think this would give clients a chance to adapt to INDI Library 2.0.0 changes.

knro:
It's not working with KStars... investigating.

pawel-soja:
ohhh I see, I rebuilt everything again and KStars is not connecting. I'll take care of it.

knro:
I'm not sure.. as it says in the git commit it was a temporary solution. IIRC, one issue is that a property is deleted in BaseDevice before removeProperty on line 645 has a chance to propagate to client. So this the removeProperty signal was made to be in Qt as a BlockingConnection type signal (i.e. blocks the calling thread until the slot is executed). Under some circumstances, the slot ends up calling another code in the blocked thread and this led to a dead lock.

If now the INDI::Property pointer is kept in memory even when it is removed from BaseDevice, then BlockingConnection is perhaps not required since any functions relying on this property can gracefully continue for a cycle or two before they are deleted by the client. So the question is: Do clients still have a valid access to the property even after it was removed in BaseDevice?

pawel-soja:
Thanks for the explanations.

> Do clients still have a valid access to the property even after it was removed in BaseDevice?

Yes.

I will analyze some more cases to make sure and check how the implementation in KStars looks like.

Currently, I am interested in one curiosity in BaseMediator.
1. I understand that `BaseMediator` was created only for clients. Yes?
2. The `virtual void newText(ITextVectorProperty *tvp);` function is called only for elements added during XML parsing. Yes?

If so, I understand, but it leads to illogical behavior:
```cpp
#include <libindi/basedevice.h>
#include <iostream>

using namespace INDI;

class Mediator: public BaseMediator
{
protected:
    // old api interface
    void newText(ITextVectorProperty *) override
    {
        std::cout << "newText old api" << std::endl;
    }

    // new api interface
    void newText(PropertyText) override
    {
        std::cout << "newText new api" << std::endl;
    }
};

int main(int, char *[])
{
    Mediator mediator;
    {
        PropertyText text{1};
        text.setName("text");
        
        BaseDevice device;

        // mediator way to detect new property
        device.setMediator(&mediator);

        // watchProperty way to detect new property
        device.watchProperty("text", [](INDI::PropertyText)
        {
            std::cout << "watchProperty" << std::endl;
        });

        device.registerProperty(text);
    }
    return 0;
}
```

#### Output:
```
$ ./main
watchProperty
```
`newText` is missing.
Is this behavior of the BaseMediator intentional?

pawel-soja:
@knro I have a big doubt about BaseMediator

#### Now is:
- BaseMediator::newProperty
- BaseMediator::removeProperty

is called when a property is defined/deleted and
- BaseMediator::newText
- BaseMediator::newSwitch
- BaseMediator::newBlob
- BaseMediator::newLight
- BaseMediator::newNumber

is called when a change comes, so it's very unreadable.

#### why not (long version):
- BaseMediator::defText
- BaseMediator::defSwitch
- BaseMediator::defBlob
- BaseMediator::defLight
- BaseMediator::defNumber
- BaseMediator::newText
- BaseMediator::newSwitch
- BaseMediator::newBlob
- BaseMediator::newLight
- BaseMediator::newNumber
- BaseMediator::removeText
- BaseMediator::removeSwitch
- BaseMediator::removeBlob
- BaseMediator::removeLight
- BaseMediator::removeNumber

long but consistent

#### or why not (simply):
- BaseMediator::defProperty(INDI::Property) // when created
- BaseMediator::newProperty(INDI::Property) // when value changed (confusing name with previous functionality)
- BaseMediator::removeProperty(INDI::Property) // when removed

sample names - exact function names to be determined

#### or
- BaseMediator::updateProperty(INDI::Property) // when created or value changed
- BaseMediator::removeProperty(INDI::Property) // when removed

BaseMediator is a key functionality at the moment and it would be nice to think about it and do it right.

knro:
```
BaseMediator::newProperty(INDI::Property) // when created
BaseMediator::updateProperty(INDI::Property) // when value changed
BaseMediator::removeProperty(INDI::Property) // when removed
```

This is perfect and concise. newXXX was there before INDI::Property was introduced, but now with INDI::Property, it is no longer needed. However, we can only remove in 2.0. For now you **add** updateProperty so that existing clients can transition to this and then in 2.0 you can remove.

knro:
@pawel-soja Let me know once you want me to review & test this again.

pawel-soja:
Okay, I made the changes according to the idea above. I don't have access to the hardware at the moment so I can't test it thoroughly. @knro if you have time would you take a look?

`indi_imager_agent`  is already using the new version of `BaseMediator`.
https://github.com/indilib/indi/blob/8a033ffc0138e867eb646efb340a6bbe5700e3ed/drivers/agent/agent_imager.h#L48-L50

`INDI::Property` no longer uses a pointer to `BaseDevice`
https://github.com/indilib/indi/blob/8a033ffc0138e867eb646efb340a6bbe5700e3ed/libs/indibase/property/indiproperty.h#L66-L69

knro:
Thanks! Absolutely, I'll test it now.

knro:
Found a regression, driver interface is missing

image

pawel-soja:
I was looking through the code on my phone and found this funny bug ;) Did you catch anything else?

knro:
Nothing yet. KStars otherwise appears to be OK. No deadlocks or crashes so far. Will continue testing.

pawel-soja:
> Found a regression, driver interface is missing
> 
> image

fixed

pawel-soja:
@knro the values that are displayed are quite magical. Is this intentional? Is it better to display what's going on in text.

knro:
@pawel-soja I'm not following, can you clarify?

pawel-soja:
I mean something like this:
image

knro:
No, the number is actually used by clients to know which interfaces are supported by the device. It has to be a number.

pawel-soja:
Okay I got it, I was confused that probably KStars displays the value from the `DRIVER_INFO->DRIVER_INTERFACE` property and uses `getDriverInterface` in the logic.
So `getDriverInterface` should always be based on the value from `DRIVER_INFO->DRIVER_INTERFACE`. I need to make corrections then.

knro:
It's already set in [setDriverInterface](https://github.com/indilib/indi/blob/56952b1bd419e00f4ce1d2370a5253d91f75e8f4/libs/indibase/defaultdevice.cpp#L802) here

pawel-soja:
I understand, but there is still a possibility that someone will type directly into the property bypassing the `setDriverInterface` function.
```cpp
auto driverInterface = getText("DRIVER_INFO")->findWidgetByName("DRIVER_INTERFACE");
if (driverInterface)
    driverInterface->setText("12345");
```
And the function will return an incorrect value:
https://github.com/indilib/indi/blob/56952b1bd419e00f4ce1d2370a5253d91f75e8f4/libs/indibase/defaultdevice.cpp#L796-L800

I suggest to remove `interfaceDescriptor` and always base on property.
```cpp
uint16_t BaseDevice::getDriverInterface() const
{
    D_PTR(const BaseDevice);
    auto driverInterface = getText("DRIVER_INFO")->findWidgetByName("DRIVER_INTERFACE");
    return driverInterface ? atoi(driverInterface->getText()) : 0;
}

void BaseDevice::setDriverInterface(uint16_t value)
{
    D_PTR(BaseDevice);
    auto driverInterface = getText("DRIVER_INFO")->findWidgetByName("DRIVER_INTERFACE");
    if (driverInterface)
        driverInterface->setText(std::to_string(value));
}
```

knro:
Sure, let's proceed with that.

pawel-soja:
@knro ready for review.

knro:
No issues detected so far! Thanks! Will report if I find any regressions.

PR #1770, opened 2022-11-27
Use modern INDI::PropertyXXX in INDI::Property

pawel-soja:
The most important changes:
- Deleted `void *BaseDevice::getRawProperty(const char *name, INDI_PROPERTY_TYPE type) const`
- Deleted `void Property::setProperty(void *p)` and `void *Property::getProperty() const`
- Deleted `void Property::setType(INDI_PROPERTY_TYPE t)`
- Different type returned by `INDI::Property::getXXX()` - `INDI::PropertyXXX` instead of `INDI::PropertyView<XXX>`

### DIFF
--- a/libs/indidevice/basedevice.cpp
+++ b/libs/indidevice/basedevice.cpp
@@ -129,12 +129,6 @@ IPerm BaseDevice::getPropertyPermission(const char *name) const
     return IP_RO;
 }
 
-void *BaseDevice::getRawProperty(const char *name, INDI_PROPERTY_TYPE type) const
-{
-    INDI::Property *prop = getProperty(name, type);
-    return prop != nullptr ? prop->getProperty() : nullptr;
-}
-
 INDI::Property BaseDevice::getProperty(const char *name, INDI_PROPERTY_TYPE type) const
 {
     D_PTR(const BaseDevice);
--- a/libs/indidevice/basedevice.cpp
+++ b/libs/indidevice/basedevice.cpp
@@ -478,9 +478,9 @@ bool BaseDevice::isConnected() const
     if (!svp)
         return false;
 
-    auto sp = svp->findWidgetByName("CONNECT");
+    auto sp = svp.findWidgetByName("CONNECT");
 
-    return sp && sp->getState() == ISS_ON && svp->getState() == IPS_OK;
+    return sp && sp->getState() == ISS_ON && svp.getState() == IPS_OK;
 }
 
 void BaseDevice::attach()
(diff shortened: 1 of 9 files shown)


pawel-soja:
@knro will you take a look?

knro:
Thanks. Looks like it returns non-existing properties filled with random data. This is from KStars debug:

image

pawel-soja:
Thank you very much!

A property with the `invalid` flag was returned.
It is characterized by being static and has empty fields and 16 widgets. When converting to a low-level pointer, it should be converted to `nullptr` so that it is compatible with the older implementation:
`kstars/indi/indiconcretedevice.cpp`
```cpp
INDI::PropertyView<ISwitch> *ConcreteDevice::getSwitch(const QString &name) const
{
    return m_Parent->getBaseDevice()->getSwitch(name.toLatin1().constData()); // convert INDI::PropertySwitch to pointer
}
```

I'll fix it in a moment.

pawel-soja:
@knro ready. I am impressed by your vigilance!

knro:
@pawel-soja Thanks for the quick fix! I will actually now test this live in the observatory just to check for any possible regressions.

knro:
I found more issues. I have these functions defined:

void propertyToJson(INDI::Property property);

this internally checks the type of the property and call another function

void propertyToJson(ITextVectorProperty *tvp);

It calls using property->getText();

However, this now just calls again the INDI::Property signature function which leads to an infinite loop. This would require changes to KStars but we're in code freeze now.

pawel-soja:
Ohhh, I see. I wonder how to ensure compatibility....

pawel-soja:
For some time, you can use `INDI::PropertyXXX` instead of `IXXXVectorProperty *`, so setting it to `2.0.0` makes the most sense. Thanks.

Let me know if you have any problems.

You made my day with this recursive problem... Hah

pawel-soja:
@knro Example of conversion from `INDI::PropertySwitch` to `INDI::Property` and back from `INDI::Property` to `INDI::PropertySwitch`.
```cpp
#include <libindi/indiproperty.h>
#include <libindi/indipropertyswitch.h>
#include <iostream>

void test(INDI::PropertySwitch)
{
    std::cout << __PRETTY_FUNCTION__ << std::endl;
}

void test(INDI::Property property)
{
    std::cout << __PRETTY_FUNCTION__ << std::endl;
    switch(property.getType())
    {
        case INDI_SWITCH:
            test(INDI::PropertySwitch(property));
        /* ... */
        default:;
    }
}

int main(int, char *[])
{
    // Create typed property
    INDI::PropertySwitch switchProperty(2);

    // Cast typed property to generic property
    INDI::Property genericProperty = switchProperty;

    // Test it!
    test(genericProperty);
    return 0;
}
```

Output:
```
void test(INDI::Property)
void test(INDI::PropertySwitch)
```

knro:
Ok let me test this thoroughly tomorrow (still running other tests in the observatory) and if it works well we can include this for 1.9.9

pawel-soja:
After changes to KStars it should be ok.
Still wondering how to make it backwards compatible.
This case is probably a checkmate :laughing: .
I'm super curious.

knro:
Thanks it works great now! Will make the necessary changes to KStars.

knro:
Ok I had to revert this. Turns out the amount of required changes is substantial indeed so this should be delay to INDI 2.0.0

pawel-soja:
I was a little surprised in the morning seeing that you put it in the master.
If you don't find a good solution, I'm thinking about introducing a typedef for prettier template names (WidgetView/PropertyView) and replacing the behavior at a later stage.

PR #1773, opened 2022-11-28
Refactor indiapi structures

pawel-soja:
It comes back like a boomerang.

The goal is to arrange the fields in the structures so that they are close to each other.
This will help with the property work to ultimately manage memory (which is a nuisance in blobs).

With this organization, I can introduce a more generic structure to the library and simplify some implementations.
```cpp
/* this is just an example */

typedef struct _IGeneric
{
    /** Helper info */
    void *aux;
    /** Pointer to parent */
    void *parent;
    /** Index name */
    char name[MAXINDINAME];
    /** Short description */
    char label[MAXINDILABEL];

    union
    {
        struct
        {
            /** Allocated text string */
            char *text;
        } itext;

        struct
        {
            /** GUI display format, see above */
            char format[MAXINDIFORMAT];
            /** Range min, ignored if min == max */
            double min;
            /** Range max, ignored if min == max */
            double max;
            /** Step size, ignored if step == 0 */
            double step;
            /** Current value */
            double value;
        } inumber;

        struct
        {
            /** Switch state */
            ISState s;
        } iswitch;

        struct
        {
            /** Light state */
            IPState s;
        } ilight;

        struct
        {
            /** Format attr */
            char format[MAXINDIBLOBFMT];
            /** Allocated binary large object bytes - maybe a shared buffer: use IDSharedBlobFree to free*/
            void *blob;
            /** Blob size in bytes */
            int bloblen;
            /** N uncompressed bytes */
            int size;
        } iblob;
    };
} IGeneric;

typedef struct _IGenericVectorProperty
{
    /** Helper info */
    void *aux;
    /** Device name */
    char device[MAXINDIDEVICE];
    /** Property name */
    char name[MAXINDINAME];
    /** Short description */
    char label[MAXINDILABEL];
    /** GUI grouping hint */
    char group[MAXINDIGROUP];
    /** Current property state */
    IPState s;
    /** Widgets comprising this vector */
    void *widgets;
    /** Dimension of widgets[] */
    int count;
    /** ISO 8601 timestamp of this event */
    char timestamp[MAXINDITSTAMP];
    /** Current max time to change, secs */
    double timeout;
    /** Client accessibility hint */
    IPerm p;

    union
    {
        struct
        {
            /** Switch behavior hint */
            ISRule r;
        } iswitch;
    };
} IGenericVectorProperty;
```

Apart from changing the location of the field data, the most significant change is the removal of `aux0, aux1, aux2` and leaving only `aux`.
This is not a problem with `INDI Core` and `INDI 3-rdparty`.
**A small problem arises with the `KStars` implementation** where `aux2` is used, among other things.

I leave the topic as a draft open for discussion.

### DIFF
--- a/drivers/video/v4l2driver.cpp
+++ b/drivers/video/v4l2driver.cpp
@@ -839,7 +839,7 @@ bool V4L2_Driver::ISNewNumber(const char * dev, const char * name, double values
 
         for (int i = 0; i < ImageAdjustNP.nnp; i++)
         {
-            unsigned int const ctrl_id = *((unsigned int *)ImageAdjustNP.np[i].aux0);
+            unsigned int const ctrl_id = *((unsigned int *)ImageAdjustNP.np[i].aux);
             double const value = ImageAdjustNP.np[i].value;
 
             LOGF_DEBUG("  Setting %s (%s) to %f, ctrl_id = 0x%X", ImageAdjustNP.np[i].name,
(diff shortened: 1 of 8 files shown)


knro:
Ok I think this would even lead to more significant changes for 2.0.0. There is already a large burden on 3rd party client code to be updated (if they happen to rely on INDI::BaseClient).

PR #1774, opened 2022-11-28
Refactor - use PropertyView/WidgetView

pawel-soja:
The main goal is to get rid of the use of direct access to fields defined in indiapi.h and the use of getters/setters via `PropertyView/WidgetView`.
It will also make it easier to switch to high-level `INDI::PropertyXXX`.

In addition, I added typedefs to make the names friendlier, and in the future, it could be easily replaced with full-fledged classes.
https://github.com/indilib/indi/blob/3f172f6052646941b6a533da29739a952e8a418e/libs/indidevice/property/indipropertyview.h#L38-L48


knro:
Quite impressive!! I applaud you for taking the time to do this. I think from now on, I will not accept MRs for new drivers that use the old-style INDI properties.

pawel-soja:
Hey thanks! Finally, I want to mark a couple of methods as deprecated and that there will be no warnings related to this in this repository.

PR #1776, opened 2022-11-30
Add libindiclient shared library

pawel-soja:
Solving the problem of building a deb package.

### DIFF
--- a/debian/libindi-dev.links
+++ b/debian/libindi-dev.links
@@ -1,2 +1,3 @@
+usr/lib/libindiclient.so.1 usr/lib/libindiclient.so
 usr/lib/libindidriver.so.1 usr/lib/libindidriver.so
 usr/lib/libindilx200.so.1 usr/lib/libindilx200.so
--- a/debian/libindi1.install
+++ b/debian/libindi1.install
@@ -1,3 +1,5 @@
+usr/lib/*/libindiclient.so.1
+usr/lib/*/libindiclient.so.1.9.*
 usr/lib/*/libindidriver.so.1
 usr/lib/*/libindidriver.so.1.9.*
 usr/lib/*/libindilx200.so.1


pawel-soja:
`indiclient.so` never built before, when I added options to build static and shared libraries, by default the shared library is additionally built. From what I can see, other libraries in `INDI Core` want to link it instead of static.

pawel-soja:
Ahhh fail. I will add package build testing to CI later.

knro:
indiclient library now fails to build on KDE Craft build system due to missing zlib.h

I added ZLIB_INCLUDE_DIR to a couple of places, but it still fails to find the file. CMake reports that zlib was found in the system, so that was not an issue.

```
D:\CraftRoot\build\_\dec7706e\indiclient\libs\indicore\indililxml.h(488): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data
[32mD:\CraftRoot\dev-utils\bin\ninja.exe: b"D:\\CraftRoot\\build\\_\\dec7706e\\indiclient\\libs\\indicore\\indililxml.h(488): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data"[0m
D:\CraftRoot\build\_\dec7706e\indiclient\libs\indidevice\basedevice.cpp(42): fatal error C1083: Cannot open include file: 'zlib.h': No such file or directory
[32mD:\CraftRoot\dev-utils\bin\ninja.exe: 
```
Any ideas?

pawel-soja:
I think I know what it is. Give me 1 hour and I'll be at the computer and upload the fix.

pawel-soja:
You were on the right track! `indidevice` compiles as an object library and required the `zlib.h` header to compile as shown by the error. To compile objects, linking of libraries is not required, but access to headers is required.

Following your lead, I put a similar commit on master.

Let me know if anything else comes up.

PR #1785, opened 2022-12-14
Shelyak spectrograph

eco747:
We would like to add a new type of driver: spectrograph.
Then Shelyak driver


knro:
Thank you! INDI already includes a [JSON parser](https://github.com/indilib/indi/blob/master/libs/json.h), so no need to add an additional JSON parser here.

eco747:
ok, but is my merge will be accepted ?
i will change that after merge.

knro:
but why? you can remove json.* now and rebuild the driver against INDI library's json. No need to include files to be deleted later.

knro:
btw, you can make any number of changes to the PR. You don't need to make a new PR.

eco747:
done :)
i remove all old custom json & used indi json lib.

knro:
@ecochard I just pushed some changes to the driver to make it complaint with INDI standards. Please check it out and re-test as I made substantial changes that might have broke the driver.

PR #1786, opened 2022-12-14
INDI v2.0.0 changes

knro:
Changes required for INDI v2.0.0 package.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,10 +44,10 @@ endif()
 # ####################################  INDI version  ################################################
 # N.B. DO NOT Forget to update version also in indiapi.h
 # Proper way is to use indiversion.h.cmake file but this would break make existing applications so let us stick to the old proven way
-set(INDI_SOVERSION "1")
-set(CMAKE_INDI_VERSION_MAJOR 1)
-set(CMAKE_INDI_VERSION_MINOR 9)
-set(CMAKE_INDI_VERSION_RELEASE 9)
+set(INDI_SOVERSION "2")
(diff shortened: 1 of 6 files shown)


knro:
@pawel-soja Do we need to start removing deprecated interfaces for 2.0.0?

pawel-soja:
You mean `BaseMediator` to remove fragments marked for version 1.x.x? I know that was the intention, but this will force all clients to be updated, including those written in Python. I'm thinking about an additional option when compiling, so that I can use the old interface when there will be an armageddon on the forums about "why it doesn't work for me" - extend the time of reimplementation of clients.

I see that after your changes not all tests/builds pass. You can push changes to branch 2.0.0. officially working on it. I'll try to make the appropriate changes.

knro:
Yes of course it would be nice if we can somehow maintain backward compatibility while issuing deprecation warnings. I'll close this PR for now until you find a better solution to handle backward compabitlity.

pawel-soja:
@knro I am preparing changes, please answer the above questions.

Thanks!

knro:
You mean about maintaining backward compatibility in 2.0.0?

pawel-soja:
Here, my point is that with subsequent versions, you do not have to edit and replace so many files. I am currently making changes and deprecated markings depending on the compiled version. I propose to allow compilation of versions 2 and 1 in the future.

knro:
I see, so the library won't support both at run time, only by compiling it with flags?

pawel-soja:
Yes, for example by changing version 2.0.0 to 1.10.0, you will be able to compile a library compatible with version 1. Now, I would like to understand the changes you made to this pull request.

PR #1794, opened 2022-12-23
Fix building in Cygwin

gitplcc:
Fix building:
- Static and dynamic  link libraries (except Qt client).
- Drivers (except V4L2).
- Server.
- Unit and integration tests.

PR #1795, opened 2022-12-24
Fix build on non-Linux

sthibaul:


sthibaul:
(@pinotree for information)

knro:
Thanks, this is related to #1794 so need to wait until that is tested and merged first. Is there any overlap?

sthibaul:
In libs/indiclient/baseclient.cpp, yes, but we used exactly the same fix.

knro:
Please rebase

sthibaul:
Here it is.

PR #1804, opened 2023-01-02
Fix warnings for Windows

pawel-soja:


knro:
@pawel-soja 

Found one issue when compiling with KStars under Windows. `mode_t` is defined twice, in indiutility.h and also in win32-msvc/qplatformdefs.h. In there they just define it directly:

`typedef int mode_t;`

What's the recommended method to ensure it's not defined again? is this something that needs to be done at INDI or client level?

PR #1805, opened 2023-01-02
Deprecated use of Property/BaseDevice as pointer.

pawel-soja:
Changes are backwards compatible.

It is recommended to refer directly to `INDI::PropertyXXX` and `INDI:BaseDevice` without dereferencing.
The project source codes have been corrected to use the correct form of object reference.


knro:
Awesome!! We'd probably need something similar in 3rd party, but it's a LOT of work to convert everything to use modern INDI properties.

PR #1812, opened 2023-01-19
Refactor INDI::GPS to be inline with INDI 2.0.0

tstibor:


tstibor:
Just checked indi-3rdparty for `$ grep -r "LocationN\[" .` looks like this PR will break all driver using _LocationN_ rather than _LocationNP_.  Is there any name convention, e.g xxxxN for numbers, xxxxS for switches, etc... or xxxxNP for numbers?

knro:
Yes. all properties end with the following.

+ Switch: SP
+ Number: NP
+ Light: LP
+ Text: TP
+ Blob: BP

With INDI v2.0.0 we are moving away from using old style properties (i.e. INumberVectorProperty..) to new style property (i.e. INDI::PropertyNumber) so moving forward we won't have INumber, ISwitch, IText..etc declared in the header file like what current drivers do.

tstibor:
Thanks for the information Jasem. I checked this PR against _indi-3rdparty_ and one needs to change all drivers
inheriting from _INDI::GPS_, eg. _LocationN_ -> _LocationNP_ etc... and additional modifications. Probably it would make sense
to add to this PR a kind a compatability reference variables with the variable names/types used  _indi-3rdparty_. Then GPS code in  _indi-3rdparty_ would not break and one could step by step migrate to 2.0.0?

PR #1840, opened 2023-02-05
Support for XISF files

nouspiro:
To test this you need to compile and install LibXISF https://gitea.nouspiro.space/nou/libXISF

### DIFF
--- a/cmake_modules/FindLibXISF.cmake
+++ b/cmake_modules/FindLibXISF.cmake
@@ -1,0 +1,23 @@
+find_path(LibXISF_INCLUDE_DIR
+  NAMES libxisf.h
+)
+
+find_library(LibXISF_LIBRARY
+  NAMES libXISF XISF
+)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(LibXISF
+  FOUND_VAR LibXISF_FOUND
+  REQUIRED_VARS
+    LibXISF_LIBRARY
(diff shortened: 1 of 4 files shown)


nouspiro:
Currently it support only LZ4+SH compression that is enabled when you enable compression in INDI panel. Not sure how to implements selection of all compression codecs no compression, zlib, lz4, lz4hc. Then there is also option for byte shuffling which increase it to 7 options.

knro:
I think that's very good for first release. We should keep it simple and probably just pick out the most optimal configuration for the compression we believe is suitable for our users.

nouspiro:
From my testing LZ4+SH is fastest option that gives good results. Maybe add environment variable that override compression? I will add CLI tool to LibXISF repo to recompress any XISF files.

knro:
Thanks, looks like we need to add libxisf to the PPA as well so this can get built. Can you add debian directory with the necessary files for the library? and I'll then see if I can get a PPA build going for it. then we need to add it to INDI control file as a new depedency.

nouspiro:
Ok done

knro:
Thanks, we have it now in the [Nightly PPA](https://code.launchpad.net/~mutlaqja/+recipe/libxisf-daily).

Can you please add libexif-dev to debian/control as well so that INDI pulls this now.

nouspiro:
You mean into indi/debian/control like this?

knro:
Yes! Thank you.

PR #1847, opened 2023-02-13
Feature/xisf as plugin

nouspiro:


knro:
Can you please check #1848? I think there is a regression there.

nouspiro:
Yes it was. Now it should behave better that it doesn't stop at first bad value but will continue to write other values.

knro:
Can you please check the debian packaging error?

nouspiro:
Wrapper lib doesn't get build because it can't find LibXISF. Most likely github build doesn't have it? It created build package when I was doing it locally with debuild.

```
dh_install: warning: Cannot find (any matches for) "usr/lib/*/libindixisf.so.*" (tried in ., debian/tmp)

dh_install: warning: libindi-xisf missing files: usr/lib/*/libindixisf.so.*
dh_install: error: missing files, aborting
```

knro:
@pawel-soja Do we need to update the packaging workflow to include libxisf? It's built now in the Ubuntu PPA.

pawel-soja:
I'm analyzing the case now. I probably need to add packages to dockerfiles.

pawel-soja:
I see I missed something :) Is the "libxisf" library available as a package for debian/ubuntu etc? Should I prepare an additional build for docker from sources https://gitea.nouspiro.space/nou/libXISF ?

nouspiro:
Not yet. I want to package it for debian but that will probably take quite some time. BTW I am currently rewriting libxisf so it doesn't require Qt.

knro:
It is actually already included in the [stable](https://code.launchpad.net/~mutlaqja/+recipe/libxisf-stable) and [nightly](https://code.launchpad.net/~mutlaqja/+recipe/libxisf-daily) PPAs.

PR #1852, opened 2023-02-17
Feature/qhy gps timestamp

nouspiro:
Pass exposure start timestamp from GPS into SER timestamp.


knro:
Can this be generalized to any GPS device and not QHY specific?

nouspiro:
I moved that QHY specific offset to qhy driver.

PR #1854, opened 2023-02-20
Re-do Gemini SetTrackMode and SetTrackEnabled

const86:
The old code says to change track mode we set the 130 property to
(131 + mode) value. That is not correct. Instead we should set the
(131 + mode) property with no value. I validated this change by
monitoring the track mode with handset.

Disabling tracking with parking sleep makes sense but seems to
interfere with INDI's state management badly. I did not figure out
what exactly was wrong.
This bug looks relevant: https://bugs.kde.org/show_bug.cgi?id=438995
Inspired by how other LX200 flavors work (AP for example), I disable
tracking by engaging the Terrestrial mode. This is also what user
is suggested to do manually if they want to stop tracking. Validated
in KStars by un/parking, toggling tracking, and slewing in different
permutations.

### DIFF
--- a/drivers/telescope/lx200gemini.cpp
+++ b/drivers/telescope/lx200gemini.cpp
@@ -2098,41 +2098,13 @@ bool LX200Gemini::setGeminiProperty(uint32_t propertyNumber, char* value)
 
 bool LX200Gemini::SetTrackMode(uint8_t mode)
 {
-    int rc = TTY_OK, nbytes_written = 0;
-    char prefix[16] = {0};
-    char cmd[16] = {0};
-
-    snprintf(prefix, 16, ">130:%d", mode + 131);
-
-    uint8_t checksum = calculateChecksum(prefix);
-
-    snprintf(cmd, 16, "%s%c#", prefix, checksum);
-
-    LOGF_DEBUG("CMD: <%s>", cmd);
-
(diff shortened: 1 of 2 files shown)


knro:
@amendolajamie Can you please check this change?

amendolajamie:
Using the old code, I can reproduce the issue using L5 firmware on a gemini 2, but can not reproduce the issue on L6 firmware using a gemini 2. 


For Gemini 2 the command set can be found here: https://gemini-2.com/web/L5V2_1serial.html  The L6 command set is unchanged.

Tracking rates:
130, 131..137 |   |   | 131: Sidereal132: King Rate133: Lunar134: Solar135: Terrestrial Mode136: Closed Loop137: Comet/User Defined | Tracking Rate. 130 can be used for requesting.
-- | -- | -- | -- | --

There is also the "sleep" command 
:hN# |   | Sleep Telescope: stop tracking, blank displays.
-- | -- | --
:hW#	 	Wake Up Telescope, resume tracking.

If we are going to use "terrestrial" I think that it should be added to the tracking rate property, and not used to start stop tracking. Instead maybe Sleep/Wakeup would be a better fit, although not perfect as it turns off the display. 

I will check to see why the old method works with L6 firmware and does not with L5. 

Jamie

amendolajamie:
This looks good, thanks! The relevant bug seems to be an issue with ekos, not with the gemini driver.

Jamie

PR #1860, opened 2023-02-27
Adding CFITSIO to cmakelists for indi XISF and changing linking step …

rlancaste:
…to make it work better on MacOS

### DIFF
--- a/libs/indixisf/CMakeLists.txt
+++ b/libs/indixisf/CMakeLists.txt
@@ -8,9 +8,9 @@ find_package(LibXISF)
 find_package(Qt5Core)
 find_package(CFITSIO)
 
-include_directories(${CFITSIO_INCLUDE_DIR})
-
 if(LibXISF_FOUND AND Qt5Core_FOUND AND CFITSIO_FOUND)
+
+	include_directories(${CFITSIO_INCLUDE_DIR})
 
     add_library(${PROJECT_NAME} SHARED
         indixisf.cpp
--- a/libs/indixisf/CMakeLists.txt
+++ b/libs/indixisf/CMakeLists.txt
@@ -6,11 +6,11 @@ include_directories(../indicore)
 
 find_package(LibXISF)
 find_package(Qt5Core)
+find_package(CFITSIO)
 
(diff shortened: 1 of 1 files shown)


knro:
Apparently this caused MinGW to fail?

rlancaste:
The following REQUIRED packages have not been found:
[77](https://github.com/indilib/indi/actions/runs/4287243289/jobs/7467868579#step:4:78)

[78](https://github.com/indilib/indi/actions/runs/4287243289/jobs/7467868579#step:4:79)
 * CFITSIO, A library for reading and writing data files in FITS (Flexible Image Transport System) data format, <http://heasarc.gsfc.nasa.gov/fitsio/fitsio.html>
[79](https://github.com/indilib/indi/actions/runs/4287243289/jobs/7467868579#step:4:80)
   Provides INDI with FITS I/O support.
[80](https://github.com/indilib/indi/actions/runs/4287243289/jobs/7467868579#step:4:81)

[81](https://github.com/indilib/indi/actions/runs/4287243289/jobs/7467868579#step:4:82)

rlancaste:
find_package(CFITSIO REQUIRED)
 
Maybe it should not be required?

rlancaste:
I am not sure why this would still come up as being CFITSIO required when I removed the requirement

knro:
@pawel-soja Any ideas what's causing the MinGW failure? Also Debian packages have been failing for quite a while now after adding libxisf. I updated the dependencies, but it had no effect. The docker containers needs to be updated?

pawel-soja:
I'll check it.

pawel-soja:
https://github.com/indilib/indi/blob/754d269598067861a36ff41bd66f3b9934757518/CMakeLists.txt#L145

FeatureSummary has a `TYPE REQUIRED` for this package. So if there is a reference to `find_package(CFITSIO)` somewhere, it will require the package.

If `CFITSIO` is to be optional, I suggest changing `TYPE REQUIRED` to `TYPE OPTIONAL`

knro:
This is no longer required after removing Qt5 dependency

PR #1865, opened 2023-03-08
Add driver for myDCP4ESP32 Dew Contoller

hcomet:
The myDCP4ESP32 Dew Controller is a DIY dew controller developed by Robert Brown. This is a driver implementation for it based on the MyFocuserPro2, USB_Dewpoint and SQM drivers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,6 +93,7 @@ OPTION(INDI_BUILD_WEBSOCKET "Build INDI with Websocket support" OFF)
 OPTION(INDI_FAST_BLOB "Build INDI with Fast BLOB support" ON)
 OPTION(INDI_BUILD_SHARED "Build shared library" ON)
 OPTION(INDI_BUILD_STATIC "Build static library" ON)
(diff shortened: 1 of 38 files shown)


knro:
Thank you! We also need to add documentation for this in indilib.org website device section.

hcomet:
I tried to submit documentation but I could not submit the HTML. I was also limited to uploading 3 images. Could not add the 4th.

I've attached the HTML file and the 4th image here.
[MyDCP4ESP32.zip](https://github.com/indilib/indi/files/10938359/MyDCP4ESP32.zip)

knro:
Thanks, I've added all the images. Are there any images for the actual final controller? Couldn't find images online.

hcomet:
I've attached a picture of my controller build with the PCB as a background. Hopefully its usable.

I also noticed that I had copyright symbol in the description that is no displaying properly. Looks like I left the semicolon off of it in the HTML. There's a "&copy Robert Brown" that should be "&copy; Robert Brown". It displayed proper when loading locally in a browser but doesn't display properly on the website. If possible could you update it?

Thanks.
myDCP4ESP32_controller

PR #1867, opened 2023-03-10
Remove XISF wrapper and adjust changes for LibXISF 0.2.0

nouspiro:


libs/indibase/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 3.13)
 project(indidriver C CXX)
 
+set(CMAKE_CXX_STANDARD 17)
+
```
knro, line 5: This should be specified in cmake_modules/CMakeCommon.cmake line 14

knro:
Ok looks like std::random_shuffle needs to be replaced as well by std::shuffle

nouspiro:
Hm just found out that std::from_chars is not implemented on older gcc versions. Gotta replace it with something else.

nouspiro:
Wanted to use std::from_chars and std::to_chars to get locale independed string->float conversion but it is not implemented properly. I see that indi use AutoCNumeric to ensure correct conversion so I will use that.

PR #1871, opened 2023-03-16
Fixed SestoSenso2 calibration

Frostregen:
This contains 2 fixes:

1. A rather small JSON fix for goOutToFindMaxPos command
2. A more complex UI flow issue

The UI flow issue occured, when you needed to press the "CALIBRATION_NEXT" button for the second time. Apparently clicking the button for a second time de-selected it, causing the call in line 514 of sestosenso2.cpp to return -1:
auto current_switch = CalibrationSP.findOnSwitchIndex();

Which in turn caused a crash in the followup array access.
The proposed fix is to turn the button group into a radiogroup ( ISR_1OFMANY ), which will not de-select the button, thus returning the wanted index.

Calibration worked fine after those 2 fixes on my own SestoSenso2.

Greetings,
Frostregen

PS: SestoSenso2 has the latest firmware: 03.05.01

### DIFF
--- a/drivers/focuser/primalucacommandset.cpp
+++ b/drivers/focuser/primalucacommandset.cpp
@@ -402,7 +402,7 @@ bool SestoSenso2::storeAsMinPosition()
 *******************************************************************************************************/
 bool SestoSenso2::goOutToFindMaxPos()
 {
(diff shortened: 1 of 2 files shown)


knro:
Thank you!

PR #1874, opened 2023-03-21
Update skywatcherAPIMount.cpp

mwillis87:
- Changed debug output datatypes for freq (Lines 1148 & 1197) and Rate to floats for both Axis1 and Axis2.
- Used std::fabs instead of std::abs (Lines 1137 & 1185) due to the previous change of TrackingRate[AXIS]  to a double.


knro:
Thanks, any luck fine-tuning PID to get some proper tracking? I feel we might be missing something here?

PR #1880, opened 2023-03-27
Add Flip functionality

amendolajamie:
Add flip functionality while maintaining compatibility. Discussion here: https://github.com/indilib/indi/discussions/1875#discussioncomment-5437822

There is a corresponding change in ekos to call flip when doing meridian flips. 
https://invent.kde.org/education/kstars/-/merge_requests/857


amendolajamie:
OK, I'm reviewing the update in inditelescope.cpp, and am confused by the original code. It looks like it has a bug with setting the values. Please look at the inline comment.

sterne-jaeger:
Now it looks good, we could go for a merge.

knro:
How about other scope that can flip? EQMod can flip but it is not marked as such. Is there a definition for CAN_FLIP? does it mean the mount has explicit flip support vs. passive flipping when GOTO falls outside some range?

sterne-jaeger:
Valid point, Jasem. As far as I know most of modern mounts have it in some way configurable whether mounts should flip. This new flip function introduced here follows the Gemini style which have a dedicated firmware command for a slew combined with a forced flip.

Other mounts like SkyWatcher or Avalon have this separated. There is a setting for meridian flips (none, auto, forced) and the mount firmware decides for a new Goto command whether it should change pier side or not.

In EKOS and its meridian flip implementation we always assume that the mount changes automatically the pier side when the mount crossed the meridian. It seems that some mounts don't do this reliably (e.g. Gemini) and therefore it would be better if we explicitly have the option to force a flip combined with a GOTO.

To be honest I'm very to this since the client software partially will take over the responsibility which pier side is appropriate - which could create new problems. But I understand that some mounts don't handle automatic flips reliable and therefore need explicit commands.

That's why I would say let's try this with Gemini only and be careful with down porting this to other mounts. Nevertheless, if we never try it, we never know if it is a better way...

amendolajamie:
The new capability:  TELESCOPE_CAN_FLIP will allow drivers to choose if it makes sense to do a special command to flip. Existing drivers will remain unchanged unless it makes sense to do a specialized flip. In the case of ekmod, or Avalon, if the existing implementation works perfectly fine, then it's not needed, but if it is more reliable, then the driver can be updated to claim the TELESCOPE_CAN_FLIP capability and do whatever is needed to make the flip. I believe that eqmod would make the command forced flip, then call the goto. The "forced flip" command only forces the next goto to flip then resets back to normal.

PR #1889, opened 2023-04-17
Enable testing from the top-level build directory

panicgh:
Users currently need to manually execute the unit and integration tests separately from their respective subdirectories, and it is not possible to run both from a single command.

This commit makes the tests runnable also from the top-level build directory (e.g. with `make test`).


knro:
Thank you! We need more tests

Issue #1842, opened 2023-02-05
Driver "indi_lacerta_fmc_focus" (!) (for Firmware version 3.1): relative positioning inward causes out of range position


timpeKH:
When the driver "indi_lacerta_fmc_focus" (indilib1 2.0.0)  runs on a remote server and a client connects to it from another machine via network,
then a relative positioning command in inward direction ("focus in") causes a command (according to log: DEBUG	232.861839 sec	: CMD [: M -1 #]). This command is ignored by the device itself  and interestingly the Lacerta MFOC hand control panel is not affected. But the absolute position within the _INDI control panel_ gets wrong (4294967295 is 0xFFFFFFFF is -1). 
This yields an inability to command any other position unless the "sync" command is used to correct the wrong absolute position. 

_It is probably relevant to note that this issue does not occur when the device is connected locally and the device driver is run as a local service._

It is further of interest to note that the root cause of this bug is probably also responsible for another bug for another driver.
Pls. see also [https://github.com/indilib/indi/issues/1834](https://github.com/indilib/indi/issues/1834) for the driver "indi_lacerta_mfoc_focus". As opposed to this bug the driver "indi_lacerta_mfoc_focus" seems to crash completely due to the negative number in the position command. Further the firmware of the latter driver is not capable to re-connect until the hardware hand control panel was switched off and on again. 

**To Reproduce**
1. Run driver  "indi_lacerta_fmc_focus" on a remote server (e.g. Astroberry or Rasbberry/Ubuntu server running an indiserver)
2. Connect from a local client to the remote indiserver via KStars -> Tools -> Device -> Device Manager -> Client => external indiserver
3. Open INDI Contorl Panel
4. Enter relative position command as in the following screenshot (e.g. 500 steps or any other sensible number without making the hardware running into motion limits):
![Step_01_Screenshot_2023-02-05_16-59-36](https://user-images.githubusercontent.com/124307800/216837349-4da5b0d8-e05d-48e1-8bac-f8e6a511c2f5.png)
5. The result should look as on the following screenshot - **note the absolute position (!)**:
![Step_02_Screenshot_2023-02-05_16-57-55](https://user-images.githubusercontent.com/124307800/216837375-72b5178f-da51-4f1f-affd-7302255c60df.png)
 

**Expected behavior**
Correct inward position movement of focuser device

**Desktop (please complete the following information):**
[...]

Issue #1841, opened 2023-02-05
Driver "indi_lacerta_fmc_focus" (!) (for Firmware version 3.1) hangs up when disconnecting @ 100% kernel load


timpeKH:
For the Lacerta MFOC a new firmware 3.1was introduced together with a new driver "indi_lacerta_mfoc_fmc_focus"
with libindi1 version 2.0.0. According to Lacerta this driver is specific for the Lacerta MFOC with firmware version 3.1.
The driver hangs up when disconnecting in the INDI control panel. Even though all other hardware drivers of the indiserver process are disengaged the "indi_lacerta_mfoc_fmc_focus" remains active and generates a high processor load of 100% for one kernel. Establishing a connection to the Lacerta MFOC is not possible in this state. The driver process must be terminated manually using bash/kill or other tool and restarted before a connection to the device is possible again.
The behaviour can be reproduced.

Unlike the driver "indi_lacerta_mfoc_focus" for older firmware versions (which crashes when doing relative position command in **inward** direction -> causes negative position number which causes firmware to crash, see [https://github.com/indilib/indi/issues/1834](https://github.com/indilib/indi/issues/1834)), the device does not need to be switched off and on to be able to reconnect.

Although the driver log files in debug mode did not show anything helpful from my naive perspective I attached a log file.

**To Reproduce**
Exact steps to reproduce the behavior.
1. Run driver indi_lacerta_mfoc_fmc_focus in the KSTARS -> Tools -> Device -> Device Manager -> [ Runs as local service or server]
2. Alternatively to step 1. the behaviour is the same on a remote server (e.g. a Raspberry)
3. Connect to "Lacerta MFOC FMC".
4. Disconnect from "Lacerta MFOC FMC".
5. Check the processes with a suitable tool (e.g. htop). One can see the process "indi_lacerta_mfoc_fmc_focus" run on one kernel usually at 100% load (for the specific kernel).
6. Kill the process "indi_lacerta_mfoc_fmc_focus" manually in htop or via bash

**Expected behavior**
Client disconnects with driver/server remaining responsive at low processor load waiting for a new connection

**Screenshots**
![Screenshot_2023-02-05_16-45-44](https://user-images.githubusercontent.com/124307800/216835366-a4e100e8-6373-461e-854b-3e5c1f21220a.png)
![Screenshot_2023-02-05_17-50-50](https://user-images.githubusercontent.com/124307800/216835625-b546c966-811e-4383-8cdb-83b800829a8f.png)

[...]

Issue #1834, opened 2023-02-02
Lacerta MFOC Driver Hangs Up When Relative Positioning


timpeKH:
When I issue a relative position command in the INDI Control Panel, the Lacerta MFOC reacts with a
strange and wrong absolute position (huge, unrealistic number). Further the MFOC stops any reaction and communication with the INDI-Client. No disconnect or re-connection is possible unless the Lacerta MFOC handbox is switched off and restarted.

**Setup and Software Versions Used**
I use a Lacerta MFOC with firmware version 3.03. The MFOC is connected via USB to a Raspberry Pi 4, 8GB running Ubuntu Server arm64, 22.04 LTS (Jammy) as a INDI-host. So the Rapsberry Pi is where the indiserver runs. As clients EKOS or CCDCiel were used on a Laptop under XUbuntu 22.04. LTS ( KSTARS version 3.6.2 Stable Build 2022-11-30T18:14:22Z)
The driver package used on the RPI4 which provided the Lacerta MFOC driver (AFAIK) was/is : 
Package: libindi1
Version: 1.9.9+202211300913~ubuntu22.04.1

**To Reproduce**
1. Connect Lacerta MFOC Handbox with firmware 3.03 via USB and power on
2. Start indiserver and run driver the process for Lacerta MFOC: indi_lacerta_mfoc_focus along with a CCD simulator driver or any minimum configuration to use an indi client and the INDI Control Panel (My indiserver runs on a RPI4, 8GB with Ubuntu Server 22.04 LTS (Jammy). 
3. connect to indiserver via EKOS client (I connect via another Laptop to the RPI4) and start INDI Control Panel
4. In the INDI Control Panel go to the tab for the "Lacerta MFOC" and there to the "Main Control" tab. Press "connect" to connect to the Lacerta MFOC.
5. First set a "Absolute Position" value (e.g. 3000) and send the command by pressing the "set" button
7. Next set a "Relative Position" value (e.g. 1000) and send the command by pressing the "set" button
8. Check what happens to the feedback status value of "Absolute Position". In my case a huge integer number appears while the "Relative Position" command does not yield any focuser motion. Now the focuser should be completely non-reactive.
9. Try to shut down the indiserver process
10. Check if the Lacerta MFOC driver process "indi_lacerta_mfoc_focus" was stopped together with the indiserver. In my case this process keeps running until terminated by a kill signal.

[...]

PR #1894, opened 2023-05-07
Bugfix/mfoc rel movement error

chkettu:
fix for #1834, #1842 (possibly #1841)


drivers/focuser/lacerta_mfoc.cpp
```diff
@@ -445,12 +437,22 @@ IPState lacerta_mfoc::MoveAbsFocuser(uint32_t targetTicks)
 IPState lacerta_mfoc::MoveRelFocuser(FocusDirection dir, uint32_t ticks)
 {
     // Calculation of the demand absolute position
-    uint32_t targetTicks = FocusAbsPosN[0].value + (ticks * (dir == FOCUS_INWARD ? -1 : 1));
+    uint32_t targetTicks = 0;
+    if (dir == FOCUS_INWARD) {
+      targetTicks = FocusAbsPosN[0].value - ticks;
+    } else {
+      targetTicks = FocusAbsPosN[0].value + ticks;
+    }
```
knro, line 445: How is this any different from the line you removed above?

drivers/focuser/lacerta_mfoc.cpp
```diff
@@ -474,15 +476,18 @@ uint32_t lacerta_mfoc::GetAbsFocuserPosition()
 
     int nbytes_written = 0;
     int nbytes_read = 0;
+    int count = 0;
+
+    tty_write_string(PortFD, MFOC_cmd, &nbytes_written);
+    LOGF_INFO("CMD <%s>", MFOC_cmd);
```
knro, line 482: LOGF_DEBUG should be used. LOGF_INFO is for informational messages, not debug.

drivers/focuser/lacerta_mfoc.cpp
chkettu: The ole line worked on my 32b Linux-Notebook, but not on a 64b-Raspberry PI where it results to "-1" targetTicks. I splitted the line for better debugging options - and with this change it gives the right results. Don't know why...

knro: I think a safer way is this:

auto targetTicks = std::clamp(FocusAbsPosN[0].value + (ticks * (dir == FOCUS_INWARD ? -1 : 1)), FocusAbsPosN[0].min, FocusAbsPosN[0].max)

drivers/focuser/lacerta_mfoc.cpp
chkettu: changed to debug

drivers/focuser/lacerta_mfoc.cpp
chkettu: Cool - I'll change this today evening.

chkettu: change commited - plz review again

PR #1897, opened 2023-05-08
Esatto: Port state tracking from EsattoArco

const86:
Before this change Ekos is confused when exactly the focuser completes requested move
and what is the end position. AF Overscan feature is totally unusable. Without overscan
autofocus sort of works, but often ends a bit off the requested position. It also does not
quite return to the start position in case of an error. With this change all above is fine.
Temperature and voltage are ported too.


knro:
Alright so the timers were not updating the state machine correctly?

const86:
The timers are redundant, because there is already one in base class and EsattoArco is fine with just that one. The real problems lie in what happens in timers callbacks:

1. The `TimerHit` and the dedicated motion timer callback interfered with each other.
2. Under condition that we were busy and the focuser has just completed the movement, we should update the position again, otherwise the value may be just a bit outdated.

So, the change includes some refactoring, thus it is bigger than needed to fix the actual misbehavior.

PR #1900, opened 2023-05-23
Moonlite focuser : basic find home implementation

gehelem:
Hi
i just added à "go home" property, sending Moonlite :PH01# command
(copied/pasted from Myfocuserpro2)
i don't know if this one is handled by original Moonlite focuser, but i think it's convenient to have it available for hacked Arduino focusers.
G.


knro:
It's not handled by original MoonLite unless there has been a firmware update to say otherwise?

gehelem:
This command is described here :
https://indilib.org/media/kunena/attachments/1/HighResSteppermotor107.pdf
but it's not clear if it's specific to DRO focuser

PR #1903, opened 2023-05-30
Add one more check to ensure minimum count is met

knro:
When sending commands, sometimes we receive response from another command in the buffer despite flushing the buffer. So in case the response does not meet the minimum count then we retry.

### DIFF
--- a/drivers/telescope/ioptronv3driver.cpp
+++ b/drivers/telescope/ioptronv3driver.cpp
@@ -73,7 +73,12 @@ bool Driver::sendCommandOk(const char *command)
     return false;
 }
 
-bool Driver::sendCommand(const char *command, int count, char *response, uint8_t timeout, uint8_t debugLog)
+bool Driver::sendCommand(const char *command,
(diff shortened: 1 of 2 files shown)


JulesArcher:
Hi @knro I am also having the same error message when using Wifi  connection on my CEM26 as reported earlier:
 
[2023-09-25T18:01:01.845 PDT INFO ][ org.kde.kstars.indi] - iOptron CEM26 : "[ERROR] Error setting RA/DEC. "
[2023-09-25T18:01:46.918 PDT INFO ][ org.kde.kstars.indi] - iOptron CEM26 : "[ERROR] bool IOPv3::Driver::getCoords(double*, double*, IOPv3::IOP_PIER_STATE*, IOPv3::IOP_CW_STATE*): Expected 20 bytes but received 23. "

Is there a modified driver with the fix that I could test on my system?  

Thanks!

knro:
Can you please check both wired and wireless connections?

Issue #1528, opened 2021-08-29
libindiclient leaks last blob on driver restart


pludov:
Upon removal of devices or property, blob data are not freed. As a result, a memory leak of blob will occurs when a driver that has received blob is restarted:

1. Compile the dummy-streamer example program joined
2. Start an indi server  with ccd_simulator
3. Start dummy-streamer
4. Activate streaming on the ccd_simulator: dummy-streamer logs it blob receptions
5. Restart the ccd_simulator (through indi fifo)
6. Restart streaming on the ccd_simulator

Repeating 5&6 will make dummy-streamer program memory size. You can use bigger frame size to make the problem more obvious. Alternatively, running dummy-streamer with valgrind will report the following leak:

``` 
==26462== 1,328,925 bytes in 1 blocks are definitely lost in loss record 28 of 28
==26462==    at 0x4C33D2F: realloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==26462==    by 0x11FCC3: INDI::BaseDevice::setBLOB(_IBLOBVectorProperty*, xml_ele_*, char*) (basedevice.cpp:731)
==26462==    by 0x122837: INDI::BaseDevice::setValue(xml_ele_*, char*) (basedevice.cpp:692)
==26462==    by 0x10FF1D: INDI::BaseClientPrivate::dispatchCommand(xml_ele_*, char*) (baseclient.cpp:579)
==26462==    by 0x110CF5: INDI::BaseClientPrivate::listenINDI() (baseclient.cpp:440)
==26462==    by 0x53376DE: ??? (in /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.25)
==26462==    by 0x4E456DA: start_thread (pthread_create.c:463)
==26462==    by 0x5CDA71E: clone (clone.S:95)
```

My though: blob data freeing is commented out in indipropertyview.h, so there is simply no free on removal. I am not sure the impact of activating it back however.

```cpp
template <>
struct WidgetView<IBLOB>: PROPERTYVIEW_BASE_ACCESS IBLOB
{
    ...
    ~WidgetView()                                          { /* free(this->blob); */ }
    void clear()                                           { /* free(this->blob); */ memset(this, 0, sizeof(*this)); }
```

knro:
[...]

PR #1909, opened 2023-06-23
Fix leaks last blob on driver restart

pawel-soja:
Hi :smile: 

Referring to [libindiclient leaks last blob on driver restart](https://github.com/indilib/indi/issues/1528)
I'm uploading as a draft, I have some new equipment and I will thoroughly test if it's ok.

I wish you a clear sky!


knro:
@pludov Can you please check from your side as well?

knro:
@pawel-soja is this still work in progress?

pawel-soja:
> @pawel-soja is this still work in progress?

Hey, I use KStars and the library with this fix all the time. I don't see a regression.
Has anyone tested too?

PR #1914, opened 2023-07-26
drivers: auxiliary: use stdint.h types (fix musl build)

thesamesam:
Followup to 7df8ca6d0de4157268de625c88705c2d96349876. This regressed in 37860f05cb3e4aa55cf7f0d67968b1bcf579e6c3 with the addition of non-standard u_int* types (note the extra underscore).

Bug: https://bugs.gentoo.org/873403


thesamesam:
Thanks!

PR #1917, opened 2023-08-01
Fix build bug on Windows

AstroAir:
Just fix building bug on Windows platform.I hava tested on Msys2(Mingw64)

### DIFF
--- a/libs/indicore/indidevapi.c
+++ b/libs/indicore/indidevapi.c
@@ -11,6 +11,11 @@
 #include <assert.h>
 #include <sys/stat.h>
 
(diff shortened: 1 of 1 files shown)


knro:
Great thanks, that works! Do you have any ideas if the failure in the Visual Studio is similar?

AstroAir:
@knro I looked at the error logs of the build and found that most of them were due to calling proprietary libraries on the Linux platform on the Windows platform, which resulted in errors. Maybe we need to differentiate the platform through macros and rewrite some codes.

PR #1924, opened 2023-08-21
Fix bug in relative positioning of Lacerta MFOC

singraber:
Fixes issue #1842. Lacerta MFOC with newer firmware (tested with version 3.1.01, driver `indi_lacerta_mfoc_fmc_focus`) fails to correctly change position when a relative change in the `Focus In` direction is selected.

The problematic code is `(ticks * (dir == FOCUS_INWARD ? -1 : 1))`, where `ticks` is an unsigned and `-1` a signed integer. `-1` gets converted into a huge unsigned value before the multiplication is carried out. Hence, the new desired `targetTicks` value is huge instead of decreased by the desired amount.

The same code can be found in the "old" driver `lacerta_mfoc.cpp`, the patch is applied there as well (untested).


singraber:
May also fix #1834

chkettu:
@singraber good point - thx for the fix

PR #1930, opened 2023-09-02
Bugfix empty active telescope

sterne-jaeger:
Avoid driver crash if active telescope is not set

The current coordinates are snooped only if ACTIVE_TELESCOPE value is not empty. An attempt to an empty value leads to a crash.


sterne-jaeger:
Changes on master, might already be fixed.

PR #1934, opened 2023-09-13
Use ZSTD compression for XISF images when available

nouspiro:
PixInsight 1.8.9-2 added support for ZSTD compression which is much better than LZ4. So when compression is enabled use ZSTD.


nouspiro:
Require LibXISF 0.2.9

knro:
Can you please update the Ubuntu branch so I can update PPA?

PR #1936, opened 2023-09-21
Planewave mount

knro:
+ Add preliminary support to Planewave mounts.
+ Add INI Parser to parse ini files.
+ Add httplib to support sending requests for http servers and also to create a local http server if required. No need to use libcurl when httplib offers better modern API.


mattiaverga:
Can you please make possible to optionally link against system provided httplib instead of hard bundling it?

PR #1940, opened 2023-09-25
focuser crash fixed+some properties added

anjok:


drivers/telescope/lx200_OpenAstroTech.h
```diff
@@ -79,6 +81,9 @@ class LX200_OpenAstroTech : public LX200GPS
     INumber DecLimitsN[2];
     INumberVectorProperty DecLimitsNP;
 
+    INumber HeaterN[2];
+    INumberVectorProperty HeaterNP;
+
```
knro, line 86: Please use INDI::PropertyNumber

PR #1946, opened 2023-10-02
Option to use system provided httplib

mattiaverga:
Some Linux distributions have their own httplib packaged, this will add an option to use the system provided one (default to use bundled library).


knro:
System provided httplib requires CMake find file as well?

mattiaverga:
> System provided httplib requires CMake find file as well?

Oh, yes, I haven't realized httplib provides CMake find file. 
Should I add a `find_package()` when the option is ON?

knro:
Exactly

mattiaverga:
Ok, not much cmake/C knowledge here, but I hope to have done it right...

CMakeLists.txt
```diff
@@ -165,6 +168,20 @@ install(FILES
...
+# ##################################################################################################
+# ####################################  bundled libs  ##############################################
+# ##################################################################################################
+if(INDI_SYSTEM_HTTPLIB)
+    find_package(httplib REQUIRED)
+    include_directories(${HTTPLIB_INCLUDE_DIR})
+    set(SYSTEM_HTTPLIB 1)
+    message(STATUS "Using system provided httplib version ${HTTPLIB_VERSION}")
+else()
+    include_directories(libs/httplib)
```
knro, line 180: Maybe add ${CMAKE_CURRENT_SOURCE_DIR}/ prefix before libs?

knro:
@pawel-soja Do you know what is causing the failure? I can reproduce the issue locally, it's related to the %include "indiapi.h" is this a new issue though?

pawel-soja:
@knro I quickly looked through what was wrong and I think I found it. I've linked a new pull request to this thread. Once it is built, the problem is found and solved.

PR #1954, opened 2023-10-18
Handling tracking mode in the pegasus driver, and enable King mode.

amendolajamie:
This moves the tacking logic into the pegasus driver. It now handles king mode, and updates when the mount get's an update from another client like unity mobile. Please review @chrysikos 

Jamie


chrysikos:
@amendolajamie Thank you for the update. I appreciate your effort in moving the tracking logic into the Pegasus driver and handling king mode. It looks good!!

PR #1961, opened 2023-11-08
Snapcap driver with network connections

miceno:
This PR adds a new feature to SnapCap driver: TCP connections.

It also fixes an issue related to the Force button.


knro:
Thank you!

PR #1964, opened 2023-11-16
Added TELESCOPE_CAN_FLIP to capabilities

amendolajamie:
Probably need to discuss if this makes sense. Currently it only flips to the current coordinate, and does not goto the new coordinate. 

Also added error checking to the flip commands.

Please review @chrysikos @knro


knro:
Just to confirm, there is an *explicit* flip command support for this mount?

amendolajamie:
Yes, but not with a coirdinat.

chrysikos:
Command MN moves the Ra/Dec axis to the opposite pier side. Subject to meridian limits which silently ignores the flip.

amendolajamie:
I'm going to cancel this since the flip behavior does not also goto a flip location, it just flips to the current location.

PR #1967, opened 2023-12-01
fix(docker): add zmq dependency, output OS info

TallFurryMan:


TallFurryMan:
Don't be fooled by the CI right now, this PR applies to the GitHub Action "docker".

knro:
Thank you! I didn't look in here I just changed the root directory script!

PR #1970, opened 2023-12-02
Allow using system provided json library

mattiaverga:
Some Linux distributions have their own Niels Lohmann's json library packaged, this will add an option to use the system provided one (default to use bundled library).


mattiaverga:
I think I should also make indi provide a `HAVE_BUNDLED_JSON` flag, so indi-3rdparty targets requiring `json.h` know that they should require system's json library... but I'm not expert in cmake and didn't found a way to do that, yet.

knro:
Can you please resolve the conflicts above?

mattiaverga:
Rebase done.

About indi-3rdparty, currently both project use the `INDI_SYSTEM_JSONLIB` option manually.
I think I could have 3rdparty to check for the indi provided json library in `FindINDI.cmake` and require the system library if indilib was built with system json, but I think I would need to change the indi provided json header file name to distinguish from the upstream one. AFAIK, using `find_path()` will always check in system's paths, so using the same name for indi provided and upstream header file will not distinguish the two...

knro:
I think using a different header name is OK if this makes it work across indi-core and indi-3rdparty reliably in both scenarios.

mattiaverga:
I will update https://github.com/indilib/indi-3rdparty/pull/861 ASAP

mattiaverga:
Do we also want to add a workflow to test build using system libraries (json and http)?

Issue #1983, opened 2024-01-01
Cannot connect to flip-flat with virtual serial port unless 'Drop RTS' checks are bypassed as follows


jfwells:
I'm using INDI on an RPi wih  stepper/relay hats. I've impemented my flip-flat as a local python daemon that spawns a local serial pty.

This works fine, but only if the "Drop RTS" hardware ioctl  checks in drivers/auxiliary/flipflat.cpp are bypassed 

I have simply commented out the two 'return false;' statements on lines 154 and 161. but if the drop RTS is needed then this won't work for others with 'real' hardware.

[edit... I will submit a PR that pre-emptively handshakes and returns true, which will work in my case -- on failure it will then drop RTS and try again for real flip-flats]

knro:
It's obviously required for the actual physical port otherwise it wouldn't be there. This driver does not get any update so you can safely fork and modify to fit your use case.

jfwells:
> It's obviously required for the actual physical port otherwise it wouldn't be there. This driver does not get any update so you can safely fork and modify to fit your use case.

Thanks. Yes, the key part is that it doesn't have to return false on failure immediately, as this prevents the virtual serial port from succeeding.

...guess I  need to figure out how to maintain a single forked driver without recompiling all INDI

PR #1984, opened 2024-01-01

jfwells:
Try premptively pinging the flip-flat without dropping the RTS line so that the handshake succeeds in the case of a virtual serial port (pty with no ioctl).

(originally submitted as bug #1983 , but hopefully this PR will work to solve).


jfwells:
Apologies, need to fix an error

PR #1985, opened 2024-01-01

jfwells:
Try premptively pinging the flip-flat without dropping the RTS line so that the handshake succeeds in the case of a virtual serial port (pty with no ioctl).

(originally submitted as bug #1983 , but hopefully this PR will work to solve).

### DIFF
--- a/drivers/auxiliary/flip_flat.cpp
+++ b/drivers/auxiliary/flip_flat.cpp
@@ -144,6 +144,14 @@ bool FlipFlat::Handshake()
     }
 
     PortFD = serialConnection->getPortFD();
+    
+    /* Try handshake first in case of virtual serial port where ioctl would fail*/
+    if (ping())
+    {
+        return true;
+    } 
+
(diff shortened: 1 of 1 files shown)


knro:
was this tested on actual device? would ping fail immediately on real device or would this introduce an necessary delay for users of the physical device?

jfwells:
I have only tested on my device unfortunately. My previous implementation was on an arduino, which I could test to ensure it does not hang up the port, but I don't have a real device to test with unfortunately.

I expect most in the wild implementations are arduino clones though.

jfwells:
I suppose the alternative would be to leave the rts drop first, but do not return false on error, but instead try a handshake directly in that case.

This way, we would not introduce a delay for "real" hardware users, but it would also work in my case.

knro:
I think this would work.. I have Flip flat so I can test and let you know.

jfwells:
> I think this would work.. I have Flip flat so I can test and let you know.

Thanks. Shall I close this and resubmit?

knro:
Sure

PR #1987, opened 2024-01-02

jfwells:
… (revised)

If initial IOCTL call fails, try premptively pinging the flip-flat anyway,  so that the handshake can succeed in the case of a virtual serial port (e.g. a pty with no IOCTL).

This shouldn't impact flip-flat driver users with 'real' or clone hardware, but will allow hardware implementations that use a virtual serial port (e.g. in my case, using stepper/relay hats via gpio on a rpi with a small python daemon)


knro:
Test and works fine on FlipFlat

jfwells:
thank you for testing!

PR #1989, opened 2024-01-09
PegasusAstro SaddlePowerBox

chrysikos:


drivers.xml
```diff
@@ -645,6 +645,10 @@
             <driver name="Pegasus FlatMaster">indi_pegasus_flatmaster</driver>
             <version>1.2</version>
         </device>
+        <device label="Pegasus SPB" manufacturer="Pegasus Astro">
```
knro, line 648: I wonder if using a different alias might make this easier? how about "Pegasus Saddle" ?

chrysikos: @evansgl is that ok ? Or do you preferer something else?

evansgl: Let's use the exact product name:  "Saddle Powerbox"

knro: Sounds good. So full name should be "Pegasus Saddle Powerbox" @chrysikos

PR #2001, opened 2024-02-20
Sesto Senso 2 changing state before changing position

### LINKED ISSUE #1999: Sesto Senso 2 Not reporting focuser position correctly
**Describe the bug**
This is a follow up to Issue #1995.
See this forum thread for more details...
https://indilib.org/forum/focusers-filter-wheels/14322-re-kstars-3-6-9-af-overscan-not-working-correctly.html#99023

Having fixed Issue #1995 there is the following issue:
When asked to do an absolute move, the focuser moves to the requested position and gives interim progress updates to Ekos::Focus so it can provide user feedback on progress. As an example let say the focuser is asked to move from 2000 to 10000 with an Overscan=100.
Ekos requests a focuser move to 10000+100 = 10100
What should happen is that every 500ms Indi updates Ekos::Focus with position and state, so something like
2050 IPS_BUSY
5000 IPS_BUSY
7000 IPS_BUSY
9000 IPS_BUSY
10100 IPS_OK

What actually happens is (see post https://www.indilib.org/forum/focusers-filter-wheels/14322-re-kstars-3-6-9-af-overscan-not-working-correctly.html?start=12#99165)
2000 IPS_BUSY
2000 IPS_BUSY
2000 IPS_BUSY
2000 IPS_BUSY
2000 IPS_OK
10100 IPS_OK

So there are 2 issues here:
1. The driver is not updating Ekos::Focus with interim progress but is returning the start position of the move. This is "nice to have" functionality but its absence won't break Focus functionality.
2. The driver is setting status to IPS_OK when the focuser has finished moving but is not updating the focuser position correctly (it is still sending the start position of the move). This breaks the Overscan functionality in Focus.

**To Reproduce**
Command an Overscan move either manually or with Autofocus

**Expected behavior**
As above.

**Screenshots**
See forum thread above.

**Desktop (please complete the following information):**
 - OS:
 - Version [e.g. 1.7.6, GIT..etc]

**Log Files**
See forum thread above

Ryther:
Artistic Style: done
Testing Suite: done

Tested with latest Kstars 3.6.9:
- Move In: Ok
- Move Out: Ok
- AF Overscan: Ok
- Rapid IN/OUT movements: Ok

Fixes #1999

### DIFF
--- a/drivers/focuser/sestosenso2.cpp
+++ b/drivers/focuser/sestosenso2.cpp
@@ -894,13 +894,11 @@ bool SestoSenso2::AbortFocuser()
 ////////////////////////////////////////////////////////////////////////////////////////////////////////
 void SestoSenso2::checkMotionProgressCallback()
 {
+    IDSetNumber(&FocusAbsPosNP, nullptr);
+    lastPos = FocusAbsPosN[0].value;
     if (isMotionComplete())
     {
-        FocusAbsPosNP.s = IPS_OK;
-        FocusRelPosNP.s = IPS_OK;
         IDSetNumber(&FocusRelPosNP, nullptr);
-        IDSetNumber(&FocusAbsPosNP, nullptr);
-        lastPos = FocusAbsPosN[0].value;
 
         if (CalibrationSP.getState() == IPS_BUSY)
         {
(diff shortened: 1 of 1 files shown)


knro:
Thank you!

PR #2004, opened 2024-02-22
added support for the :GC# command for Moonlite focuser

hbwtrs:
My focusser runs via the Moonlite protocol and stores the temperature compensation coefficient in the flash memory of the focuser. Moonlite protocol supports this by the GC (get coefficient) command.

### DIFF
--- a/drivers/focuser/moonlite.cpp
+++ b/drivers/focuser/moonlite.cpp
@@ -205,6 +205,27 @@ bool MoonLite::readTemperature()
     return true;
 }
 
+bool MoonLite::readTemperatureCoefficient()
+{
+    char res[ML_RES] = {0};
+
+    if (sendCommand(":GC#", res) == false)
+        return false;
+
+    uint8_t coefficient = 0;
+    int rc = sscanf(res, "%hhX", &coefficient);
+    if (rc > 0)
+        // Signed HEX two digits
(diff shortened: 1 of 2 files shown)


drivers/focuser/moonlite.cpp
```diff
@@ -205,6 +205,27 @@ bool MoonLite::readTemperature()
...
+    char res[ML_RES] = {0};
+
+    if (sendCommand(":GC#", res) == false)
+        return false;
+
+    uint8_t coefficient = 0;
+    int rc = sscanf(res, "%hhX", &coefficient);
+    if (rc > 0)
+        // Signed HEX of two digits
+        TemperatureSettingNP[1].setValue(static_cast<int8_t>(coefficient) / 2.0);
```
knro, line 219: TemperatureSettingNP[Coefficient].setValue(...)

Btw, did you test the rest of the driver since @naheedsa made significant changes today to it.

hbwtrs: I noticed the code changes that were made after my first addition, so had to adapt a little bit. Itś daytime now, but I tested the driver with mount, telescope and focuser attached, all fields seems to work as before.

hbwtrs: Tested on different systems today, with different hardware connected, no issues found. When I connect the focuser to a different computer, the temperature coefficient is taken from the focuser with the GC command as the driver initializes, as is the indented behavior.

Issue #1992, opened 2024-02-03
Pegasus PPBA/M missing auto dew aggressiveness setting


ijessen:
**Describe the bug**
The PPBA/PPBM driver doesn't handle the auto dew "aggressiveness" setting.  It looks like support was added for this parameter in a later serial protocol update (Jan 2021), [see here](https://pegasusastro.com//wp-content/uploads/2022/05/Pocket_Powerbox_Advance_Serial_Command_Table.pdf).

knro:
Thank you for the report. Do you think you can work on a PR for this?

ijessen:
Yep - am planning to

ijessen:
I started on this but my PPBM doesn't seem to respond to the aggressiveness commands like the documentation says it should.  I have a request in to Pegasus for info.

![image](https://github.com/indilib/indi/assets/2163199/c36201f1-b518-4275-9471-d4522bfad810)

ijessen:
Pegasus support finally responded that they will update their documentation at some point, hopefully soon

PR #2006, opened 2024-02-28
Support setting autodew aggression (#1992)

ijessen:
Also fixed two little bugs:

1. stopChar being overridden causing auto search to fail
2. firmware version not being read


knro:
Thank you!

PR #2012, opened 2024-03-01
Fix iguider issues

beaglebot:
I've resolved the exposure issue https://github.com/indilib/indi/issues/1722 with the iOptron iGuider/iPolar cameras. 

Unfortunately fixing this has brought up another problem with excessively slow capture startup times. I've partially resolved this, however there's still a reliability issue where the camera sometimes refuses to start after being stopped, or ignores exposure commands. 

Any suggestions on dealing with this last issue would be appreciated!

### Notes on Exposure Issue
The iGuider manual and ASCOM dialogs give the impression that negative values need to be used to set the Absolute Exposure time on the camera. This turns out to be not true. Sniffing the USB bus while changing exposure settings in ASCOM shows the negative values do not get sent to the USB Absolute Exposure Controls. Instead _somewhat_ conventional ticks are sent. See the table below:

|    Duration (s) |  Value from USB sniff |
|-|-|
|    0.01  |         1 |
 |   0.02   |        2   |
 |   0.05   |        5     |
 |   0.1     |       10      |
 |   0.2     |       20        |
 |   0.5     |       39        |
 |   1        |      78          |
 |   1.5     |       156       |
 |   2        |      312        |
 |   2.5     |       625        |
 |   3        |      1250       |
|    3.5     |       2500      |

By 'somewhat', I mean that the ticks are not linear with respect to duration. It appears they are linear until 0.2 seconds, then exponential afterwards. Presumably this is done to give more resolution at lower exposure times.  We can recover the formula for the exponential equation by assuming 4 seconds corresponds to 5000 ticks (the max). This gives us:
```
        ticks <= 0.2 ? duration * 100 : 19.53 * pow(4, duration)
```
Where the constant 19.53 is  5000 / 4^4. This matches the values in the table exactly when rounded.

Plugging in this conversion factor works, and the resulting frame timings in the debug logs are correct.

See https://github.com/indilib/indi/commit/ddce625ee8c502503ec6d3509e55bf5378618b95

### Notes on the slow capture start up issue

This brought up another issue. If you attempt to start the capture with a multi-second exposure time, the camera takes many, many seconds before delivering the first frame. For example:

| Initial exposure time (s) | Delay before first frame delivered (s) |
|-|-|
| 1.0 | 4.0 |
| 2.0 | 8.0 |
| 3.0 | 12 |
|4.0 | 27 |

This looks exponential-ish, as if there's a delay somewhere in the driver or camera based on the tick count.

ASCOM seems to deal with this by rarely shutting down the stream. As soon as the ASCOM driver is activated, the camera immediately begins sending frames over USB, which presumably the driver just discards. Clicking stop/start in PHD doesn't result in any USB commands being sent to the camera - this must just affect whether the driver forwards the frames to PHD or discards them. Even after disconnecting the camera in PHD, the frames continue to be sent over USB. A little wasteful perhaps, but it seems to work.

A dirty hack to get around this is to set the initial exposure to 0.01s (1 tick), start the capture, wait for the first frame to be delivered, then set the exposure to the desired value (earlier doesn't work). This works, and results in a delay of about 150ms, which is acceptable. As long as the stream is shutdown lazily, rather than immediately after a frame is delivered, PHDs frequent frame requests don't incur this delay and everything works well.

See https://github.com/indilib/indi/commit/e0b86fef13a3e9aee0005d1385eb8249c7efbbaf

*** Ongoing issues

Well, almost. I can get some reasonable runs with PHD where everything works: I can stop and start the capture, set different exposure rates and gains successfully, however at some point something fails and the capture won't restart.

Occasionally when this happens, I can issue repeated IOCTLs to abort the capture, or re-set the values of USB controls, and this will restart it, however many times this does not work. When attempting to change the 'Capture Size' I get an 'Resource busy' error, suggesting the driver/camera wasn't correctly shutdown.

Occasionally, it will restart but run at 9 or 20fps, and ignore attempts to change the exposure, again as if the driver/camera is stuck in some other mode.

This doesn't occur on windows. One difference is that when ASCOM needs to stop the capture or change a control it issues an ABORT_PIPE URB followed by a SYNC_RESET_PIPE_AND_CLEAR_STALL URB. Linux/libuvc/v4l2 doesn't do this and just issues the ABORT_PIPE URB. 

There's an interesting comment around this in the libuvc source: https://github.com/torvalds/linux/blob/87adedeba51a822533649b143232418b9e26d08b/drivers/media/usb/uvc/uvc_video.c#L2254. In our case the interface alt_setting is changed after the ABORT_PIPE, so alt_setting is still set to 6, so the branch in the code above isn't taken, and the  `usb_clear_halt` isn't called. 

I could try tweaking this function in the kernel to see if it fixes our problem, but even if it did, I imagine it's hard to get a patch applied to the kernel - I can't imagine linus risking breaking a bunch of UVC cameras for our iGuider.

Any thoughts/suggestions very welcome!

### DIFF
--- a/drivers/video/v4l2driver.cpp
+++ b/drivers/video/v4l2driver.cpp
@@ -62,7 +62,7 @@ V4L2_Driver::V4L2_Driver(std::string label, std::string path)
     strncpy(defaultVideoPort, path.c_str(), 256);
     strncpy(configPort, path.c_str(), 256);
 
-    setVersion(1, 1);
+    setVersion(1, 2);
 
     allocateBuffers();
 
@@ -940,11 +940,47 @@ bool V4L2_Driver::setManualExposure(double duration)
     // INT control for manual exposure duration is an integer in 1/10000 seconds
     long ticks = lround(duration * 10000.0f);
 
-    // iOptron iPolar & iGuider have different tick rate
-    // With max about 5000 ticks
+
+    /*
+    iGuider and iPolar do not have a linear relationship between duration and ticks, as demonstrated in the table 
+    below. It contains exposure durations in seconds as set in ASCOM, and the resulting values sent to the camera 
+    via a SET CUR USB request to the 'Exposure Time (Absolute)' control in the USB Video Control Input Terminal. 
+    
+    The third column is an approximation to this value using the formula: 
+
+        duration <= 0.2 ? duration * 100 : 19.53 * pow(4, duration)
+
+    The constant 19.53 above iscalculated as 5000 / 4^4, as it appears the maximum tick of 5000 corresponds 
+    to a 4 second duration.
+
+    Note that the negative exposure values seen in the iOptron iGuider manual and some of the settings dialogs appear
+    to used in the ASCOM driver but aren't sent to the camera over USB.
+
+    Duration (s)   Value from USB sniff   Approximation
+    0.01           1                      1.00
+    0.02           2                      2.00
+    0.05           5                      5.00
+    0.1            10                     10.00
+    0.2            20                     20.00
+    0.5            39                     39.06
+    1              78                     78.12
+    1.5            156                    156.24
+    2              312                    312.48
+    2.5            625                    624.96
+    3              1250                   1,249.92
+    3.5            2500                   2,499.84
+
+    */
     if (strstr(getDeviceName(), "iGuider") || strstr(getDeviceName(), "iPolar"))
     {
-        ticks = std::min(5000l, lround(duration * 1333.33));
+        if (duration <= 0.02f)
+        {
+            ticks = lround(duration * 100.0f);
+        }
+        else
+        {
+            ticks = lround(19.53 * pow(4, duration));
+        }
     }
 
     /* First check the presence of an absolute exposure control */
@@ -1005,6 +1041,16 @@ bool V4L2_Driver::setManualExposure(double duration)
                           duration, StackModeSP[m_StackMode].getName());
             }
             ticks = AbsExposureN->max;
+            if (strstr(getDeviceName(), "iGuider") || strstr(getDeviceName(), "iPolar"))
+            {
+                frame_duration.tv_sec  = 4;
+                frame_duration.tv_usec = 0;
+            }
+            else
+            {
+                frame_duration.tv_sec  = ticks / 10000;
+                frame_duration.tv_usec = (ticks % 10000) * 100;
+            }
         }
         /* We can't expose as long as requested and stacking is not configured, bail out */
         else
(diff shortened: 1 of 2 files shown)


knro:
That's some great analysis @beaglebot , the changes I've made recently as I have access to iGuider and was never able to control it properly before. I concur the relationship isn't linear but I got confirmation from iOptron that the absolute exposure is interpolated differently within the camera from the standard absolute_exposure definition in units of 100 uS.

1. Changes to the Kernel could fix this? Even if it doesn't immediately apply to most of us, it will eventually *land* in the future (if accepted) so if you're confident with this change, I'd say go ahead as it will help other use cases as well. Did you happen to test this locally as well?
2. I didn't test this yet with Ekos or PHD2, just tried a few simple capture sessions and it seems to hold fine, but my test wasn't extensive in any way. How often do you get the PHD2 capture issues? does it occur after 30 seconds? 5 minutes? or random?

knro:
Let me test and see if this works with iGuider, thanks!

drivers/video/v4l2driver.cpp
```diff
@@ -806,6 +854,16 @@ bool V4L2_Driver::ISNewNumber(const char * dev, const char * name, double values
         return true;
     }
 
+    // Ensure duration is set to one of the allowed durations, or, if we're going to stack, a value greater than the maximum
+    if (isIOptron() && strcmp(name, "CCD_EXPOSURE") == 0 && values[0] < MAX_IOPTRON_DURATION) {
+        int n = findIndexOfNearestValue(values[0], IOptronDurations);
+        if (values[0] != IOptronDurations[n]) 
+        {
+            LOGF_WARN("Changing duration to nearest value supported by camera (from %fs to %fs)", values[0], IOptronDurations[n]);
```
knro, line 862: Could this have the effect of being spammy if a user keep using the same non-standard exposure repeatedly only to this this warning over and over?

drivers/video/v4l2driver.cpp
```diff
@@ -881,6 +939,17 @@ bool V4L2_Driver::StartExposure(float duration)
...
+
+        // iGuider/iPolar occasionally fails to start. We deal with this in the same was
+        // as the iOptron ASCOM driver on windows: if it hasn't responded with a frame in
+        // 10 seconds, stop and start the capture.
+        if (isIOptron())
+        {
+            valid_frame_has_arrived = false;
+            if (-1 != iOptronWatchdogTimer)
+                IERmTimer(iOptronWatchdogTimer);
+            iOptronWatchdogTimer = IEAddTimer(IOPTRON_WATCHDOG_PERIOD_IN_MS, (IE_TCF *)iOptronWatchdogCallback, this);
```
knro, line 951: Please use the new style INDI::ElapsedTimer for this

drivers/video/v4l2driver.cpp
```diff
@@ -1128,6 +1216,31 @@ void V4L2_Driver::stdtimerCallback(void * userpointer)
...
+{
+    V4L2_Driver * p = (V4L2_Driver *)userpointer;
+
+    if (p->valid_frame_has_arrived)
+    {
+        p->iOptronWatchdogTimer = -1;
+        return;
+    }
+
+    INDI::Logger::getInstance().print(p->getDeviceName(), INDI::Logger::DBG_SESSION, __FILE__, __LINE__, "Valid frame hasn't arrived yet. Restart capturing.");
```
knro, line 1231: Should this be warning? or debug even?

drivers/video/v4l2driver.cpp
beaglebot: Yeah, good thought. I guess this is just a guider/polar alignment camera so users aren't going to be too concerned about exact exposures times. I'll remove it.

PR #2029, opened 2024-03-15
feat: Dark Dragons Astro Devices

rickbassham:
Adds 2 new devices to INDI, the DragonLAIR roof controller and the DragonLIGHT flat panel controller.

### DIFF
--- a/drivers/auxiliary/dragonlight.cpp
+++ b/drivers/auxiliary/dragonlight.cpp
@@ -115,6 +115,11 @@ bool DragonLIGHT::EnableLightBox(bool enable)
     std::string endpoint = enable ? "/indi/turnon" : "/indi/turnoff";
 
     auto result = cli.Post(endpoint);
+    if (!result)
+    {
+        LOG_ERROR("Unable to connect.");
+        return false;
+    }
 
     IDLog("EnableLightBox: %d\n", result.value().status);
     IDLog("EnableLightBox: %s\n", result.value().body.c_str());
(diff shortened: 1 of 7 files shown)


drivers/auxiliary/dragonlight.cpp
```diff
@@ -0,0 +1,381 @@
...
+}
+
+const char *DragonLIGHT::getDefaultName()
+{
+    return "DragonLIGHT";
+}
+
+bool DragonLIGHT::EnableLightBox(bool enable)
+{
+    LOGF_INFO("EnableLightBox: %d\n", enable);
```
knro, line 111: No need for \n and you probably want to use LOGF_DEBUG for this, not INFO

drivers/auxiliary/dragonlight.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+    auto result = cli.Post(endpoint);
+    if (!result)
+    {
+        LOG_ERROR("Unable to connect.");
+        return false;
+    }
+
+    IDLog("EnableLightBox: %d\n", result.value().status);
+    IDLog("EnableLightBox: %s\n", result.value().body.c_str());
```
knro, line 125: Don't use IDLog. If this was for development, you can remove now.

drivers/auxiliary/dragonlight.cpp
```diff
@@ -0,0 +1,381 @@
...
+    }
+
+    if (result.value().status == 200)
+    {
+        return true;
+    }
+    else
+    {
+        return false;
+    }
```
knro, line 169: return (result.value().status == 200);

PR #2032, opened 2024-03-21
Fix setting ScopeInfo property

tstibor:
Setting Scopeinfo property with previous settings: min, max, step, val:

`ScopeInfoNP[FocalLength].fill("FOCAL_LENGTH", "Focal Length (mm)", "%.2f", 10, 10000, 100, 0); ScopeInfoNP[Aperture].fill("APERTURE", "Aperture (mm)", "%.2f", 10, 3000, 100, 0);`

resulted in error:

Error: Invalid range for Aperture (mm) (APERTURE). Valid range is from 10 to 3000. Requested value is 0 [DEBUG] Configuration successfully saved for SCOPE_INFO. Error: Invalid range for Focal Length (mm) (FOCAL_LENGTH). Valid range is from 10 to 10000. Requested value is 0 [DEBUG] Configuration successfully saved for SCOPE_INFO.

because at least one property value is 0 by default however, min values are set to 10mm, so the value cannot be set via the property settings. In addition the step size is decreased to 1mm and enum values are in uppercase letters. Moreover, the state of the property is IPS_OK only when the update is successful, otherwise IPS_ALERT.

### DIFF
--- a/libs/indibase/indiccd.h
+++ b/libs/indibase/indiccd.h
@@ -728,8 +728,8 @@ class CCD : public DefaultDevice, GuiderInterface
         INDI::PropertyNumber ScopeInfoNP {2};
         enum
         {
-            FocalLength,
-            Aperture
+            FOCAL_LENGTH,
+            APERTURE
         };
 
         // Websocket Support
--- a/drivers/ccd/ccd_simulator.cpp
+++ b/drivers/ccd/ccd_simulator.cpp
@@ -539,7 +539,7 @@ int CCDSim::DrawCcdFrame(INDI::CCDChip * targetChip)
 
     exposure_time *= (1 + sqrt(GainNP[0].getValue()));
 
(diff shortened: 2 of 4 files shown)


libs/indibase/indiccd.h
```diff
@@ -728,8 +728,8 @@ class CCD : public DefaultDevice, GuiderInterface
         INDI::PropertyNumber ScopeInfoNP {2};
         enum
         {
-            FocalLength,
-            Aperture
+            FOCAL_LENGTH,
+            APERTURE
```
knro, line 732: is there a reason for this change? any other drivers using these enums?

tstibor: It is used also in `guide_simulator.cpp` and `ccd_simulator.cpp`. Checking other parts of the code, it seems to be conform to use capital letter enums definitions. So far all
```
enum {
 CAPITAL_LETTERS
 ...
 ...
}

```
have this capital letter pattern.

PR #2049, opened 2024-04-27
Bugfix: update focuser temperature in alluna tcs2 driver

peter-englmaier:
I noticed, that the focuser temperature was not updated in kstars and other clients. The reason was, that the  .apply() method was not called when reading new values from the device.

Other (minor) changes:
- update temperature more often (every 10 seconds instead of once every 5 minutes)
- update fan speed less frequently (very 30 seconds instead of every 3 seconds)
- corrected the copyright message in the header file to reflect my authorship (getting it in sync with the source file).

### DIFF
--- a/drivers.xml
+++ b/drivers.xml
@@ -1,13 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <driversList>
- 	<devGroup group="Spectrographs">
+    <devGroup group="Spectrographs">
         <device label="Shelyak Usis" manufacturer="Shelyak">
             <driver name="Shelyak Usis">shelyak_usis</driver>
             <version>1.2</version>
         </device>
-	</devGroup>
+    </devGroup>
     <devGroup group="Telescopes">
-		<device label="Shelyak Usis" manufacturer="Shelyak">
+        <device label="Shelyak Usis" manufacturer="Shelyak">
             <driver name="Shelyak Usis">shelyak_usis</driver>
             <version>1.2</version>
         </device>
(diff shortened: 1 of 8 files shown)


knro:
Thank you @peter-englmaier can you please check the MacOS errors?

peter-englmaier:
@knro: This seems to be quite deeply rooted in indi. I have not touched any code near that. 

It seems, on a mac, the 1-bit wide field 'ok' is interpreted as a signed integer. So instead of `sp->ok = 1`  one needs to set  `sp->ok = -1`, however I rather would change the type to unsigned. With this the error is gone, but instead it fails for some other reason. Perhaps something has changed with the macos compiler defaults?

peter-englmaier:
@knro I do not grasp why the MacOS build fails. The compiler complains about missing function prototypes, but those are clearly present at the beginning  of the source code:
```
static int compile(int prec);
static int execute(double *result);
static int parse_fieldname(char name[], int len);
```
Can you please help?

knro:
Yes sorry, hopefully the issue is now fixed. I'm running the builds again.

peter-englmaier:
@knro Ok, I see you have committed some changes. I try to rebase and see if it compiles again.

peter-englmaier:
Looks like my merge has messed things up, I should probably start from scratch...

peter-englmaier:
I hope #2055 is going to compile... it is rebased on the latest master

peter-englmaier:
This PR is replaced by #2055, so I am closing it.

PR #2054, opened 2024-05-03
Enable Astrotrac 360 driver guide pulse

alpha-frank:
Unguided dithering in Ekos does not work with the Astrotrac 360 when Astrotrac is selected with "Guide via" in the Kstars optical train. List of changes to potentially enable this feature and some possibly bug fixes:

- initialize the GuideRate property (currently not initialized). The new default is set to 0.1x, which is the default rate for the mount according to the author of the X2 plugin. However, users might need to set a longer guide pulse to see movement at the default rate. Note, I'm not sure if the missing initialization is a bug or was omitted by design.

- check if the mount is tracking before issuing a guide pulse

- Added new function to return current tracking rate

- use the current tracking rates rather than custom rates to calculate the mount movement in arcseconds. The custom rates will only be used if the track mode is set to "Custom".

- if the DEC tracking rate is 0, which is the normal case, then the velocity for a south guide pulse is set to negative. This ensures that DEC guide pulses are not always in the same direction as RA when the DEC rate is 0.

- reset tracking when the guide pulse is finished

I have tested the changes with my mount but the test coverage is very limited. I don't use a guide camera (hence unguided dithering) so I don't know what effects, if any, these changes might have on guided dithering or guiding in general.

### DIFF
--- a/drivers/telescope/astrotrac.cpp
+++ b/drivers/telescope/astrotrac.cpp
@@ -96,6 +96,11 @@ bool AstroTrac::initProperties()
     EncoderNP[AXIS_DE].fill("AXIS_DE", "Declination", "%.2f", -3600, 3600, 100, 0);
     EncoderNP.fill(getDeviceName(), "MOUNT_ENCODERS", "Encoders", MOTION_TAB, IP_RW, 60, IPS_IDLE);
 
+    /* How fast do we guide compared to sidereal rate */
+    GuideRateNP[AXIS_RA].fill("GUIDE_RATE_WE", "W/E Rate", "%g", 0, 1, 0.1, 0.1);
+    GuideRateNP[AXIS_DE].fill("GUIDE_RATE_NS", "N/S Rate", "%g", 0, 1, 0.1, 0.1);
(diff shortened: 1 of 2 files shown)


knro:
Isn't 0.1 guide rate odd? For most mounts, it's set to 0.5.

alpha-frank:
> Isn't 0.1 guide rate odd? For most mounts, it's set to 0.5.

Good point, I forgot to mention that this rate only relates to the ST-4 guide interface. Unfortunately, I don't know why the rate is fixed at 0.1x for this mount, which is odd, but I don't find an interface command to set this value. Therefore, it's important to note that the INDI settings will not change the ST-4 defaults. I guess this might be the reason why the guide rate settings are currently not available in the control panel.

One option might be to hide the guide rate settings and initialize the driver rate to a fixed value of 0.1x to match the ST-4 rate or 0.5x to match most other mounts. Otherwise, expose the guide rate settings in the control panel, as I did, and decide on the most appropriate default value - I have no problem with 0.5x. Either way, the ST-4 defaults won't be affected and driver documentation will need to be updated to inform users of the new behavior.

PR #2055, opened 2024-05-04
Bugfix: update focuser temperature in alluna tcs2 driver

peter-englmaier:
I noticed, that the focuser temperature was not updated in kstars and other clients. The reason was, that the  .apply() method was not called when reading new values from the device.

Other (minor) changes:
- update temperature more often (every 10 seconds instead of once every 5 minutes)
- update fan speed less frequently (very 30 seconds instead of every 3 seconds)
- corrected the copyright message in the header file to reflect my authorship (getting it in sync with the source file).


peter-englmaier:
It looks to me, that the build errors have nothing to do with my PR. I guess, I have to wait until this has been sorted out.

PR #2058, opened 2024-05-11
Add debhelper package - fix package build

pawel-soja:
Hi Jasem,

I noticed that *.deb packages are not being built in GitHub Actions. Probably something changed on Ubuntu and I added the 'debhelper' package.
For everything to work fully, after accepting the pull request, you need to rebuild the docker images used by the build system.

I can also take care of it, and if there are any problems, I can fix the problems directly in the repository.

### DIFF
--- a/.github/docker/Dockerfile.ubuntu
+++ b/.github/docker/Dockerfile.ubuntu
@@ -30,4 +30,4 @@ RUN apt-get install -y \
 
 # Install deb package tools
 RUN apt-get install -y \
-    cdbs
+    cdbs debhelper


knro:
Thank you! How to rebuild the docker images exactly? I think we need to add this to the README (or developer README) since I forgot how!

pawel-soja:
@knro Can you remind me where the developer README is?

knro:
Sorry there is none, we can add to the main README or perhaps create another one under ~/.github ? What do you think?

PR #2059, opened 2024-05-17
fix(indiserver): avoid strict warning failure on socket struct copy.

TallFurryMan:
See http://man7.org/linux/man-pages/man7/unix.7.html for relevant information about sockaddr_un:sun_path.


knro:
Thanks. @pawel-soja Do you know what's going on with the above MacOS + PyINDI errors? I don't believe we made any changes to those. Should we rebuild the dockers?

PR #2061, opened 2024-05-22
Implement DustCapInterface

peter-englmaier:
I finally figured out how to implement the DustCapInterface. The device controls the robotic dust cap for the main mirror. 
However, it does not close the light path, so light (and even dust) can still reach the sensor.

Without this modification, the driver can control the DustCap, but it is not registered as a DustCap device.


peter-englmaier:
Basically, I have replaced `CoverSP` with `DustCapSP` and reverted to old style because the definition of `DustCapSP` is outside my code in the `DustCapInterface` definition.

PR #2063, opened 2024-05-23
Fix build - SWIG 4.2.0

pawel-soja:
After upgrading Ubuntu from 22.04 to 24.04, there is a problem building PyIndi.
The new SWIG version 4.2.0, with the current code, can no longer deduce types correctly. The posted fixes declare types explicitly in templates, and by declaring new ones via `using`.

I also added the `break-system-packages` flag to `pip3` to install the `requirements-test.txt` packages from the `pyindi-client` repository.

### DIFF
--- a/.github/workflows/linux-pyindi.yml
+++ b/.github/workflows/linux-pyindi.yml
@@ -51,7 +51,7 @@ jobs:
 
       - name: Install test deps
         run: |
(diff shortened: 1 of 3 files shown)


knro:
Thank you! Need to do a release for pyINDI next week.

PR #2068, opened 2024-05-30
Make openweathermap show rain and snow data

peter-englmaier:
I noticed that openweathmap reports rain, but it does not show up in the driver panel. The API reports the rain like this:
```
  "rain": {
    "1h": 0.4
  },
```

But, apparently indi driver expects this:
```
  "rain": {
    "h": 0.4
  },
```

I could not figure out, if the format has changed or if this just is not the right data.
Should this PR change be made?

### DIFF
--- a/drivers/weather/openweathermap.cpp
+++ b/drivers/weather/openweathermap.cpp
@@ -253,15 +253,30 @@ IPState OpenWeatherMap::updateWeather()
         weatherReport["wind"]["speed"].get_to(wind);
         // Cloud
         weatherReport["clouds"]["all"].get_to(clouds);
-        try
+        // Rain (does not exist in all reports)
+        if (weatherReport.contains("rain"))
         {
-            // Rain
-            weatherReport["rain"]["h"].get_to(rain);
-            // Snow
-            weatherReport["snow"]["h"].get_to(snow);
+            if (weatherReport["rain"].contains("h"))
+            {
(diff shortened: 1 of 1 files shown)


peter-englmaier:
@knro Before you accept this PR, please note, that the source code says about rain and snow values: "these values do not exist in all reports."  I did not find documentation on openweathermap stating that this information is not always present. I only see an example with the data and the data is in the format I see, when I call the API myself. But, it might be, that in some places the value is given with "h". I could change the code, to first look for "h", and if not found look for "1h". This would make it backwards compatible.

knro:
yes we should probably account for h and if not found then 1h to be safe.

peter-englmaier:
Ah, good point. Yes, indeed there is a method. It also works without catching exceptions.

By the way, it is normal that observatory status is 'green', when is raining and weather status is 2 and status code 520? All three should indicate danger. But status is 'green'!
grafik
I have not much experience with weather or observatory status in kstars. So I do not know what is normal. When I adjust the parameters for wind or temperature, I can get orange or red status.

knro:
There is a LOT of work required for the weather handling in *both* observatory + scheduler modules in KStars. For example, scheduler need more complex handling of weather but right now it's a simple go and no go and it's not very reliable as well. If you can perhaps help us in this regard it would be a tremendous help to KStars!

PR #2072, opened 2024-06-09
Weather watcher dynamic labels

escribana:
This PR adds dynamic (user definable) labels for the weather parameters. While the default ones are still the same as before, a user can now name the labels at will. Even the critical parameters are selectable! This increases the usability and comprehensibility of the driver in a wide range.
The storing procedure is as simple a possible: Enter the desired names in the "Labels" tab, save the configuration in"Options", disconnect and reconnect the driver and the new labels are active. After that one can now check the critical parameters and repeat the above procedure.

There is a small drawback in that not all special characters are allowed. But normally one can find easily a possible transcription.

Example:
Default labels: "Rain", "Temperature", "Wind", "Gust", "Clouds", "Humidity", "Pressure" "Forecast"
New labels: "Precipitation [%]", "Ambienttemp [C]", "Windspeed [km/h]", "Cloudcover [%]", "", "", "", ""

WeatherWatcherMain
WeatherWatcherLabels
WeatherWatcherParameters

### DIFF
--- a/drivers/weather/weatherwatcher.cpp
+++ b/drivers/weather/weatherwatcher.cpp
@@ -30,6 +30,9 @@
 #include <memory>
 #include <cstring>
 
+const char *LABELS_TAB  = "Labels";
+const char *PARAMETERS_TAB = "Parameters";
+
 // We declare an auto pointer to WeatherWatcher.
 static std::unique_ptr<WeatherWatcher> weatherWatcher(new WeatherWatcher());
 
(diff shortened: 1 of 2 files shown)


knro:
Wouldn't we lose then all standard definitions for these property? clients can no longer use WEATHER_TEMPERATURE to check for temperature, right?

escribana:
I did not find any references to "WEATHER_TEMPERATURE" in KStars, INDI and INDI-3rdparty. I checked the observatory module and it seems to work correct.  But i will do some more tests.
BTW, are there any other clients to consider?

escribana:
Going back to original naming of properties:

knro:
Thanks! looks good.

PR #2074, opened 2024-06-15
Add button for custom resolution in simulator CCD

escribana:
Add button for custom resolution to reestablish the possibility to work with the resolution defined in simulator config.

### DIFF
--- a/drivers/ccd/ccd_simulator.cpp
+++ b/drivers/ccd/ccd_simulator.cpp
@@ -196,7 +196,10 @@ bool CCDSim::initProperties()
     for (uint8_t i = 0; i < Resolutions.size(); i++)
     {
         std::ostringstream ss;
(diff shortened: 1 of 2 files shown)


knro:
So when custom is selected, it should use whatever is in simulator config? this is not implemented yet.

escribana:
I do not understand: I tested with EKOS and it works!
But there is indeed  little problem insofar the image settings are not set immediately in the Image Settings tab of the INDI Control Panel. Do you address this with your comment?

knro:
I guess what I mean is this: if you select custom, then what happens? how is the resolution set/updated?

escribana:
The problem arose with 3.7.2. I could not use my custom resolution in Simulator Config anymore, since it always jumped back to the preset resolution 1280x1024. With this "Custom" button this resolution is preserved again. So is "Config" a better designation for the button?

PR #2082, opened 2024-07-17
Add New Driver Terrans PowerBoxGo V2

iTerrans:
This is the official driver for Terrans PowerBoxGo V2, which belongs to the same series as Terrans PowerBoxPro V2. Thank you.

### DIFF
--- a/drivers/auxiliary/waveshare_modbus_relay.cpp
+++ b/drivers/auxiliary/waveshare_modbus_relay.cpp
@@ -52,6 +52,9 @@ bool WaveshareRelay::initProperties()
 
     registerConnection(tcpConnection);
 
+    FirmwareVersionTP[0].fill("VERSION", "Version", "1.00");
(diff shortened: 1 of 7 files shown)


drivers/auxiliary/Terrans_PowerBoxGo_V2.h
```diff
@@ -0,0 +1,125 @@
...
+        ISwitchVectorProperty USBFSP;
+        
+        ISwitchVectorProperty StateSaveSP;
+        
+        //Sensor Date
+        INumber InputVotageN[1];
+        INumberVectorProperty InputVotageNP;
+        
+        INumber InputCurrentN[1];
+        INumberVectorProperty InputCurrentNP;
```
knro, line 109: Please use new INDI::PropertySwitch, INDI::PropertyText, INDI::PropertyNumber ...etc as the old style properties (e.g. IXXXProperty..etc are deprecated).

PR #2087, opened 2024-07-27
Feature - ZWO AM5 Buzzer setup

pawel-soja:
Hello!

I have an AM5 mount and I noticed that the "INDI Control Panel" does not have the ability to change the buzzer volume. The property was not associated with the setBuzzer function and the command was missing the trailing sign (#).

I checked the command by intercepting the transmission from `StarGazing`.
https://github.com/pawel-soja/ZWO-AM5-Proxy

### DIFF
--- a/drivers.xml
+++ b/drivers.xml
@@ -81,11 +81,11 @@
         </device>
         <device label="ZWO AM5 WiFi" manufacturer="ZWO">
             <driver name="ZWO AM5">indi_lx200am5</driver>
-            <version>1.1</version>
+            <version>1.2</version>
         </device>
         <device label="ZWO AM5 USB" manufacturer="ZWO">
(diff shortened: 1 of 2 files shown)


pawel-soja:
The pull request will build correctly after applying https://github.com/indilib/indi/pull/2088 changes. If they are accepted, I can merge to be sure.

I know that drivers are numbered, I can bump the version to `setVersion(1, 2);` but I think there was also information about the version in some files. I don't remember where what was and whether to change anything. Please advise.

knro:
Yes bumping the version would be great, thanks!

pawel-soja:
I bumped the version to 1.2 in the files:
- drivers.xml
- lx200am5.cpp

@knro Is it OK? Will you accept the pull request?

PR #2103, opened 2024-08-22
Add workaround for OnStepX failing connection

gy-lehel:
OnStepX controllers have a tendency to come up with a serial buffer blocked by garbage. As a result of this handshake with the mount will consistently fail, until that garbage is cleared.
To work around this issue the :GVP# command is sent to the mount twice.


azwing:
Thant's fine with me as far it fixes a firmware issue.

PR #2123, opened 2024-09-24
use MOTION_CONTROL_MODE _FOO enums

naheedsa:


hiro3110i:
@naheedsa I closed this issue once, but the issue still remains, I found.
The joystick axes are different from previous assignment.
Also, they do not match the axes in the Ekos mount control GUI.
Further investigation and fixes are required.

PR #2133, opened 2024-10-08
Migrate INDI Astrolink4

naheedsa:


naheedsa:
@knro , should i also delete this driver from INDI 3rd party drivers?

PR #2134, opened 2024-10-08
Migrate Astromech focuser

naheedsa:


naheedsa:
@knro , defined inside .cpp file

/* Define Driver version */
#define INDI_ASTROMECHFOC_VERSION_MAJOR 0
#define INDI_ASTROMECHFOC_VERSION_MINOR 2

knro:
We don't need that. Just set it directly in the constructor i.e.

setVersion(0, 2);

naheedsa:
> We don't need that. Just set it directly in the constructor i.e.
> 
> setVersion(0, 2);

Done

PR #2138, opened 2024-10-11
Update ioptronv3driver.cpp

joe13905179063:
add iOptron HAE16 HAE18, and Fix ":SPP0#" causes timeout issue


knro:
Thank you, please fix the error in the build.

PR #2142, opened 2024-10-30
Upgrade CheapoDC driver for new device firmware features

hcomet:
CheapoDC driver release to add support for firmware 2.x features
        - bug fix for Longitude range checking (actually a firmware bug).
        - adds support for Weather Device snoop for local temperature/humidity instead of using CheapoDC weather query
        - move Latitude/Longitude settings to more common Site Management Tab with Location settings
        - Location Snoop enabled by default for Telescope Simulator 
        - maintain backwards compatibility with previous firmware version
Weather Simulator
         - added Humidity as a weather value so that dew point may be calculated by snooping devices. (ie: CheapoDC)

### DIFF
--- a/drivers/auxiliary/cheapodc.cpp
+++ b/drivers/auxiliary/cheapodc.cpp
@@ -138,51 +138,28 @@ bool CheapoDC::initProperties()
     EnableSnoopTemperatureSP[1].fill("DISABLE", "Disable", ISS_ON);
     EnableSnoopTemperatureSP.fill(getDeviceName(), "ENABLE_SNOOP_TEMPERATURE", "Snoop Temp", OPTIONS_TAB, IP_RW, ISR_1OFMANY, 0, IPS_IDLE);
 
- 
-    IUGetConfigText(getDeviceName(), "TEMPERATURE_SNOOP", "TEMPERATURE_DEVICE", temperatureDevice, MAXINDIDEVICE);
-    IUGetConfigText(getDeviceName(), "TEMPERATURE_SNOOP", "TEMPERATURE_Property", temperatureProperty, MAXINDINAME);
-    IUGetConfigText(getDeviceName(), "TEMPERATURE_SNOOP", "TEMPERATURE_ATTRIBUTE", temperatureAttribute, MAXINDINAME);
-
    // TEMPERATURE_Property text case kept for backwards compatibility to previous version saved config files
     SnoopTemperatureDeviceTP[0].fill("TEMPERATURE_DEVICE", "Device", temperatureDevice);
     SnoopTemperatureDeviceTP[1].fill("TEMPERATURE_Property", "Property", temperatureProperty);
     SnoopTemperatureDeviceTP[2].fill("TEMPERATURE_ATTRIBUTE", "Attribute", temperatureAttribute);
     SnoopTemperatureDeviceTP.fill(getDeviceName(), "TEMPERATURE_SNOOP", "Temperature Device", OPTIONS_TAB, IP_RW, 0, IPS_IDLE);
 
-    
-    if (IUGetConfigOnSwitchIndex(getDeviceName(), "ENABLE_SNOOP_TEMPERATURE", &snoopTemperatureIndex) == 0) 
-        setSnoopTemperature = (snoopTemperatureIndex == INDI_ENABLED);
-    
     // Weather snoop setting for a Weather device ie: Weather station
     EnableSnoopWeatherSP[0].fill("ENABLE", "Enable", ISS_OFF);
     EnableSnoopWeatherSP[1].fill("DISABLE", "Disable", ISS_ON);
     EnableSnoopWeatherSP.fill(getDeviceName(), "ENABLE_SNOOP_WEATHER", "Snoop Weather", OPTIONS_TAB, IP_RW, ISR_1OFMANY, 0, IPS_IDLE);
-    
-    IUGetConfigText(getDeviceName(), "WEATHER_SNOOP", "WEATHER_DEVICE", weatherDevice, MAXINDIDEVICE);
-    IUGetConfigText(getDeviceName(), "WEATHER_SNOOP", "WEATHER_PROPERTY", weatherProperty, MAXINDINAME);
-    IUGetConfigText(getDeviceName(), "WEATHER_SNOOP", "TEMPERATURE_ATTRIBUTE", weatherTempAttribute, MAXINDINAME);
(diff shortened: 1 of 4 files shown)


drivers/auxiliary/cheapodc.cpp
```diff
@@ -112,61 +112,90 @@ bool CheapoDC::initProperties()
...
+    IUGetConfigText(getDeviceName(), "WEATHER_SNOOP", "WEATHER_DEVICE", weatherDevice, MAXINDIDEVICE);
+    IUGetConfigText(getDeviceName(), "WEATHER_SNOOP", "WEATHER_PROPERTY", weatherProperty, MAXINDINAME);
+    IUGetConfigText(getDeviceName(), "WEATHER_SNOOP", "TEMPERATURE_ATTRIBUTE", weatherTempAttribute, MAXINDINAME);
+    IUGetConfigText(getDeviceName(), "WEATHER_SNOOP", "HUMIDITY_ATTRIBUTE", weatherHumidityAttribute, MAXINDINAME);
+    
+    SnoopWeatherDeviceTP[0].fill("WEATHER_DEVICE", "Device", weatherDevice);
+    SnoopWeatherDeviceTP[1].fill("WEATHER_PROPERTY", "Property", weatherProperty);
+    SnoopWeatherDeviceTP[2].fill("TEMPERATURE_ATTRIBUTE", "Temp Attribute", weatherTempAttribute);
+    SnoopWeatherDeviceTP[3].fill("HUMIDITY_ATTRIBUTE", "Humidity Attribute", weatherHumidityAttribute);
+    SnoopWeatherDeviceTP.fill(getDeviceName(), "WEATHER_SNOOP", "Weather Device", OPTIONS_TAB, IP_RW, 0, IPS_IDLE);
```
knro, line 170: You can simplify this, no need to use IUGetConfigText. Instead, just add this after .fill line

`SnoopWeatherDeviceTP.load();`

This should load the property from config file.

hcomet:
I noticed the CheapoDC listing on indilib.org went missing during the website migration so resubmitted.

drivers/auxiliary/cheapodc.cpp
```diff
@@ -528,6 +575,16 @@ bool CheapoDC::Ack()
...
     FWversionTP.apply();
 
+    if (resp[0] == '1')
+    {
+        fwVOneDetected = true;
+    }
+    else
+    {
+        fwVOneDetected = false;
+    }
```
knro, line 585: Can be simplified to:

`fwVOneDetected = resp[0] == '1';`

knro:
> I noticed the CheapoDC listing on indilib.org went missing during the website migration so resubmitted.

@naheedsa Please check

drivers/auxiliary/cheapodc.cpp
hcomet: Should I be looking to stop using "IUGetConfig..." in general? For instance replacing
```
if (IUGetConfigOnSwitchIndex(getDeviceName(), "ENABLE_SNOOP_TEMPERATURE", &snoopTemperatureIndex) == 0) 
        setSnoopTemperature = (snoopTemperatureIndex == INDI_ENABLED);
```
with something like
```
EnableSnoopTemperatureSP.load();
snoopTemperatureIndex = EnableSnoopTemperatureSP.findOnSwitchIndex();
setSnoopTemperature = snoopTemperatureIndex == INDI_ENABLED;
```

hcomet:
Sorry another question. Looking at the suggested simplification of using Property.load(). There is a call to loadConfig(); in CheapoDC::updateProperties(). I assume the best thing would be to not do any calls to Property.load() in CheapoDC::initProperties(). Also move the couple of EnableSnoop... switch checks to after the loadConfig();

Assuming this would be more correct. Some of what happens when copying other drivers.

hcomet:
Made the requested simplification changes plus moved the loads to updateProperties.

hcomet:
Also saw that my drive submission on indilib.org was accepted. I must have done something wrong when submitting the description HTML because it is not rendering properly.

knro:
Property.load does NOT call ISNewXXX if that's what you mean, it simply initializes the property values.

drivers/auxiliary/cheapodc.h
```diff
@@ -104,41 +116,66 @@ class CheapoDC : public INDI::DefaultDevice
...
+    unsigned int previousTemperatureMode = WEATHER_QUERY;
+    int snoopWeatherIndex = INDI_DISABLED; // Default is Disabled
+    int previousSnoopWeatherIndex = snoopWeatherIndex;
     char locationDevice[MAXINDIDEVICE] = {"Telescope Simulator"};
     char locationProperty[MAXINDINAME] = {"GEOGRAPHIC_COORD"};
     char locationLatAttribute[MAXINDINAME] = {"LAT"};
     char locationLongAttribute[MAXINDINAME] = {"LONG"};
     char temperatureDevice[MAXINDIDEVICE] = {"Focuser Simulator"};
     char temperatureProperty[MAXINDINAME] = {"FOCUS_TEMPERATURE"};
     char temperatureAttribute[MAXINDINAME] = {"TEMPERATURE"};
```
knro, line 167: What do you need these for? GEOGRAPHIC_COORD and FOCUS_TEMPERATURE are both **standard** INDI properties. If you want to snoop other devices, you need to have a single property [ActiveDeviceTP](https://github.com/indilib/indi/blob/9d6e87abfddec3d79a4ddf34b0b693f542eb6fb8/libs/indibase/indiccd.cpp#L491) for all the different types of devices you need to snoop.

If you check the example above, you'll see how INDI::CCD snoops the properties from other devices.

hcomet: Looking back likely not understanding how ActiveDeviceTP works. I will look at it.

Thanks Jasem.

hcomet: Looking at INDI:CCD, I see that it uses IUGetConfigText to fill the intial text for the ActiveDeviceTP property with any values that may be  found in the config file. This is similar to what I intended above. Moving to using only Property.load(); removes this function.  Is there a better or preferred way to do this?

https://github.com/indilib/indi/blob/9d6e87abfddec3d79a4ddf34b0b693f542eb6fb8/libs/indibase/indiccd.cpp#L479C1-L498C1

knro: You're right, this code is outdated and need to be removed.

knro: INDI::CCD code is updated.

hcomet: Updated to using ActiveDeviceTP property.

Will update the driver listing on indilib.org on Monday.

PR #2147, opened 2024-11-05
Sim crash fix - looks like found root cause

artyom-beilis:
It was uninitialised value of mcRate
I added initialisation boundary checking and print in case something goes wrong

But seems to be root cause was missing initialization since I saw invalid values reported and after checking in Valgrind it complained at these spots.

### DIFF
--- a/drivers/telescope/scopesim_helper.cpp
+++ b/drivers/telescope/scopesim_helper.cpp
@@ -190,16 +190,20 @@ void Axis::update()         // called about once a second to update the position
         //LOGF_DEBUG("move %s: change %f, position %f", b, change, position.Degrees());
     }
 
+    int rate = std::max(-4,std::min(4,mcRate));
+    if(rate != mcRate)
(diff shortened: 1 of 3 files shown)


drivers/telescope/scopesim_helper.cpp
```diff
@@ -190,16 +190,20 @@ void Axis::update()         // called about once a second to update the position
         //LOGF_DEBUG("move %s: change %f, position %f", b, change, position.Degrees());
     }
 
+    int rate = std::max(-4,std::min(4,mcRate));
+    if(rate != mcRate)
+        fprintf(stderr,"Invalid mcRate Detected = %d\n",mcRate);
```
knro, line 195: LOGF_ERROR would be appropriate here.

artyom-beilis: Ok I'll fix later today

PR #2161, opened 2025-01-19
Feature for ZWO AM5 - Heavy Duty Mode and Fix slew rate

pawel-soja:
Hi

I often use a light and heavy tube on the ZWO AM5 mount.
With a heavier load, e.g. Celestron EdgeHD 800, or at low temperatures, the mount stops when aiming at the target or parking. For this reason, I propose a similar feature which is in the "SkyAtlas" program.
image

The mode enable/disable is now located in `INDI Control Panel -> ZWO AM5 USB -> Motion Control -> Heavy Duty Mode`
image

AM5 has a special command `:SRl720#` (720x speed) and `:SRl1440#` (1440x speed) which sets the maximum speed limit, regardless of other settings. To read, use the command `:GRl#`, which has also been implemented. The mount remembers the settings, so after a restart of the mount they are remembered.

I also added a fix for the slew rate selection, because the mount has 9 modes, not 10 as declared in the `SLEW_MODES` constant, as well as the `:Rx#` command where x takes values ​​from 1 to 9.

image

### DIFF
--- a/drivers/telescope/lx200am5.h
+++ b/drivers/telescope/lx200am5.h
@@ -106,11 +106,6 @@ class LX200AM5 : public LX200Generic
 
         // Heavy duty control
         INDI::PropertySwitch HeavyDutyModeSP {2};
-        enum
-        {
-            HeavyDutyModeOff = 0,
-            HeavyDutyModeOn  = 1,
-        };
 
         // Meridian Flip Control
         INDI::PropertySwitch MeridianFlipSP {2};
--- a/drivers/telescope/lx200am5.h
+++ b/drivers/telescope/lx200am5.h
@@ -171,6 +171,6 @@ class LX200AM5 : public LX200Generic
         // Maximum buffer for sending/receiving.
         static constexpr const uint8_t DRIVER_LEN {64};
         // Slew Modes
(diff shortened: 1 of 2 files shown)


drivers/telescope/lx200am5.h
```diff
@@ -104,6 +104,14 @@ class LX200AM5 : public LX200Generic
             High
         };
 
+        // Heavy duty control
+        INDI::PropertySwitch HeavyDutyModeSP {2};
+        enum
+        {
+            HeavyDutyModeOff = 0,
+            HeavyDutyModeOn  = 1,
+        };
```
knro, line 113: For such toggles, it's better to remove the enum and use INDI_ENABLED (0) and INDI_DISABLED (1) which are used for other switches. It just makes all the enable/disable switches all consistent for that first element is always "Enabled" and 2nd element "Disabled".

pawel-soja: ok I will correct it

pawel-soja:
Now it is `Enabled`/`Disabled` and I used `INDI_ENABLED`/`INDI_DISABLED`.

image

knro:
Thank you! Looks great!

PR #2171, opened 2025-02-21
Add driver for cloudwatcher solo from lunaticoastro

peter-englmaier:
Hi Jasem

I have a cloudwatcher rain/cloud sensor attached to a solo, both from lunaticoastro.com. The solo also works with your more generic driver `weatherwatcher.cpp`, but my new driver attempts to make configuration easier. It is based on your `weatherwatcher.cpp`.

This is a draft version of the driver - it works, but I would like to get some feedback and hopefully answers to the following questions.

I am puzzled by the lack of an `unsetCriticalParameter` method. When parameters are made critical or uncritical, it seems, one **always** needs to save parameters and restart the driver. Correct? For me, this seems not very intuitive. Perhaps the critical parameters should be made read-only with setPermission in 'connect' and be made read-write with setPermission when the driver is disconnect? Is this allowed?

Best, Peter

### DIFF
--- a/drivers/weather/CMakeLists.txt
+++ b/drivers/weather/CMakeLists.txt
@@ -23,13 +23,13 @@ add_executable(indi_watcher_weather ${weatherwatcher_SRC})
 target_link_libraries(indi_watcher_weather indidriver ${CURL})
 install(TARGETS indi_watcher_weather RUNTIME DESTINATION bin)
 
-# ########## Solo Weather Station ###############
-SET(solo_SRC
-    solo.cpp)
+# ########## AAGSolo Weather Station ###############
+SET(aagsolo_SRC
+    aagsolo.cpp)
 
-add_executable(indi_solo ${solo_SRC})
-target_link_libraries(indi_solo indidriver ${CURL})
-install(TARGETS indi_solo RUNTIME DESTINATION bin)
+add_executable(indi_aagsolo ${aagsolo_SRC})
+target_link_libraries(indi_aagsolo indidriver ${CURL})
+install(TARGETS indi_aagsolo RUNTIME DESTINATION bin)
 
 # ########## Weather Safety Proxy ###############
 SET(weathersafetyproxy_SRC
--- a/drivers/weather/CMakeLists.txt
+++ b/drivers/weather/CMakeLists.txt
@@ -27,9 +27,9 @@ install(TARGETS indi_watcher_weather RUNTIME DESTINATION bin)
 SET(aagsolo_SRC
     aagsolo.cpp)
 
-add_executable(indi_aagsolo ${aagsolo_SRC})
-target_link_libraries(indi_aagsolo indidriver ${CURL})
-install(TARGETS indi_aagsolo RUNTIME DESTINATION bin)
+add_executable(indi_aagsolo_weather ${aagsolo_SRC})
+target_link_libraries(indi_aagsolo_weather indidriver ${CURL})
+install(TARGETS indi_aagsolo_weather RUNTIME DESTINATION bin)
 
 # ########## Weather Safety Proxy ###############
 SET(weathersafetyproxy_SRC
--- a/drivers/weather/CMakeLists.txt
+++ b/drivers/weather/CMakeLists.txt
@@ -23,6 +23,14 @@ add_executable(indi_watcher_weather ${weatherwatcher_SRC})
 target_link_libraries(indi_watcher_weather indidriver ${CURL})
 install(TARGETS indi_watcher_weather RUNTIME DESTINATION bin)
 
+# ########## Solo Weather Station ###############
+SET(solo_SRC
+    solo.cpp)
+
+add_executable(indi_solo ${solo_SRC})
+target_link_libraries(indi_solo indidriver ${CURL})
+install(TARGETS indi_solo RUNTIME DESTINATION bin)
+
 # ########## Weather Safety Proxy ###############
 SET(weathersafetyproxy_SRC
     weather_safety_proxy.cpp)
--- a/drivers/weather/aagsolo.cpp
+++ b/drivers/weather/aagsolo.cpp
@@ -1,7 +1,7 @@
 /*******************************************************************************
   Copyright(c) 2025 Jasem Mutlaq and Peter Englmaier. All rights reserved.
 
-  INDI Solo Watcher Driver
+  INDI AAG Solo CloudWatcher Driver
 
   This program is free software; you can redistribute it and/or modify it
   under the terms of the GNU General Public License as published by the Free
@@ -22,7 +22,7 @@
   file called LICENSE.
 *******************************************************************************/
 
-#include "solo.h"
+#include "aagsolo.h"
 #include "locale_compat.h"
 
 #include <curl/curl.h>
@@ -31,8 +31,8 @@
 #include <cstring>
 
 
-// We declare an auto pointer to Solo.
-static std::unique_ptr<Solo> solo(new Solo());
+// We declare an auto pointer to AAGSolo.
+static std::unique_ptr<AAGSolo> aag_solo(new AAGSolo());
 
 static size_t write_data(void *contents, size_t size, size_t nmemb, void *userp)
 {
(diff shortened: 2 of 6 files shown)


knro:
Thank you for the driver.  What are exactly the issues with using weatherwatcher? can you elaborate?

Properties permission cannot be updated on the fly. But it's possible after you deleted the properties to update their permissions and then define them again.

peter-englmaier:
The more **generic** weatherwatcher driver allows me to monitor any URL or any file and look for key/value pairs. This is great, but I had some difficulty getting it to work. I had to find out the URL to call (exact path), look up the file and figure out how the driver is supposed to use it. While figuring out how the whole alerting works, I was also juggling with the property labels, and how to get the critical parameters to work and show up in kstars and understand how they are used. In the end, I was reading the source code, just to get definitive answers for all my questions. ;-)

With the new driver, one needs to set is the FQDN or IP of the solo device, but not the URL and the correct property labels. Otherwise, the two drivers do not differ.

To make using the new driver also more simple, I will try to delete the properties, as you suggested.

knro:
Can you suggest an update to to README and documentation to make this driver more usable? I mean the WeatherWatcher driver. Since you've already been through is, it would be great if you can add a summary and/or guide to the README for prospective users.

peter-englmaier:
Yes, I can add to the README of weatherwatcher. Where do I find the file to edit?

I now, made some more changes to the new driver, and moved the "criticality" flags to the options tab:
image
The options tab is always visible, even when the driver is not connected.
When changes are made while the driver is connected, a warning is displayed:

**Changing criticals requires driver reconnect**

My original plan, to make switches readonly while connected did not work. Giving the users a heads-up with a warning message seems to be next best solution to me. With the flags on the options tab, this is also more logical.

The criticality switches allow to specify which measurements must be in OK range to consider the observatory safe. The 'safe' switch is the overal status reported by the solo itself. Solo has its own configuration screen, so users might want to use the 'safe' value from solo instead of specifying ok ranges in the indi driver.

The driver can now be merged, if you agree. I can also write some documentation for how to use the driver with the solo.

peter-englmaier:
I suggest to following text for this page:
https://www.indilib.org/weather/weather-watcher.html

# Features

Weather Watcher driver is a generic driver for accessing weather data from different sources. Data can be collected from a local weather station or a remote weather provider. The Weather Watcher driver just needs a file or an URL with the data listed in lines like:

temp=18.100000

In this example,  'temp' is the name of the parameter, '=' is the seperator, and '18.1' is the value. Lines not matching this format are ignored by the driver.

Please first find out the file name or URL to use. This information should be found in the product manual.
E.g. /home/weather/data.txt, or http://192.168.1.1:10800/cgi-bin/cgiLastData

Below, we use the SOLO device as an example. However, if you do own a SOLO, please use the indi_solo driver instead. If you have a cloudwatcher without SOLO, please use the aagcloudwatcher driver.

# Operation

The SOLO provides weather information at the URL 'http://192.168.1.1:80/cgi-bin/cgiLastData' (replace 192.168.1.1 with the URL or hostname of your solo device. You may also specify a different port after the colon). To test the access, open the URL in a browser. Without asking for a password, it should display something like this:

clouds=-0.140000
cloudsSafe=2
temp=9.870000
wind=-1
windSafe=1
gust=-1
rain=3200
rainSafe=1
lightmpsas=16.85
lightSafe=1
switch=0
safe=0
hum=80
humSafe=1

Not all of these parameters need to be present and only a few are of interest. All data are treated as numeric, that is WeatherWatcher driver doesn't make assumption on unit of measurement. So your wind data can be in Kph or Mph, only the threshold has to be updated.

The driver can read several of these parameters and clients (i.e. kstars/ekos), will display the data. The driver has min/max values for deciding, if the value is in a good, warn, or alert state. In addition, the driver can classify some parameters as 'critical'. When any such critical parameter is outside the min/max range, the driver will report 'unsafe' condition and the client can abort and shutdown the observatory. The scheduler in Kstars/ekos in particular, will only warn you when parameters are in warn state, without changing its behaviour, but abort imaging and shutdown observatory when any critical parameter is in 'alert' state.

# Configuration

First, open the indi device panel and go to options tab. Here you can specfy the URL or file to monitor in 'Source / File'. 
Next, you can specify the parameters (keywords) you want to be found in the url/file. Just specify the ones you are interested in.
Press 'save', to store your changes.

Next, go to the 'main control' tab and connect the driver.

Then, go to the 'Parameters' tab and select the parameters you consider 'critical'. You should also specify the min and max range of OK values. Values outside this range are considered dangerous (alert). Please note, that order is importent. Min value must be smaller than max value! You can also specify a warn percentage. The purpose of the warn percentage is to get you a warning when the value gets to close to the boundary of the ok range. The following graphic visualizes the range for the temperature parameter:

[ insert graphic with temperature range ]

Note: if the minimum of the parameters is set to zero, a lower warning is not issued because the driver assumes that the parameter is positive only (e.g. Wind).


Next, go back to 'options' panel and save your configuration again. Then go to 'main control' and disconnect and then reconnect the driver. The panel should now display the status of all the critical parameters (green/yellow/red light) as well as the overall status.

Please note, whenever you want to change a parameter to be a critical parameter or not, you have to save settings, then reconnect the driver. Other changes become active immediately.

**Important note:** Please test the configuration in your setup before allowing the observatory to operate automatically.

naheedsa:
> Thanks! @naheedsa Please add to documentation

Are thre any screenshots or images so we can also add that?

knro:
Not at the moment.

knro:
Thanks, a couple of things:
1. Need a better driver name, solo is quite generic.
2. Need an entry in drivers.xml

peter-englmaier:
The device is advertised as "AAG Solo CloudWatcher by Lunático" and the documention keeps caling it 'solo' (after Han Solo, of course).

I would suggest the name aagsolo. The default device hostname on the network is also aagsolo.
Would that be acceptable?

Driver.xml is now included.

drivers/weather/CMakeLists.txt
```diff
@@ -23,6 +23,14 @@ add_executable(indi_watcher_weather ${weatherwatcher_SRC})
 target_link_libraries(indi_watcher_weather indidriver ${CURL})
 install(TARGETS indi_watcher_weather RUNTIME DESTINATION bin)
 
+# ########## AAGSolo Weather Station ###############
+SET(aagsolo_SRC
+    aagsolo.cpp)
+
+add_executable(indi_aagsolo ${aagsolo_SRC})
```
knro, line 30: Traditionally, it should be indi_aagsolo_weather to be consistent with other INDI drivers.

drivers/weather/aagsolo.cpp
```diff
@@ -0,0 +1,388 @@
+/*******************************************************************************
+  Copyright(c) 2025 Jasem Mutlaq and Peter Englmaier. All rights reserved.
+
+  INDI AAG Solo CloudWatcher Driver
```
knro, line 4: Update copyright since you wrote this driver.

drivers/weather/aagsolo.cpp
```diff
@@ -0,0 +1,388 @@
...
+    criticalSP[7].fill("CRITICAL_8", "Light", ISS_OFF);
+    criticalSP[8].fill("CRITICAL_9", "Safe", ISS_OFF);
+    criticalSP.fill(getDeviceName(), "CRITICALS", "Criticals", OPTIONS_TAB, IP_RW, ISR_NOFMANY, 0, IPS_IDLE);
+    defineProperty(criticalSP);
+    loadConfig(true, "CRITICALS");
+
+    soloInfoTP[0].fill("CWINFO", "cwinfo", nullptr);
+    soloInfoTP[1].fill("DATATIME", "GMT Time", nullptr);
+    soloInfoTP.fill(getDefaultName(), "DEVICEINFO", "Device Info", INFO_TAB, IP_RO, 60, IPS_IDLE);
+    defineProperty(soloInfoTP);
```
knro, line 222: no defineProperty should be called in initProperties. It should be called in updateProperties.

Also, for:

```criticalSP.fill(getDeviceName(), "CRITICALS", "Criticals", OPTIONS_TAB, IP_RW, ISR_NOFMANY, 0, IPS_IDLE);```

You can simply add after that ```critialSP.load()``` and it would load the config, no need to explicitly call load config.

drivers/weather/CMakeLists.txt
peter-englmaier: OK, will change that.

peter-englmaier:
Hi Jason
Tested successfully with my equipment and latest published versions of kstars and indilib.
Anything else should improve?

peter-englmaier:
> > Thanks! @naheedsa Please add to documentation
> 
> Are thre any screenshots or images so we can also add that?

@naheedsa 
The documentation above is for the indi_watcher_weather driver.  I can certainly improve on that. Also, I need to write documentation for the new indi_aagsolo_weather.

knro:
@peter-englmaier Please send the documentation as well for the new drivers, thanks!

naheedsa:
> indi_aagsolo_weather

Hello @peter-englmaier , the site url you provided is unreachable now. Can you please check again or provide us an PDF file?

naheedsa:
Requires password.

Screenshot_20250420_081306

peter-englmaier:
> > indi_aagsolo_weather
> 
> Hello @peter-englmaier , the site url you provided is unreachable now. Can you please check again or provide us an PDF file?

@naheedsa 
Oh, sorry. I removed the password requirement. 
https://erfa.englmaier.ch/index.php/s/7c88LxeA8Frg28W
seems to work now!

PR #2176, opened 2025-03-01
Improve readability of the indiserver

SloppyJaconda:
`indiserver.cpp` is just a single file, which is a nightmare for readability and maintainability purposes.

This PR seeks to break `indiserver.cpp` into component files. This means header/cpp for each of the classes, a set of utils files for the various static functions, and a `constants.hpp` for some of the `#defines` and other variables that are shared across the classes.

While there have not been any logic changes to this changes, I posted a discussion over in #2175 to get better guidance on the general process indi devs use for validating their changes to the indiserver. 

~~I'm looking to make changes to the workflow files to make the `github.repository_owner` variable lowercase so the github action validation doesn't fail for people with upper case letters in their username (like me). I'll post that in a future commit.~~ I figured out why the workflows didn't work on my fork, it was looking for the name `indilib` for the `ghcr.io` url, rather than the repo owner's name.


indiserver/ClInfo.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+    /* snag enableBLOB -- send to remote drivers too */
+    if (!strcmp(roottag, "enableBLOB"))
+        crackBLOBHandling(dev, name, pcdataXMLEle(root));
+
+    if (!strcmp(roottag, "pingRequest"))
+    {
+        setXMLEleTag(root, "pingReply");
+
+        Msg * mp = new Msg(this, root);
```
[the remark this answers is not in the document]
SloppyJaconda: I realize that in the `queuingDone` method there is a `delete(this)` that is used however that is not a good pattern to use.

indiserver/ClInfo.cpp
```diff
@@ -0,0 +1,323 @@
...
+    fflush(stderr);
+#endif
+}
+void ClInfo::q2Clients(ClInfo *notme, int isblob, const std::string &dev, const std::string &name, Msg *mp, XMLEle *root)
+{
+    /* queue message to each interested client */
+    for (auto cpId : clients.ids())
+    {
+        auto cp = clients[cpId];
+        if (cp == nullptr) continue;
```
SloppyJaconda, line 113: This `ConcurrentSet<T>` is an odd design and it's not exactly clear what this container is doing that cannot be otherwise accomplished by using `std::map` instead.

indiserver/ClInfo.cpp
```diff
@@ -0,0 +1,323 @@
...
+                continue;
+        }
+
+        /* shut down this client if its q is already too large */
+        unsigned long ql = cp->msgQSize();
+        if (isblob && maxstreamsiz > 0 && ql > maxstreamsiz)
+        {
+            // Drop frames for streaming blobs
+            /* pull out each name/BLOB pair, decode */
+            XMLEle *ep      = NULL;
```
SloppyJaconda, line 152: `nullptr` should be used instead of `NULL`.

indiserver/Constants.hpp
```diff
@@ -0,0 +1,42 @@
...
+#endif
+
+extern int verbose;
+extern unsigned int maxstreamsiz;
+extern unsigned int maxqsiz;
+extern char *ldir;
+extern int maxrestarts;
+
+class Fifo;
+extern Fifo * fifo;
```
SloppyJaconda, line 42: These externs were just added as a means of sharing them temporarily between the files. With some light refactoring of the locations they're in use, these can all be removed.

indiserver/Constants.hpp
```diff
@@ -0,0 +1,42 @@
...
+#define MAXWSIZ       49152 /* max bytes/write */
+#define SHORTMSGSIZ   2048  /* buf size for most messages */
+#define DEFMAXQSIZ    128   /* default max q behind, MB */
+#define DEFMAXSSIZ    5     /* default max stream behind, MB */
+#define DEFMAXRESTART 10    /* default max restarts */
+#define MAXFD_PER_MESSAGE 16 /* No more than 16 buffer attached to a message */
+#ifdef OSX_EMBEDED_MODE
+#define LOGNAME  "/Users/%s/Library/Logs/indiserver.log"
+#define FIFONAME "/tmp/indiserverFIFO"
+#endif
```
SloppyJaconda, line 33: Unless there's some concrete reason why these are `#define` we should instead make them `constexpr`

indiserver/ConcurrentSet.hpp
```diff
@@ -0,0 +1,128 @@
...
+    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+*/
+
+#pragma once
+
+#include <map>
+#include <vector>
+
+template<class M>
+class ConcurrentSet
```
SloppyJaconda, line 26: As I mentioned in another comment, I don't see what this is accomplishing that couldn't otherwise be done with a standard container.

indiserver/DvrInfo.cpp
```diff
@@ -0,0 +1,325 @@
...
+    {
+        log(fmt("read <%s device='%s' name='%s'>\n",
+                tagXMLEle(root), findXMLAttValu(root, "device"), findXMLAttValu(root, "name")));
+    }
+
+    /* that's all if driver is just registering a snoop */
+    /* JM 2016-05-18: Send getProperties to upstream chained servers as well.*/
+    if (!strcmp(roottag, "getProperties"))
+    {
+        this->addSDevice(dev, name);
```
SloppyJaconda, line 47: `dev` is not checked for nullness before this is used. Again `basic_string` cannot be constructed from nullptr and this will crash.

indiserver/DvrInfo.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+        mp->queuingDone();
+
+        return;
+    }
+
+    /* that's all if driver desires to snoop BLOBs from other drivers */
+    if (!strcmp(roottag, "enableBLOB"))
+    {
+        Property *sp = findSDevice(dev, name);
```
SloppyJaconda, line 64: `name` is also not checked for null

SloppyJaconda (commented):
I went to a few of the files that were created by splitting up indiserver.cpp and gave a quick review, pointing out areas that could be improved for code safety and simplicity.

SloppyJaconda:
Some of the future improvements I would like to make, assuming you are open to these sorts of PRs would be to improve the memory ownership model and to modernize some of the design paradigms.

indiserver/ClInfo.cpp
```diff
@@ -0,0 +1,323 @@
...
+        q2Clients(this, isblob, dev, name, mp, root);
+    }
+
+    mp->queuingDone();
+}
+
+void ClInfo::close()
+{
+    if (verbose > 0)
+        log("shut down complete - bye!\n");
```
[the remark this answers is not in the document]
knro: agreed. We use the same style in INDI logger, which should probably be used here.

knro:
Absolutely! This is great, but since it touches the very core of INDI, it needs to be reviewed and tested by multiple core developers. 

@pludov made significant improvements to the INDI server in the last time it was updated. It would be if Ludovic can also review the changes.

SloppyJaconda:
> it needs to be reviewed and tested by multiple core developers.
@knro absolutely. This isn't a change that I'd expect to be rushed through. I was wondering though, for these kinds of core changes is there a specific test suite you all run against it to ensure that it behaves as expected? It doesn't appear that the GoogleTest suite hits any of the indiserver functionalities. Also I'd be more than happy to contribute towards a test suite if there are any docs on the expected behaviors that I can model the tests after.

pludov:
Hi ! Glad to see that happening !  I'll happily review ...

I felt the job was not finished when I added shared memory blobs, and I had in mind to someday reorg indiserver.cpp. however, in the meantime, I've learned Rust and I think it could be a good fit for a next indiserver ( memory safety and performance of course, but also the ecosystem provides many bits we need here , so the actual code could be shrinked by much...)

There is a test suite dedicated to indiserver in the integs directory.  It validates basic startup, parsing, routing, blobs, ...

indiserver/ClInfo.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+    const char *dev  = findXMLAttValu(root, "device");
+    const char *name = findXMLAttValu(root, "name");
+    int isblob       = !strcmp(tagXMLEle(root), "setBLOBVector");
+
+    /* snag interested properties.
+     * N.B. don't open to alldevs if seen specific dev already, else
+     *   remote client connections start returning too much.
+     */
+    if (dev[0])
```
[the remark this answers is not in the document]
SloppyJaconda: resolving since these are going to be addressed in future patches.

indiserver/ClInfo.cpp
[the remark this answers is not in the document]
SloppyJaconda: resolving since these are going to be addressed in future patches.

SloppyJaconda:
Resolved all of my comments since they are going to be addressed in future patches.

pludov:
That looks good to me ! Test suite didn't suffer :+1:

knro:
Awesome work @SloppyJaconda !

PR #2177, opened 2025-03-02
Feat(): alpaca dome and alpaca weather safety

grm:
I added two alpaca devices that can be configured.
I did not have the opportunity to test every commands for the alpaca dome because the remote site I use only permits opening the roof.
I guess others should also work.

### DIFF
--- a/drivers.xml
+++ b/drivers.xml
@@ -953,6 +953,10 @@
             <driver name="Uranus Meteo Sensor">indi_uranus_weather</driver>
             <version>1.0</version>
         </device>
+        <device label="Weather Safety Alpaca" manufacturer="Others">
+            <driver name="Weather Safety Alpaca">indi_weather_safety_alpaca</driver>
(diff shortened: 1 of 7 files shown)


knro:
Thank you! Why is there a need for a separate weather safety driver specifically for Alpaca?

drivers.xml
```diff
@@ -949,6 +953,10 @@
             <driver name="Uranus Meteo Sensor">indi_uranus_weather</driver>
             <version>1.0</version>
         </device>
+        <device label="Weather Safety Alpaca" manufacturer="Others">
```
knro, line 956: Maybe change manufacturer to Alpaca? I know it's not technically a manufacturer but it would be nice grouping for all future alpaca-compatible drivers.

grm: Perfect, I'll do that :)

grm: Done, i have changed for both :)

PR #2181, opened 2025-03-03
Fix(): Test fixing pyindi-client build

grm:


grm:
Hello

This is a fix to make pyindi test pass.

ia m not sure to have kept the original philosophy.
I had to : 
- run black before testing because the code was not formatted properly
- Add an env variable to deactivate gsc before running test. There may be a better solution without deactivating GSC :)

knro:
Wow thank you so much @grm 

Feels good to see everything in the green again, Bravo!

PR #2190, opened 2025-03-06
Fix(): Correct UPB V2 auto dew at startup

grm:


knro:
Normally client load the driver config are connection is established, so why is this necessary?

grm:
I am not sure why but the chackboxes were not rechecked with the current driver implementation.
Let me do another check tonight with the driver built from the master branch

grm:
You are right, it is working, not sure why it did not worked before :)

PR #2192, opened 2025-03-11
Fix build warning - Removing clang-extensions for variable length arrays

SloppyJaconda:
This PR addresses warnings regarding clang extensions for dynamic arrays. 
`warning: variable length arrays in C++ are a Clang extension [-Wvla-cxx-extension]`
I replaced these locations with `std::vector` so we don't have to deal with new/delete.

Some of the `uint8_t` seem like they would be better off as a `std::byte`, but I'll leave that unless someone says otherwise.

### DIFF
--- a/drivers/dome/scopedome_usb21.cpp
+++ b/drivers/dome/scopedome_usb21.cpp
@@ -27,6 +27,7 @@
 
 #include <termios.h>
 #include <unistd.h>
+#include <vector>
 
 #define SCOPEDOME_TIMEOUT 2
 
@@ -267,7 +268,7 @@ int ScopeDomeUSB21::writeBuf(Command cmd, uint8_t len, uint8_t *buff)
     int nbytes_written = 0, rc = -1;
     char errstr[MAXRBUF];
 
-    uint8_t cbuf[BytesToWrite];
+    std::vector<uint8_t> cbuf(BytesToWrite, 0);
     cbuf[0] = header;
     cbuf[3] = CRC(0, cbuf[0]);
     cbuf[1] = len;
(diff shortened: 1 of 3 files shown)


drivers/dome/scopedome_usb21.cpp
```diff
@@ -326,12 +327,11 @@ int ScopeDomeUSB21::readBuf(Command &cmd, uint8_t len, uint8_t *buff)
...
     int nbytes_read = 0, rc = -1;
     int BytesToRead = len + 4;
-    uint8_t cbuf[BytesToRead];
-    memset(cbuf, 0, BytesToRead);
+    std::vector<uint8_t> cbuf(BytesToRead, 0);
     char errstr[MAXRBUF];
 
     // Read buffer
-    if ((rc = tty_read(PortFD, (char *)cbuf, sizeof(cbuf), SCOPEDOME_TIMEOUT, &nbytes_read)) != TTY_OK)
+    if ((rc = tty_read(PortFD, (char *)cbuf.data(), sizeof(cbuf), SCOPEDOME_TIMEOUT, &nbytes_read)) != TTY_OK)
```
knro, line 334: Are you sure this can work?? I don't think sizeof would work here. Maybe cbuf.size

Also, if we use this:

```uint8_t cbuf[BytesToWrite] = {0};```

Wouldn't that be sufficient?

SloppyJaconda: You're right, I didn't check the sizeof calculation. Let me update that
The problem is that `BytesToWrite` isn't known at compile-time. So `uint8_t cbuf[BytesToWrite] = {0};` isn't legal c++ still makes use of the Clang extension.

SloppyJaconda: updated to return the calculation of the size in bytes

PR #2193, opened 2025-03-12
Fix():  Wanderer cover v3 is now working with the scheduler

grm:


knro:
We already have Light Panel simulator. Also yesterday, I added Dust Cover Simulator. So we don't need any more simulators.

grm:
Done, I removed the simulator

drivers/auxiliary/wanderer_cover.cpp
```diff
@@ -63,6 +63,7 @@ WandererCover::WandererCover() : LightBoxInterface(this), DustCapInterface(this)
 
 bool WandererCover::initProperties()
 {
+    LOG_INFO("Initializing Wanderer Cover properties...");
```
knro, line 66: Do you really need to log this? Seems this was used during development?

drivers/auxiliary/wanderer_cover.cpp
```diff
@@ -114,11 +115,13 @@ bool WandererCover::initProperties()
     });
     registerConnection(serialConnection);
 
+    LOG_INFO("Properties initialization complete");
```
knro, line 118: Same

drivers/auxiliary/wanderer_cover.cpp
```diff
@@ -114,11 +115,13 @@ bool WandererCover::initProperties()
...
     return true;
 }
 
 bool WandererCover::Handshake()
 {
+    LOG_INFO("Starting handshake with device...");
```
knro, line 124: LOG_DEBUG

knro:
Please try to revise the LOG_XXX statements.

grm:
Done !

PR #2198, opened 2025-03-14
Needed to change libs/indibase/indiweatherinterface.* to allow alerts…

jmocenigo:
… for negative values (e.g., sky temperature).

```
Current percent warning model was like this:
    min--+         +-low-%    high-%-+        +--max
         |         |                 |        |
         v         v                 v        v
         [         (                 )        ]
danger     warning       good         warning   danger

For case where values below minimum value is actually good, needed to flip comparison to:
    min--+         +-low-%    high-%-+        +--max
         |         |                 |        |
         v         v                 v        v
         [         (                 )        ]
good       warning      danger        warning   good

```
Change I did will add extra line to EVERY parameter widget set to allow user to decide if alert behavior should be flipped or not (1/0). Probably some better way to make the developer decide when flipping option is desired, but I was not sure that could always always be known in advance. Default might be known, but user might need to change.

The added widget line appears below the widget line that sets the warning percentage.

### DIFF
--- a/drivers/weather/weatherwatcher.cpp
+++ b/drivers/weather/weatherwatcher.cpp
@@ -138,7 +138,7 @@ bool WeatherWatcher::createProperties()
             IUGetConfigNumber(getDeviceName(), "WEATHER_CLOUDS", "MAX_OK", &maxOK);
             IUGetConfigNumber(getDeviceName(), "WEATHER_CLOUDS", "PERC_WARN", &percWarn);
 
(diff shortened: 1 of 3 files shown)


knro:
Thanks, we definitely need to add this to the INDI documentation as well.

jmocenigo:
Do you mean doxygen in the code or somewhere else? If you are asking me to write that documentation , which I am happy to do, then please point me specifically to where I would do that as I am not up to speed on all the development processes (unless you just mean doxygen comments in the source code).

knro:
Doxygen is great, thanks. I'll incorporate this later to the official docs.

PR #2200, opened 2025-03-15
Fix(): wanderer cover positions are now saved when defined

grm:


drivers/auxiliary/wanderer_cover_v4_ec.cpp
```diff
@@ -330,7 +330,10 @@ bool WandererCoverV4EC::ISNewNumber(const char * dev, const char * name, double
 
             CloseSetNP.setState( (rc1) ? IPS_OK : IPS_ALERT);
             if (CloseSetNP.getState() == IPS_OK)
+            {
                 CloseSetNP.update(values, names, n);
+                saveConfig();
```
knro, line 335: saveConfig(CloseSetNP);

drivers/auxiliary/wanderer_cover_v4_ec.cpp
```diff
@@ -352,7 +355,10 @@ bool WandererCoverV4EC::ISNewNumber(const char * dev, const char * name, double
 
             OpenSetNP.setState( (rc1) ? IPS_OK : IPS_ALERT);
             if (OpenSetNP.getState() == IPS_OK)
+            {
                 OpenSetNP.update(values, names, n);
+                saveConfig();
```
knro, line 360: saveConfig(OpenSetNP);

drivers/auxiliary/wanderer_cover_v4_ec.cpp
grm: Done !

drivers/auxiliary/wanderer_cover_v4_ec.cpp
grm: Done !

PR #2203, opened 2025-03-16
Refactoring more indiserver components

SloppyJaconda:
The largest refactor here has to do with the constants.hpp file. I removed all of the `#define` values and moved them into their own namespace as `constexpr`s. I then took all of the externs and moved them into a `CommandLineArgs` structure and made a single extern called `userConfigurableArguments` so the values can be accessed through that. 

In the IndiServer.cpp file, I removed all of the explicit calls to new and replaced them with make_unique, storing the results in the main function scope to ensure proper cleanup of the resources at program exit. 

I replaced some virtual labels with override instead and changed `std::size_t` references to just `size_t`.

### DIFF
--- a/indiserver/CommandLineArgs.hpp
+++ b/indiserver/CommandLineArgs.hpp
@@ -17,17 +17,21 @@
     Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 */
 #pragma once
-#include <string_view>
-namespace indiserver::constants
-{
-constexpr unsigned indiPortDefault {7624};
-constexpr unsigned maxStringBufferLength {512};
-constexpr unsigned defaultMaxQueueSizeMB {128 * 1024 * 1024};
-constexpr unsigned defaultMaxStreamSizeMB {5 * 1024 * 1024};
-constexpr unsigned defaultMaximumRestarts {10};
+#include "Constants.hpp" 
 
-#ifdef OSX_EMBEDED_MODE
-constexpr std::string_view logNamePattern {"/Users/%s/Library/Logs/indiserver.log"};
-constexpr std::string_view fifoName {"/tmp/indiserverFIFO"}
-#endif
-} // namespace indiserver
+#include <string>
+
+class Fifo;
+
+struct CommandLineArgs {
+    size_t verbosity{0};
+    unsigned int maxStreamSizeMB{indiserver::constants::defaultMaxStreamSizeMB};
+    unsigned int maxQueueSizeMB{indiserver::constants::defaultMaxQueueSizeMB};
+    char *loggingDir{nullptr};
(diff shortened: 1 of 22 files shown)


SloppyJaconda:
Is there was any desire from the maintainers to specify a targeted C++ version on the repo at the CMake level? I'd like to add some C++20 stuff in future PRs, but I want to make sure that I'm not going to step on some toes if there's a requirement to keep it on an older version that I'm unaware of. 

Speaking of CMake, I noticed that there are varying `cmake_minimum_required` values specified across the repo. Would you like me to make a PR to align these (as well as the macros used to ensure that) or is there a reason that these are different?

indiserver/CommandLineArgs.hpp
```diff
@@ -0,0 +1,38 @@
...
+class Fifo;
+
+struct CommandLineArgs
+{
+    size_t verbosity{0};
+    unsigned int maxStreamSizeMB{indiserver::constants::defaultMaxStreamSizeMB};
+    unsigned int maxQueueSizeMB{indiserver::constants::defaultMaxQueueSizeMB};
+    char *loggingDir{nullptr};
+    int maxRestartAttempts{indiserver::constants::defaultMaximumRestarts};
+    Fifo* fifoHandle{nullptr};
```
pludov, line 33: This is not really a command line argument,  but rather a service instance built from it. (As Unix server, tcp server,... ) . Could you decouple the argument from the implementation ?

pludov:
Not much to say, good job!

indiserver/RemoteDvrInfo.cpp
```diff
@@ -20,18 +20,21 @@
...
+
 void RemoteDvrInfo::extractRemoteId(const std::string &name, std::string &o_host, int &o_port, std::string &o_dev) const
 {
     char dev[MAXINDIDEVICE] = {0};
-    char host[MAXSBUF] = {0};
+    char host[maxStringBufferLength] = {0};
 
     /* extract host and port from name*/
-    int indi_port = INDIPORT;
+    int indi_port = userConfigurableArguments->port;
```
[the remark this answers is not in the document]
SloppyJaconda: Ahh, got it. I'll switch that back

indiserver/CommandLineArgs.hpp
SloppyJaconda: Yeah I wasn't really sure if it belonged in here or separate. I'll move it out.

indiserver/RemoteDvrInfo.cpp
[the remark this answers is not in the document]
SloppyJaconda: resolved in 8e33aff

indiserver/CommandLineArgs.hpp
SloppyJaconda: Moved it out with 8e33aff

PR #2205, opened 2025-03-16
Major Fixes in skywatcherAltAz driver:

artyom-beilis:
1. Fixed incorrect tracking behavior - prevent negative trackRate and make sure it stays somewhat positive to prevent status change during tracking and makeing stacking smooth
2. Changed timeout to 300ms instead of 5s to prevent mount running away in case of UDP packet loss
3. Added movement to same direction after goto or for alignment - as synscan app does

### DIFF
--- a/drivers/telescope/skywatcherAPIMount.cpp
+++ b/drivers/telescope/skywatcherAPIMount.cpp
@@ -610,6 +610,12 @@ bool SkywatcherAPIMount::Goto(double ra, double dec)
     SilentSlewMode = (IUFindSwitch(&SlewModesSP, "SLEW_SILENT") != nullptr
                       && IUFindSwitch(&SlewModesSP, "SLEW_SILENT")->s == ISS_ON);
 
+    if(TrackState != SCOPE_SLEWING) {
+        long deltaAz  = DegreesToMicrosteps(AXIS1,AZ_BACKLASH_DEG);
+        long deltaAlt = DegreesToMicrosteps(AXIS2,ALT_BACKLASH_DEG);
+        AzimuthOffsetMicrosteps -= deltaAz;
+        AltitudeOffsetMicrosteps -= deltaAlt;
+    }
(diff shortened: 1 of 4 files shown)


drivers/telescope/skywatcherAPIMount.cpp
```diff
@@ -610,6 +610,12 @@ bool SkywatcherAPIMount::Goto(double ra, double dec)
     SilentSlewMode = (IUFindSwitch(&SlewModesSP, "SLEW_SILENT") != nullptr
                       && IUFindSwitch(&SlewModesSP, "SLEW_SILENT")->s == ISS_ON);
 
+    if(TrackState != SCOPE_SLEWING) {
+        long deltaAz  = DegreesToMicrosteps(AXIS1,AZ_BACKLASH_DEG);
+        long deltaAlt = DegreesToMicrosteps(AXIS2,ALT_BACKLASH_DEG);
+        AzimuthOffsetMicrosteps -= deltaAz;
+        AltitudeOffsetMicrosteps -= deltaAlt;
+    }
```
knro, line 618: What's the reason for this hard-coded backlash value?

artyom-beilis: I mostly mimic SynScan Pro app behaviour

- it always approaches from one direction
- you as user need to finish movement Up/Right before you can perform "sync" on an object (when aligning manually)
- And the values of back lash by default set to 0 but still the required approach is the same 

So I did it with constant value for simplicity as is the "fixed" behaviour in synscan app.

If you think it is required to use a parameter can you please guide me - what the parameter that is used for backlash and how do you configure one with proper default values - from what I understand it goes to XML, updates, merges and so on. I'm not really sure how to do it correctly.

knro: The current backlash in the header it set to 0.5 degrees... so won't this always offset the target by 0.5 degrees on both axis on each GOTO?

PR #2209, opened 2025-03-24
Ipx800

aknotwot:


knro:
Thanks, we already have an existing driver IPX800v4 driver. Do you need to make modifications against this?

aknotwot:
> Thanks, we already have an existing driver IPX800v4 driver. Do you need to make modifications against this?

Hi. Sorry. didn't add commentary. Only last commit is useful.  
I propose you some evolutions in IPX800v4 : API key is optionnal, the parse of JSON object was not working (change to object parsing). I changed a typo in drivers.xml.

aknotwot:
i can cancel pull request and , and redo it without ipx800.cpp and ipx800.h commits.

knro:
Sure please do. Only submit the fixes and changes to the existing driver. Btw, you think it would be easy to also add support the IPX800v3?

aknotwot:
> Sure please do. Only submit the fixes and changes to the existing driver. Btw, you think it would be easy to also add support the IPX800v3?

the IPX800v3 have a JSON API ....but not exactly the same syntax....It exists also IPX800v5 ..so i will first commit some ipx800v4 corrections, then i propose to rename the ipx800v4 to ipx800, and to work to adapt it to those two more device. But I own only a v4, so i can't test v3 and v5.

PR #2215, opened 2025-04-04
Fix issues affecting guide simulator

murveit:
Fixed two issues which prevented the guide simulator from working well.
1- The simulator snoops on the telescope simulator, which was not updating its RA coordinates often enough because the RA coordinate (in hours) was using an update threshold (meant for degrees).
2- The calculations for distance from center of a star was made more accurate using sub-pixel distances.

### DIFF
--- a/libs/indibase/inditelescope.cpp
+++ b/libs/indibase/inditelescope.cpp
@@ -682,7 +682,9 @@ void Telescope::NewRaDec(double ra, double dec)
         TrackStateSP.apply();
     }
 
-    if (std::abs(EqNP[AXIS_RA].getValue() - ra) > EQ_NOTIFY_THRESHOLD ||
+    // RA is in hours, so change the arc-second threshold accordingly.
+    constexpr double RA_NOTIFY_THRESHOLD = EQ_NOTIFY_THRESHOLD/15.0;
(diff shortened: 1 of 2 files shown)


libs/indibase/inditelescope.cpp
```diff
@@ -682,7 +682,9 @@ void Telescope::NewRaDec(double ra, double dec)
         TrackStateSP.apply();
     }
 
-    if (std::abs(EqNP[AXIS_RA].getValue() - ra) > EQ_NOTIFY_THRESHOLD ||
+    // RA is in hours, so change the arc-second threshold accordingly.
+    constexpr double RA_NOTIFY_THRESHOLD = EQ_NOTIFY_THRESHOLD/15.0;
+    if (std::abs(EqNP[AXIS_RA].getValue() - ra) > RA_NOTIFY_THRESHOLD ||
```
knro, line 687: Why introduce a variable? can't you just divide EQ_NOTIFY_THRESHOLD by 15.0 in the if statement?

murveit: A constexpr isn't a variable, it's a constant done at compile time. If you prefer I can switch this to the divide by 15. I thought it was more readable this way. Your preference.

PR #2233, opened 2025-05-02
Add INDI UPS Driver using NUT for monitoring UPS status

grm:
This commit introduces a new driver that integrates with Network UPS Tools (NUT) to monitor UPS status and expose it through INDI's weather interface. The driver includes properties for server address, UPS name, connection settings, and simulation controls, along with methods for querying UPS status and updating weather parameters accordingly.

### DIFF
--- a/drivers/auxiliary/ups.cpp
+++ b/drivers/auxiliary/ups.cpp
@@ -1,0 +1,577 @@
+/*******************************************************************************
+  Copyright(c) 2025 Jérémie Klein. All rights reserved.
+
+  INDI UPS Driver using NUT (Network UPS Tools)
+  This driver monitors a UPS through NUT and exposes its status through INDI's weather interface.
+  Battery level and power status are mapped to weather parameters for compatibility.
+
+  This program is free software; you can redistribute it and/or modify it
+  under the terms of the GNU General Public License as published by the Free
+  Software Foundation; either version 2 of the License, or (at your option)
+  any later version.
+
+  This program is distributed in the hope that it will be useful, but WITHOUT
+  ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+  FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+  more details.
+
+  You should have received a copy of the GNU Library General Public License
+  along with this library; see the file COPYING.LIB.  If not, write to
+  the Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+  Boston, MA 02110-1301, USA.
+
+  The full GNU General Public License is included in this distribution in the
+  file called LICENSE.
+*******************************************************************************/
+
+#include <memory>
+#include <chrono>
+#include <thread>
+#include <cstdio>
+#include <cstring>
+#include <unistd.h>
+#include <sstream>
+#include <termios.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+#include <netdb.h>
+#include "ups.h"
+#include "indicom.h"
+
+// We declare an auto pointer to UPS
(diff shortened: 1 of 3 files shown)


drivers/auxiliary/ups.cpp
```diff
@@ -0,0 +1,577 @@
...
+    // Add UPS parameters as weather parameters
+    addParameter("BATTERY_CHARGE", "Battery Charge", 10, 100, 0);
+    addParameter("BATTERY_VOLTAGE", "Battery Voltage", 12, 14, 0);
+    addParameter("INPUT_VOLTAGE", "Input Voltage", 210, 240, 0);
+
+    // Set critical parameters
+    setCriticalParameter("BATTERY_CHARGE");
+
+    // Load config before setting any defaults
+    loadConfig(true);
```
knro, line 90: No need for this.

drivers/auxiliary/ups.cpp
```diff
@@ -0,0 +1,577 @@
...
+    // Setup UPS name property
+    UPSNameTP[0].fill("NAME", "UPS Name", "ups");  // Default UPS name
+    UPSNameTP.fill(getDeviceName(), "UPS_NAME", "UPS", CONNECTION_TAB, IP_RW, 60, IPS_IDLE);
+
+    // Setup connection settings properties
+    ConnectionSettingsNP[0].fill("TIMEOUT", "Timeout (sec)", "%.0f", 1, 30, 1, 5);
+    ConnectionSettingsNP[1].fill("RETRIES", "Max Retries", "%.0f", 1, 10, 1, 3);
+    ConnectionSettingsNP[2].fill("RETRY_DELAY", "Retry Delay (ms)", "%.0f", 100, 5000, 100, 1000);
+    ConnectionSettingsNP[3].fill("RECONNECT_ATTEMPTS", "Max Reconnect Attempts", "%.0f", 0, 10, 1, 3);
+    ConnectionSettingsNP.fill(getDeviceName(), "CONNECTION_SETTINGS", "Connection", CONNECTION_TAB, IP_RW, 60, IPS_IDLE);
```
knro, line 75: You can add this as well to load property from config.

`ConnectionSettingsNP.load();`

drivers/auxiliary/ups.cpp
```diff
@@ -0,0 +1,577 @@
...
+    // Set timeout for connect
+    struct timeval tv;
+    tv.tv_sec = static_cast<int>(ConnectionSettingsNP[0].getValue());
+    tv.tv_usec = 0;
+    setsockopt(socketFd, SOL_SOCKET, SO_RCVTIMEO, &tv, sizeof(tv));
+    setsockopt(socketFd, SOL_SOCKET, SO_SNDTIMEO, &tv, sizeof(tv));
+
+    // Enable keep-alive
+    int keepalive = 1;
+    setsockopt(socketFd, SOL_SOCKET, SO_KEEPALIVE, &keepalive, sizeof(keepalive));
```
knro, line 156: is there a reason for using your own implementation instead of using INDI's own connectionTCP?

drivers/auxiliary/ups.cpp
grm: I have done it in the loadConfig funciton.
It is not the right place ?

drivers/auxiliary/ups.cpp
grm: Indeed !

drivers/auxiliary/ups.cpp
grm: Indeed, no specific reason, i migrated to indi TCP connection

knro:
Please fix the CI issues (unused parameters)

TallFurryMan:
@knro is the Fedora issue regarding ciso64 a global issue that needs investigation?

PR #2241, opened 2025-05-28
Draft: Telescope Simulator, AltAz mount  and derotation option

escribana:
- add ALTAZ mount (experimental)
- implement tracking rates of the mechanical axes for ALTAZ

### DIFF
--- a/drivers/telescope/scopesim_helper.cpp
+++ b/drivers/telescope/scopesim_helper.cpp
@@ -115,6 +115,11 @@ Axis::AXIS_TRACK_RATE Axis::TrackRate()
     return trackingRate;
 }
 
+double Axis::getTrackingRateDegSec()
+{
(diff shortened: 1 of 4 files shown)


knro:
Thanks! Would this effect how the current HA/DE axis simulation work?

escribana:
> Thanks! Would this effect how the current HA/DE axis simulation work?

No, the ALTAZ track simulation works independently and is currently triggered by view switch (star globe view -> EQ / horizontal view -> ALTAZ)
BTW: I'm currently working on ALTAZ sim FOV simulation, which is more complicated than I thought. This will be implemented in skymap. (see KStars/EKOS part of derotation option)

knro:
@escribana Thank you, does this PR alter the current behavior of the drivers if the rotation was zero?

PR #2259, opened 2025-07-20
fix issue of Celestron SCT focuser for aarch64

wgauvin:
This is a continuation of work done in commit 83030d454 (see https://github.com/indilib/indi/pull/2252) which fixed issue on an x86_64 but the issue stil existed on the aarch64.

I've found 2 issues that were causing issues on the aarch64
* the `toHexStr` could cause a buffer overflow in the sprintf.  Not sure why even the x86_64 architecture didn't pick it up
* the default baud rate was set to 19200 but doing `stty -a` on the serial dev said it was 9600. On my x86_64 server it would happily handle 19200 but on my ODROID N2+ it would crash the how xhci and I would have to completely power cycle my ODROID.

I've changed the `toHexStr` to avoid the buffer overflow and I've set the default baud rate to 9600.

If there is a clean way to warn users know about the baud rate, then I'm happy to make the change but would need to know where to make the changes.

I've been able to test this fix on both x86_64 and aarch64 machines so I'm confident that this should be the last issue of the Celestron SCT focuser for now.

### DIFF
--- a/drivers/focuser/celestron.cpp
+++ b/drivers/focuser/celestron.cpp
@@ -114,8 +114,10 @@ bool CelestronSCT::initProperties()
     // Add debugging support
     addDebugControl();
 
-    // Set default baud rate to 19200
-    serialConnection->setDefaultBaudRate(Connection::Serial::B_19200);
+    // Set default baud rate to 9600
+    // On aarch64 19200 or more seems to crash the whole USB hub.
+    // stty -a says the baud rate is 9600
+    serialConnection->setDefaultBaudRate(Connection::Serial::B_9600);
 
     communicator.setDeviceName(getDeviceName());
 
--- a/drivers/focuser/celestronauxpacket.cpp
+++ b/drivers/focuser/celestronauxpacket.cpp
@@ -43,7 +43,7 @@ char * toHexStr(buffer data)
         sz = 100;
     for (int i = 0; i < sz; i++)
     {
-        snprintf(&debugStr[i * 3], 301, "%02X ", data[i]);
+        snprintf(&debugStr[i * 3], 301 - 3*i, "%02X ", data[i]);
     }
     return debugStr;
 }


wgauvin:
Yeah I did, and just did it again to confirm.  I was able to connect at 9600 and perform a changing of the focus position.

The following is what I get when I run `stty -a < /dev/ttyACM0`  on my x86_64 server. 

```
speed 9600 baud; rows 0; columns 0; line = 0;
intr = ^C; quit = ^\; erase = ^?; kill = ^U; eof = ^D; eol = <undef>; eol2 = <undef>; swtch = <undef>; start = ^Q; stop = ^S; susp = ^Z; rprnt = ^R;
werase = ^W; lnext = ^V; discard = ^O; min = 1; time = 0;
-parenb -parodd -cmspar cs8 -hupcl -cstopb cread clocal -crtscts
ignbrk -brkint ignpar -parmrk inpck -istrip -inlcr -igncr -icrnl -ixon -ixoff -iuclc -ixany -imaxbel -iutf8
-opost -olcuc -ocrnl -onlcr -onocr -onlret -ofill -ofdel nl0 cr0 tab0 bs0 vt0 ff0
-isig -icanon -iexten -echo -echoe echok -echonl noflsh -xcase -tostop -echoprt echoctl echoke -flusho -extproc
```

From this the speed is 9600 baud and is 8N1 ( 8bit (from cs8), no parity (-parenb), and 1 stop bit (-cstopb)).

knro:
Ok thank you! Not sure it was set to a higher baud rate before. I thought it was in Celestron's documentation, apparently not.

wgauvin:
Thanks for getting this merged before the next version.  I know 3 yrs ago it worked and about a year ago it stopped but I wasn't very active with astro and thought it was my focuser had died.  I only just got back into astro and by chance saw a message about others having an issue.  Glad to help out getting it sorted, even though I doubt i would use my SCT for any decent imaging of late.

My gut feeling we're up against some changes (or combination thereof), gcc version and/or kernel causing the higher baud rate not working on the aarch.  but anyway there was definitely buffer overruns happening.

PR #2267, opened 2025-08-21
Driver for Wanderer Cover V4 Pro-EC added and known issue for V4-EC driver fixed.

WandererAstro:


WandererAstro:
> Thanks! Please note that #2266 was merged yesterday. Is this change compatible with it?

Yes, it is compatible. It is modified on #2266 after testing.

PR #2270, opened 2025-08-25
Added check for lightbox capabilities in LI::updateProperties()

murjo06:
Added a check whether the lightbox has the CAN_DIM capability set to 1, since before all lightboxes (even those that couldn't dim) had the Brightness property in the driver.

### DIFF
--- a/libs/indibase/indilightboxinterface.cpp
+++ b/libs/indibase/indilightboxinterface.cpp
@@ -89,7 +89,7 @@ bool LightBoxInterface::updateProperties()
     if (m_DefaultDevice->isConnected())
     {
         m_DefaultDevice->defineProperty(LightSP);
-        if (m_Capabilities & 0b1)
+        if (m_Capabilities & CAN_DIM)
             m_DefaultDevice->defineProperty(LightIntensityNP);
         if (!FilterIntensityNP.isEmpty())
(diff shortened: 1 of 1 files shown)


libs/indibase/indilightboxinterface.cpp
```diff
@@ -89,14 +89,16 @@ bool LightBoxInterface::updateProperties()
     if (m_DefaultDevice->isConnected())
     {
         m_DefaultDevice->defineProperty(LightSP);
-        m_DefaultDevice->defineProperty(LightIntensityNP);
+        if (m_Capabilities & 0b1)
```
knro, line 92: ```if (m_Capabilities & CAN_DIM)```

libs/indibase/indilightboxinterface.cpp
```diff
@@ -89,14 +89,16 @@ bool LightBoxInterface::updateProperties()
...
+            m_DefaultDevice->defineProperty(LightIntensityNP);
         if (!FilterIntensityNP.isEmpty())
             m_DefaultDevice->defineProperty(FilterIntensityNP);
     }
     else
     {
         m_DefaultDevice->deleteProperty(LightSP);
-        m_DefaultDevice->deleteProperty(LightIntensityNP);
+        if (m_Capabilities & 0b1)
```
knro, line 100: Same as above

PR #2272, opened 2025-08-28
Allow using hidapi from system

mattiaverga:


mattiaverga:
mmm I don't understand the failure, I've just renamed the headers to `indi_hidapi.h`. I don't know much about C. Can you give me an idea where to look?

BTW, if this PR gets approved, indi-3rdparty will need some adjustment, like for the system's jsonlib change.

PR #2287, opened 2025-10-20
Power refactor

knro:
Migrate all drivers using power or dew controller under the unified INDI::Power interface. This is a major refactor and all migrated drivers need to be tested.


knro:
Thank you, so for MyDCPCESP32, the number of available dew channels is only know after ```getActiveChannels``` is called? so the initProperties should be delayed until then?

I'm going to need more help to configure this using INDI::Power. Right now, it's a simple Power/Dew/Variable channels division and control for each.

hcomet:
For MyDCP4ESP32, the number of available channels which support automated dew control is determined by `getActiveChannels`. It checks to see if a channel has a temperature probe attached. For Channels 1, 2 and 4 they need a probe to be active. For channel 3 it only needs a probe for automated dew control. Channel 3 can be set to manual or be attached to another active channel even it it has no probe attached. `getActiveChannels` will tell you which channels to show in the Dew tab next to *Auto Dew Control*.

I can look at the changes required for MyDCPESP32 and CheapoDC to move to using INDI:Power. Should I fork from **knro/indi:power_refactor**?

Is there anything that indicates the intent of how the new Tabs should work? ie: *Auto Dew Control* enables/disables auto for a channel. In CheapoDC this would move the channel from auto to manual. In MyDHCP4ESP32 only Channel 3 has manual so it would disable the other channels if unchecked? Should it also zeror the output of a channel when set to manual?

It seems that the *Channels* line may also enable/disable a channel? The *Duty Cycles* section should show current channel output whether auto or manual? If auto, then changes to output should be rejected?

Both the controllers will reject commands that do not fit the current settings for channels so the next `readSettings` will set everything back to what the controller has.

One more question, I see that you added all the controller settings to `saveConfigItems`. In past I had only included INDI specific items in `saveConfigItems` as well as anything that the controller did not maintain in its own config. Is there a preference for how this should be done?

knro:
@hcomet Please submit a PR now to master directly.

PR #2293, opened 2025-10-30
Fix bug with driver not registering correctly

jrhuerta:
The driver is failing to register as Dust Cap and Light Box


jrhuerta:
I've been working with Gemini to give support to these flat panels and I recently did a refactor to accommodate for differences on firmware updates and I thought setting the interface on Handshake was enough. I was still using the prior version on my setup but when I updated today noticed the issue. Sorry about that.

PR #2315, opened 2025-12-23
Bug fixes for Pegasus devices

sterne-jaeger:
Introducing the INDI power device interface caused some regressions, which are fixed herewith for the PowerBox Advanced (PPBA) and Micro (same code) as well as the Saddle Power Box:
- Additional power statistics available  added on the Power tab
- Adjustable power output control implemented separately on the PPBA (the general approach for multiple channels adds unnecessary complexity)
- Default labels for power and dew channels as named on the device
- Auto dew turns all dew channels on (Pegasus behavior)
- Bug fix for disabling single dew channels (turning it off made it impossible to turn it on again)


knro:
Can you please check the  conflicts?

PR #2318, opened 2025-12-31
Prevending Redefinition of Header Files

rlancaste:
I found some cases the other day where building software based on libindi caused redefinition issues with the INDIAPI and INDIDEVAPI header files.  "#Pragma Once" should usually take care of this issue, but since these header files include defined structs, it seems it is not always enough.  This change seemed to fix the error.

### DIFF
--- a/libs/indicore/indiapi.h
+++ b/libs/indicore/indiapi.h
@@ -1,3 +1,6 @@
+#ifndef _INDIAPI_H
+#define _INDIAPI_H
+
 #if 0
 INDI
 Copyright (C) 2003 Elwood C. Downey
@@ -510,4 +513,6 @@ extern "C" {
 // FIXME: duplicated from indidevapi.h. Can we share ?
 #ifdef __cplusplus
 }
+#endif
+
 #endif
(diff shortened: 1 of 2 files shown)


rlancaste:
Here is an example of the issue.  Note that this error has not happened before last week and I have run this same script hundreds of times in the past.  But I can see how the issue could occur and it's possible that something changed on my system and/or compiler.  In these 2 header files, the structs are defined.  "#pragma once" according to what I found on multiple sources online is only supposed to be used with header files that don't define things, for header files that define structs etc, the other way of doing which this MR changes to is better in those conditions.  I could be wrong about this, but it did fix the issue when I tried it.

I do agree that we probably should not mix the two.  I could modify this MR to remove the #pragma once from these two files in addition to adding the ifdefs if you prefer.

In file included from /src/kstars/kstars/indi/indipropertyexporter.cpp:9:
In file included from /src/kstars/kstars/indi/indipropertyexporter.h:20:
In file included from /AstroRoot/CraftRoot/include/libindi/indidevapi.h:99:
AstroRoot/CraftRoot/include/libindi/indiapi.h:152:1: error: redefinition of enumerator 'ISS_OFF'
  152 | ISS_OFF = 0, /*!< Switch is OFF */
      | ^
AstroRoot/DEV_CRAFT/include/libindi/indiapi.h:149:1: note: previous definition is here
  149 | ISS_OFF = 0, /*!< Switch is OFF */
      | ^

PR #2323, opened 2026-01-20
RST135E: Set TrackState = SCOPE_SLEWING, when executing findHome(), s…

karasyyz:
RST135E: Set TrackState = SCOPE_SLEWING, when executing findHome(), so kstars HOME status would be updated correctly.

Otherwise, this mount is always stuck at the yellow indicator.


knro:
Thank you!

PR #2336, opened 2026-02-14
Custom guide rate for LX200GPS driver

jordisese:
Add property to send Guide Rate to PHD2, etc. (Custom Guide Rate in MOTION_TAB)
Only sets it in the indi driver and maintains it if options are saved. It does not communicate with the mount, only with indi clients. 
Thus, it must be set both in ekos and the mount controller to work as intended.


jordisese:
> So how would these beneficial exactly? to learn about guide rate in PHD2?

Yes, exactly. I added it so PHD2 would know which guide rate I was using. I was told that corrections are calculated using that value. I read that some mounts (zwo am5, etc.) were reporting problems with guiding and they improved results with PHD2 after adding this to their ascom/indi drivers; so I wanted to test it. Since it worked fine for me and it was a minor modification, I thought it might be useful for others. 

Anyway, guiding works without it.

PR #2339, opened 2026-02-20
Improves support for the Pegasus PowerBox Micro (PPBM) when used with the `indi_pegasus_ppba` driver.

petercoxphoto:
__Changes__

- `Handshake()` now extracts the device model from the handshake response (`PPBA_OK` → `PPBA`, `PPBM_OK` → `PPBM`). The detected type is shown in a new "Device Type" row in the Firmware tab, and in the connection log message.
- The `POWER_CHANNELS` single-toggle checkbox from `PI::PowerInterface` is replaced by a two-button __On / Off__ `ISR_1OFMANY` switch (`QUAD_OUTPUT`) in the Power tab — consistent with the LED control style and clearer to use.
- An identical __Quad Output On/Off__ control and a live __Power (Voltage / Current / Power)__ readout are added to the __Main Control__ tab, so the most-used controls are immediately visible without tab-switching.
- Both On/Off controls stay in sync with each other and with the internal `PI::PowerChannelsSP` state.

__Tested__ against PPBM firmware 2.7 connected via USB serial.

### DIFF
--- a/drivers/power/pegasus_ppba.cpp
+++ b/drivers/power/pegasus_ppba.cpp
@@ -147,8 +147,32 @@ bool PegasusPPBA::initProperties()
     ////////////////////////////////////////////////////////////////////////////
     FirmwareTP[FIRMWARE_VERSION].fill("VERSION", "Version", "NA");
     FirmwareTP[FIRMWARE_UPTIME].fill("UPTIME", "Uptime (h)", "NA");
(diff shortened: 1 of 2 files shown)


knro:
Thank you! Please do not delete any INDI::PowerInterface properties since these are supposed to form standard behavior for all Power devices. I suppose it *might* be possible to add another alias property but that would probably confuse the user more.

PR #2352, opened 2026-03-29
alignment: add plugin test suite with Halton sky sampling

ckemper67:
## Summary

This branch establishes a test harness for INDI alignment math plugins and fixes two correctness bugs in the simulator and BuiltIn plugin that the tests exposed.

### Interface changes

- **`MathPlugin::TransformTelescopeToCelestial`** gains a `JulianOffset = 0` parameter, matching the existing `TransformCelestialToTelescope` signature. This allows tests to inject a fixed Julian Date and get deterministic transforms independent of the system clock. All existing implementations updated; a TODO notes that a future refactor should move to an absolute Julian Date rather than an offset.
- **`InMemoryDatabase::SetDatabaseReferencePosition`** gains an `IGeographicCoordinates` overload alongside the existing `(double lat, double lon)` form.
- **`indicom.h`** gains angle-unit conversion macros: `HOURS_TO_DEG`, `DEG_TO_HOURS`, `ARCMIN_TO_DEG`, `DEG_TO_ARCMIN`, `ARCSEC_TO_DEG`, `DEG_TO_ARCSEC`, `HOURS_TO_RAD`, `RAD_TO_HOURS` (all guarded with `#ifndef`).
- **`BuiltInMathPlugin`** gains a `NO_PLUGIN_HOOKS` guard around its `extern "C"` factory functions, matching the pattern used by SVD and Nearest, so the translation unit can be linked directly into test binaries without symbol conflicts.

### Bug fixes

**`scopesim_helper.cpp` — AltAz azimuth convention**
`apparentHaDecToMount` was producing azimuth in scopesim's internal convention (0° = South, clockwise) without converting to INDI's `IHorizontalCoordinates` convention (0° = North). Added `+Angle(180.0)` to the primary output and a `return` to prevent fallthrough into the equatorial code path.

**`scopesim_helper.cpp` — GEM meridian flip**
`mountToApparentHaDec` initialised `prio` inside an `else` branch, leaving it uninitialised when the flip condition was true. Moved `prio = primary` before the flip check so both branches are always covered.

**`libs/alignment/BuiltInMathPlugin.cpp` — singular matrix fallthrough**
`CalculateTransformMatrices` wrapped the `MatrixMatrixMultiply` and inverse steps in an `else` guarded by the singularity check, so a singular Alpha matrix silently skipped the Beta matrix calculation entirely. Removed the `else`; the identity fallback is set and computation continues regardless.

### Test suite (`test/alignment/test_alignment_plugins.cpp`)

34 tests using a dual-[Halton sequence](https://en.wikipedia.org/wiki/Halton_sequence) design: alignment points drawn from `Halton(2,3)`, independent validation points from `Halton(5,7)`. This eliminates the correlation between training and test sets while giving good sky coverage with few points.

Coverage includes:
- **SVDMathPlugin**: polar errors, axis errors (IH/ID/CH), non-perpendicularity, small/large errors, all-terms combined, multiple geographic sites (LA, Tokyo, Sydney, Nairobi), AltAz, EQ_FORK, GEM meridian flip (north and south), convex hull fallback
- **BuiltInMathPlugin**: all-terms equatorial, AltAz, meridian flip
- **NearestMathPlugin**: round-trip (all-in-alignment), coarse quality check, AltAz

Assertions use a fixed `targetRMSArcsec` quality target rather than a fraction of injected error, with a P95 bound at `1.5×` the RMS target to catch outliers.

## Test plan

```bash
cd build/test/alignment
cmake --build . --target test_alignment_plugins
./test_alignment_plugins
```

Expected: 34 tests pass. The two `DISABLED_Nearest_MeridianFlip` tests are intentionally skipped (NearestMathPlugin does not handle flip geometry reliably without dense coverage).

### DIFF
--- a/drivers/telescope/scopesim_helper.cpp
+++ b/drivers/telescope/scopesim_helper.cpp
@@ -302,8 +302,14 @@ void Alignment::apparentHaDecToMount(Angle apparentHa, Angle apparentDec, Angle*
         Vector altAzm = Vector(apparentHa, apparentDec).rotateY(Angle(90) - latitude);
         // azimuth in scopesim convention is 0° = South, increasing clockwise West→North→East
         // INDI's IHorizontalCoordinates uses 0° = North.
-        *primary = altAzm.primary() + Angle(180.0);
-        *secondary = altAzm.secondary();
+        Angle apparentAz = altAzm.primary() + Angle(180.0);
+        Angle apparentAlt = altAzm.secondary();
+
+        Angle instrumentAz, instrumentAlt;
+        observedToInstrument(apparentAz, apparentAlt, &instrumentAz, &instrumentAlt);
+
+        *primary = instrumentAz;
+        *secondary = instrumentAlt;
         LOGF_EXTRA1("apparent HaDec to ALTAZ: ha %f, dec %f  to pri %f, sec %f", apparentHa.Degrees(), apparentDec.Degrees(), primary->Degrees(),
                     secondary->Degrees() );
         return; // Prevent fallthrough to equatorial override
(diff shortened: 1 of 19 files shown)


knro:
Thank you for the changes. Are you using the Alignment module for any specific requirement when you ran into these issues?

PR #2365, opened 2026-04-12
Fix HotPlugManager destroying connected devices during transient USB resets

cfuture81:
Some USB cameras (notably ZWO ASI) perform a USB bus reset after each frame readout as part of normal operation. The kernel reports this as a device disconnect/reconnect cycle (usb X-X.X.X: reset high-speed USB device), which triggers a udev event.

When the HotPlugManager's checkHotPlugEvents() runs during this brief window, discoverConnectedDeviceIdentifiers() (which calls ASIGetNumOfConnectedCameras()) does not see the camera on the bus. The manager treats this as a genuine disconnection and calls destroyDevice(), which:

1) Calls Disconnect() on the device
2) Calls deleteProperty(nullptr), removing all INDI properties from the server
3) Removes the device from the managed list

On the next poll cycle (typically 1 second later), the camera reappears and createDevice() builds a fresh device instance with new property definitions. However, any INDI clients that were connected (KStars/Ekos, PHD2, etc.) still hold references to the old properties. Every subsequent property update from the driver hits "Could not find property" errors, effectively breaking all client communication with the camera.

This is particularly problematic when PHD2 is used as an external guider alongside Ekos, as PHD2 maintains its own INDI client connection. The property cache invalidation causes PHD2 to fail to complete exposures, leading to camera disconnection timeouts and guiding failure.

The fix adds a check in checkHotPlugEvents(): if a managed device is not found on USB but is currently connected to a client (isConnected() returns true), skip the removal. The device will be cleaned up naturally when the INDI connection fails on the next I/O operation (in the case of a true physical disconnection), or will continue operating normally (in the case of a transient USB reset).

This fix applies to all drivers using the HotPlugManager (ASI, SVBony, PlayerOne, etc.).

### DIFF
--- a/drivers/telescope/lx200_OnStep.cpp
+++ b/drivers/telescope/lx200_OnStep.cpp
@@ -4991,6 +4991,9 @@ bool LX200_OnStep::Sync(double ra, double dec)
     currentRA  = ra;
     currentDEC = dec;
 
+    EqNP.setState(IPS_OK);
+    NewRaDec(currentRA, currentDEC);
+
     LOG_INFO("OnStep: Synchronization successful.");
     return true;
 }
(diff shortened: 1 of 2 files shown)


knro:
I am unable to reproduce this issue locally.

PR #2366, opened 2026-04-13
telescope simulator: add INDI alignment integration, and robust Alt-Az support 

ckemper67:
This turned out to be a lot more involved than I had anticipated, but it helped me uncover a couple of bugs in the new math plugin I am working on. I tried to break it up into individual commits to make it easier to review, but let me know if you want me to break it into multiple PRs

## Summary

This PR extends the telescope simulator to serve as a testbed for evaluating INDI alignment math plugins against physically realistic mount errors. I can now run a complete model building for an Alt-Az telescope after injecting errors.

The simulator already injects configurable pointing errors based on Patrick Wallace's classical 6-term mount model. This PR integrates the INDI alignment subsystem and brings Alt-Az support to feature parity with equatorial modes, so the full goto/track/sync loop can be exercised end-to-end.

### INDI alignment integration

The INDI alignment subsystem is now integrated for both equatorial and Alt-Az mount types. Syncing on stars feeds the math plugin with the instrument-to-sky discrepancies produced by the injected Wallace errors. A correct alignment model should converge goto and tracking accuracy back toward the noise floor, making it possible to evaluate plugin quality quantitatively using the CCD simulator's star fields.

### Arcminute input fields

The existing degree-valued mount model properties are hard to use for realistic error magnitudes (typical mount errors are a few arcminutes, not degrees). This PR adds companion arcminute-valued input fields for all 6 Wallace terms, with tightened limits: index errors (IH/ID) allow up to +/-180 deg to accommodate a misaligned Az zero point; all other terms continue to be limited to +/-5 deg.

### Robust Alt-Az support

Prior to this PR the Alt-Az mount type was only partially implemented. This PR brings it to feature parity with the equatorial modes:

- **Coordinate system correctness**: axisPrimary now consistently uses the INDI Az convention (North=0, East=90) throughout goto, sync, tracking, guiding, and park. The previous mixed convention caused systematic pointing errors in Az.
- **Goto and sync**: Alt-Az goto correctly converts the celestial target through `EquatorialToHorizontal`, routes through the alignment subsystem when sync points exist, and sets axis slew targets directly in Az/Alt space.
- **Tracking servo**: a parabolic 3-point window replaces the previous Patera formula for axis rate prediction. The tracking target is the alignment-corrected instrument position, so the servo tracks the sky position the alignment model has learned rather than the raw encoder position. Solar and lunar modes apply the correct RA drift rate at each sample point, consistent with EQ behaviour.
- **Guiding**: N/S/E/W guide pulses are decomposed into Az/Alt axis motion via the parallactic angle.
- **Park defaults**: Alt-Az park defaults to Az=0/Alt=0 (horizon north) instead of the equatorial default.
- **Mount type notification**: the alignment subsystem is informed of the mount type at connect time so math plugins use the correct coordinate encoding.

## Changes

- **Arcminute input fields** for all 6 Wallace model terms, with tightened limits; mirrored as degrees to `MOUNT_MODEL` for snooping by external tools
- **INDI alignment subsystem** integrated: sync points are recorded, goto corrects for learned errors, and the mount type (EQ/Alt-Az) is communicated so plugins use the right coordinate model
- **Alt-Az tracking servo**: uses alignment-corrected RA/Dec as the tracking target and a parabolic 2nd-order polynomial fit to predict axis rates; solar/lunar modes apply RA drift offset
- **Coordinate and pier-side fixes**: axisPrimary Az convention unified to INDI North=0/East=90, meridian flip detection corrected, ALTAZ park defaults corrected
- **CCD and guide simulator PE snoop fixes**: both simulators had a logic inversion (`if (!usePE)`) that meant the EQUATORIAL_PE branch was never taken — star fields were always generated from the raw encoder position rather than the true (error-injected) position. Fixed in both. The guide simulator had additional issues: `raPE`/`decPE` were missing from the header entirely (making the PE blocks permanently dead code), `USE_EQUATORIAL_PE` was not set in CMakeLists, and the snoop handler used the old `INumberVectorProperty` API. All three are corrected; both simulators now use `INDI::ObservedToJ2000` and the `PropertyNumber` API for the PE snoop path.

### DIFF
--- a/drivers/telescope/telescope_simulator.cpp
+++ b/drivers/telescope/telescope_simulator.cpp
@@ -353,9 +353,21 @@ bool ScopeSim::ReadScopeStatus()
         double JDnow   = ln_get_julian_from_sys();
         double JDoffset = dt / 86400.0;
 
+        // RA drift rate relative to sidereal for the selected track mode, in hours/sec.
+        // Nudging RA at each parabola sample by the non-sidereal offset makes the
+        // derived Az/Alt rates correct for the track mode without changing the meaning
+        // of m_targetRA/Dec.  Mirrors EQ behaviour: RA is corrected, Dec is not.
+        int trackMode = TrackModeSP.findOnSwitchIndex();
+        double raDriftHrsPerSec = 0;
+        if (trackMode == TRACK_SOLAR)
+            raDriftHrsPerSec = (TRACKRATE_SIDEREAL - TRACKRATE_SOLAR) / 3600.0 / 15.0;
+        else if (trackMode == TRACK_LUNAR)
+            raDriftHrsPerSec = (TRACKRATE_SIDEREAL - TRACKRATE_LUNAR) / 3600.0 / 15.0;
+
         auto getAltAz = [&](double JD, INDI::IHorizontalCoordinates &coords)
         {
-            INDI::IEquatorialCoordinates eq { m_targetRA, m_targetDEC };
+            double dtFromNow = (JD - JDnow) * 86400.0;
+            INDI::IEquatorialCoordinates eq { m_targetRA + raDriftHrsPerSec * dtFromNow, m_targetDEC };
             INDI::EquatorialToHorizontal(&eq, &m_Location, JD, &coords);
         };
 
--- a/drivers/telescope/telescope_simulator.cpp
+++ b/drivers/telescope/telescope_simulator.cpp
@@ -82,23 +82,14 @@ bool ScopeSim::initProperties()
     simPierSideSP.fill(getDeviceName(), "SIM_PIER_SIDE", "Sim Pier Side",
                        "Simulation", IP_WO, ISR_1OFMANY, 60, IPS_IDLE);
 
-    mountModelArcminNP[MM_IH].fill("MM_IH", "Ha Zero (IH) '", "%.1f", -10800, 10800, 0.1, 0);
-    mountModelArcminNP[MM_ID].fill("MM_ID", "Dec Zero (ID) '", "%.1f", -10800, 10800, 0.1, 0);
-    mountModelArcminNP[MM_CH].fill("MM_CH", "Cone (CH) '", "%.1f", -300, 300, 0.1, 0);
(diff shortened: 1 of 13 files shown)


drivers/telescope/telescope_simulator.cpp
```diff
@@ -94,14 +82,23 @@ bool ScopeSim::initProperties()
...
+                            "Simulation", IP_RW, 0, IPS_IDLE);
+
+    mountModelNP[MM_IH].fill("MM_IH", "Ha Zero (IH) deg", "%g", -180, 180, 0.01, 0);
+    mountModelNP[MM_ID].fill("MM_ID", "Dec Zero (ID) deg", "%g", -180, 180, 0.01, 0);
+    mountModelNP[MM_CH].fill("MM_CH", "Cone (CH) deg", "%g", -5, 5, 0.01, 0);
+    mountModelNP[MM_NP].fill("MM_NP", "Ha/Dec (NP) deg", "%g", -5, 5, 0.01, 0);
+    mountModelNP[MM_MA].fill("MM_MA", "Pole Azm (MA) deg", "%g", -5, 5, 0.01, 0);
+    mountModelNP[MM_ME].fill("MM_ME", "Pole elev (ME) deg", "%g", -5, 5, 0.01, 0);
+    mountModelNP.fill(getDeviceName(), "MOUNT_MODEL", "Mount Model (deg)",
+                      "Simulation", IP_RO, 0, IPS_IDLE);
```
knro: Is there a practical reason to keep the degrees? If it is not used elsewhere, then we can switch to the arcminute fields only to avoid any confusion.

ckemper67: I just kept them for backwards compatibility, but let me check if any other driver is using them.

ckemper67: I checked and no driver in indi or indi-3rdparty was using the degree properties. So I decided to drop the Arcmin property block completely and reuse the mountModelNP properties, but scale all of them to arcminutes.

PR #2370, opened 2026-04-17
StreamManager: add Full Depth toggle for native bit-depth streaming

cfuture81:
The INDI stream manager deliberately downscales 16-bit camera frames to 8-bit
in asyncStreamThread() to reduce streaming bandwidth. While this is a sensible
default for live preview, it discards valuable dynamic range for applications
like autoguiding where the stream frames are used for star detection and
centroid calculation.

This patch adds a "Full Depth" ON/OFF switch to the Streaming tab in the INDI
control panel, placed directly below the existing Video Stream toggle. When
enabled, the 8-bit downscale is bypassed and native 16-bit frames are sent to
clients.

Changes:
- streammanager_p.h: new FullDepthSP property with FULL_DEPTH_ON/OFF enum
- streammanager.cpp: property init, registration in ISGetProperties/
  updateProperties, switch handler in ISNewSwitch, conditional downscale
  skip in asyncStreamThread

The switch defaults to Off, preserving existing behavior. No changes are
required in camera drivers — the ASI driver (and others) already configure
the streamer with the correct pixel depth via setPixelFormat().

Tested with ZWO ASI220MM Mini (RAW16 mode) on StellarMate (aarch64).
8-bit and 16-bit stream frames verified in both KStars Ekos guide module
and PHD2.

### DIFF
--- a/libs/indibase/stream/streammanager.cpp
+++ b/libs/indibase/stream/streammanager.cpp
@@ -105,6 +105,11 @@ bool StreamManagerPrivate::initProperties()
         StreamSP.fill(getDeviceName(), "CCD_VIDEO_STREAM", "Video Stream",
                       STREAM_TAB, IP_RW, ISR_1OFMANY, 0, IPS_IDLE);
 
+    // Full Depth Streaming — right after Video Stream toggle
+    FullDepthSP[FULL_DEPTH_8BIT ].fill("FULL_DEPTH_8BIT",  "8-bit",  ISS_ON);
+    FullDepthSP[FULL_DEPTH_16BIT].fill("FULL_DEPTH_16BIT", "16-bit", ISS_OFF);
+    FullDepthSP.fill(getDeviceName(), "STREAM_FULL_DEPTH", "Stream Depth", STREAM_TAB, IP_RW, ISR_1OFMANY, 0, IPS_IDLE);
+
     // @INDI_STANDARD_PROPERTY@
     StreamTimeNP[0].fill("STREAM_DELAY_TIME", "Delay (s)", "%.3f", 0, 60, 0.001, 0);
     StreamTimeNP.fill(getDeviceName(), "STREAM_DELAY", "Video Stream Delay", STREAM_TAB, IP_RO, 0, IPS_IDLE);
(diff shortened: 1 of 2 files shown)


knro:
Shouldn't be an 8 vs 16bit toggle instead of "full depth"? that is more clear. It should default to 8 bit.

cfuture81:
@knro   pull request updated.  
image

cfuture81:
also this option for streaming bit rate can now be saved into the device settings xml of the specific camera.  (this was the last commit)

libs/indibase/stream/streammanager.cpp
```diff
@@ -910,6 +920,16 @@ bool StreamManagerPrivate::ISNewSwitch(const char * dev, const char * name, ISSt
         return true;
     }
 
+    // Full Depth Streaming
+    if (FullDepthSP.isNameMatch(name))
+    {
+        FullDepthSP.update(states, names, n);
+        FullDepthSP.setState(IPS_OK);
+        FullDepthSP.apply();
+        LOGF_INFO("Stream depth set to %s.", FullDepthSP[FULL_DEPTH_16BIT].getState() == ISS_ON ? "16-bit" : "8-bit");
```
knro: Better use this method:

```
if (FullDepthSP.isNameMatch(name))
{
            updateProperty(FullDepthSP, states, names, n, [this, states]()
            {
                LOGF_INFO("Stream depth set to %s.", states[FULL_DEPTH_16BIT].getState() == ISS_ON ? "16-bit" : "8-bit");
                return true;
            }, true);
            return true;
}
```

cfuture81: thanks  @knro   - I updated the code

cfuture81:
Updated — now using updateProperty() with callback and saveConfig=true as suggested. Also fixed the inverted log message. Thanks for the review!

libs/indibase/stream/streammanager.cpp
```diff
@@ -910,6 +920,17 @@ bool StreamManagerPrivate::ISNewSwitch(const char * dev, const char * name, ISSt
...
 
+    // Full Depth Streaming
+    if (FullDepthSP.isNameMatch(name))
+    {
+        currentDevice->updateProperty(FullDepthSP, states, names, n, [this]()
+        {
+            LOGF_INFO("Stream depth set to %s.", FullDepthSP.findOnSwitchIndex() == FULL_DEPTH_8BIT ? "16-bit" : "8-bit");
+            return true;
+        }, true);
+        return true;
```
knro: You have to pass states to the lamda like my example because FullDepthSP is *not* updated yet.

cfuture81: thank you!  code adopted.

PR #2374, opened 2026-04-22
Support for Sesto Senso 3

RononDex:
This makes the Sesto Senso 3 work properly on my test system.

Currently, these change would break other Primaluca devices (the isBusy() change should be compatible with Sesto Senso 2, however the other changes are probably not).

 @knro could you check how this could be implemented to be only applied for the Sesto Senso 3? I am no cpp dev and this is my first time going through the libindi code :)

### DIFF
--- a/drivers/focuser/primalucacommandset.cpp
+++ b/drivers/focuser/primalucacommandset.cpp
@@ -310,7 +310,7 @@ bool Focuser::isHallSensorDetected(bool &isDetected)
 *******************************************************************************************************/
 bool Focuser::getAbsolutePosition(uint32_t &position)
 {
-    return m_Communication->get(MOT_1, "ABS_POS_STEPS", position);
+    return m_Communication->get(MOT_1, "ABS_POS_STEP", position);
 }
 
 /******************************************************************************************************
--- a/drivers/focuser/primalucacommandset.cpp
+++ b/drivers/focuser/primalucacommandset.cpp
@@ -256,7 +256,7 @@ Focuser::Focuser(const std::string &name, int port)
 *******************************************************************************************************/
 bool Focuser::goAbsolutePosition(uint32_t position)
 {
-    return m_Communication->command(MOT_1, {{"MOVE_ABS", {{"STEP", position}}}});
+    return m_Communication->command(MOT_1, {{"GOTO", position}});
 }
 
 /******************************************************************************************************
@@ -310,7 +310,7 @@ bool Focuser::isHallSensorDetected(bool &isDetected)
 *******************************************************************************************************/
 bool Focuser::getAbsolutePosition(uint32_t &position)
 {
-    return m_Communication->get(MOT_1, "ABS_POS", position);
+    return m_Communication->get(MOT_1, "ABS_POS_STEPS", position);
 }
 
 /******************************************************************************************************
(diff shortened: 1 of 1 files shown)


knro:
Thanks please use astyle to format it properly as it shows too many changes due to formatting.

knro:
So there are three changes?

```return m_Communication->command(MOT_1, {{"MOVE_ABS", {{"STEP", position}}}});```

Becomes

```return m_Communication->command(MOT_1, {{"GOTO", position}});```

Then the busy part:

```
if (m_Communication->get(MOT_1, "STATUS", status))
    {
        bool busy = (status["BUSY"] == 1);
        std::string mst;
        if (status.contains("MST"))
            status["MST"].get_to(mst);
        // Motor is still moving if BUSY=1, or if MST is not yet "stop".
        // The SS3 can transiently assert BUSY=0 while MST is still
        // "CstSpeed" or "dec" during encoder-feedback transitions.
        bool motorStopped = mst.empty() || (mst == "stop");
        return busy || !motorStopped;
    }
 ```
 
 Then this get part:
 
 ```
 bool Focuser::getAbsolutePosition(uint32_t &position)
{
    return m_Communication->get(MOT_1, "ABS_POS_STEPS", position);
}
```

Did I miss anything else?

RononDex:
Yes those are all the changes. I updated the PR with the styling you requested (I could not find a .editorconfig or something in the CONTRIBUTING.md explaining how to format the code).

The isBusy change is necessary since SS3 reports BUSY=0 but MST=CstSpeed while it is stil moving, causing INDI / EKOS to believe the focuser has reached its position while the SS3 is actually still moving.

To request the ABS position from SS3 I noticed that indigo was using ABS_POS_STEP for SS3. After changing that the absolute position was correctly reported from SS3 (with ABS_POS it seems to report a different unit, maybe arc-seconds or micrometers?)

To request SS3 to go to an absolute STEP position it now requires a GOTO command instead of MOVE_ABS.

Reagarding compatibility:

- The **isBusy** check should be safe with SS2 (I can test that, I have a SS2 laying around). 
- I am not sure how the other focusers react to ABS_POS_STEP though (maybe requires newer Firmwares)
- Same with the **GOTO** command, this should probably only be applied if we are connected to a SS3 and use the "old"  **MOVE_ABS** command for all other focusers to be safe

I saw that there is a variable used in `sestosenso2.cpp` that we can use to check if we are connected to a SS3 or a SS2, but not sure how to move those changes from `primalucacommandset.cpp` to `sestosenso2.cpp`

knro:
Thank you for the changes! I decided to take this opportunity to completely separate the SS3 driver from SS2. Can you please recompile latest GIT and check if there are any issues?

RononDex:
One issue I can see: it should be "ABS_POS_STEP" (and not "ABS_POS_STEPS")
I had this wrong in this PR and just fixed it a few minutes ago.

Can you change that in SestoSenso2::getAbsolutePositionSS3?

RononDex:
Also I am a bit confused, why are those custom methods for SS3 on the SestoSenso2 class? Should they not be on the SestoSenso3 class? Also SestoSenso3.cpp calls m_SestoSenso3->isBusy() instead of the isBusySS3() method

knro:
I removed all the SS2 and SS3 customs and also the ABS_POS_STEP is already fixed in the code. Please check.

RononDex:
It seems to be working really smooth :)

I moved SS3 focuser around quite a bit, did a re-calibration and moved around some more. So far 0 issues and it stops exactly at the steps I requested it to.

Thanks for your work :) I can now get back to imaging

knro:
Great, thank you for the confirmation and help with this!

PR #2378, opened 2026-04-26
ci: add erfa as explicit build dependency

ckemper67:
ERFA is required by the SPK math plugin introduced in #2376. This PR adds
liberfa-dev / erfa-devel / erfa to all build environments so that once
#2376 merges, the SPK plugin builds everywhere without the optional guard.

Changes:
- .github/docker/Dockerfile.* — all four CI images
- debian/control — Build-Depends
- scripts/requisites-install.sh — all five package managers
- spec/libindi.spec — RPM BuildRequires


knro:
Any concrete reason to use ERFA over libnova that's already a dependency?

ckemper67:
> Any concrete reason to use ERFA over libnova that's already a dependency?

The SPK library I am vendoring in #2376  was written against the SOFA (iauXxx) API. Since SOFA carries a non-standard restrictive license incompatible with LGPL distribution, the plugin uses ERFA (the ESO/Astropy open-source SOFA fork, mathematically identical). A thin header shim (libs/alignment/spk/sofa.h) maps the API at compile time without modifying Wallace's sources.

knro:
Right, but does it offer anything that libnova does not offer already? I'm not familiar with it, but it appears we will have no astronomical calculations libraries as dependencies instead of just one.

ckemper67:
SOFA does offer a lot of the same astronomical functions as libnova, but it also has functions for coordinate transformations, rotations etc. and it is mostly those that are used in #2376. The biggest reason is that it is a dependency of SPK and to use SPK we have to link against SOFA/ERFA. 

The other option would be to rewrite SPK. I actually tried that initially, but I ran against subtle incompatibilities in the rotation semantics and was afraid that we would get other incompatibilities that are difficult to pin down. The math gets quite subtle. I was surprised myself about some of these issues. 

I documented the rotation issue in ARCHITECTURE.md in #2376, so that we don't run into the same problem later on:

#### ERFA Rotation Matrix Convention Note

The internal matrix generator (`spkVtel`) uses `iauRx`, `iauRy`, and `iauRz`. It is important to note that ERFA does **not** follow the standard "active" rotation matrix format.
1. **Passive**: They are "alias" rotations (they rotate the coordinate frame, not the vector). Each matrix is the *transpose* of standard active matrices.
2. **Left-multiplication**: They stack left-to-right rather than right-to-left.

Substituting standard $M^T_{active}$ math engines here will result in polar offset (`AN, AW`) coefficients being applied inversely, which would cause systematic pointing errors.

ckemper67:
One other option would be to vendor the ERFA sources we need inside SPK. Since SPK only uses a small subset that would be feasible. It is a bit more involved compared to simply linking against ERFA, but it would prevent the introduction of a second astronomy library just for SPK.

ckemper67:
What we would need:
  - 84 .c files (transitive closure of our 20 used functions out of 251 total ERFA source files)
  - Headers: erfa.h, erfam.h, erfaextra.h, erfadatextra.h

  Trade-offs:
  - Pro: zero external dependency, CI passes everywhere, builds on any platform, no new astronomy library
  - Con: we now own the ERFA sources, but ERFA is extremely stable (astronomy standards don't change), so this is low risk.
  - The 85 files add ~ 758 KB of source
  -  the total size of ERFA would be ~2.2 MB if we want to vendor the full ERFA source inside SPK instead of just the subset

ckemper67:
I tried another approach. The SPK kernel mostly use the math functions of ERFA, so I [created a combination of source subset and shim](https://github.com/ckemper67/indi/compare/spk-math-plugin...ckemper67:indi:spk-erfa-libnova-shim?expand=1). Let me know what you think.

knro:
What would be now the total size after your latest changes?

ckemper67:
Roughly 164k for the additional ERFA sources
```
$ du -h libs/alignment/erfa
164K    libs/alignment/erfa
```

knro:
Ok, I think it's best to leave this as an external dependency.

PR #2387, opened 2026-05-08
Alignment: add TransformCelestialToTelescopeJD / TransformTelescopeToCelestialJD

ckemper67:
## Reviewer note

This PR changes a core alignment subsystem API (`MathPlugin`, `MathPluginManagement`)
that is used by multiple drivers inside this repo and by third-party drivers in
`indi-3rdparty` (notably `indi-celestronaux` and `indi-eqmod`). Would appreciate
architectural feedback before merge.

## Summary

Refactors the INDI alignment subsystem so that the core transform methods no longer read
the system clock internally. Two new methods are added to `MathPlugin` that accept an
absolute Julian Date directly:

- `TransformCelestialToTelescopeJD(RA, Dec, JulianDate, TDV)`
- `TransformTelescopeToCelestialJD(TDV, RA, Dec, JulianDate)`

All built-in plugins (`BasicMathPlugin` / `SVDMathPlugin`, `NearestMathPlugin`,
`SPKMathPlugin`, `DummyMathPlugin`) implement the new methods as their primary
implementation. The old `TransformCelestialToTelescope` / `TransformTelescopeToCelestial`
methods become thin compat shims that resolve `ln_get_julian_from_sys() + JulianOffset`
and delegate to the new JD variants. `MathPluginManagement` gains matching `*JD` dispatch
methods; its old public methods are similarly shimmed.

## Motivation

The old interface forced every plugin to call `ln_get_julian_from_sys()` internally.
This had two consequences:

1. **Tests were non-deterministic.** The test suite worked around this with a
   `joff = fixedJD - ln_get_julian_from_sys()` hack to re-anchor transforms to a fixed
   point in time. This was fragile and obscured intent.

2. **Simulated time was impossible.** Any future support for injected/simulated time
   (e.g. KStars time simulation) requires callers to supply the JD — the plugins cannot
   intercept `ln_get_julian_from_sys()` from the outside.

## Backward compatibility

External plugins that have not yet adopted the new interface are unaffected at the
source and binary level. The new `*JD` methods are non-pure virtual. The base-class
default computes `JulianOffset = JulianDate - ln_get_julian_from_sys()` and calls the
old pure-virtual, so an unmodified external plugin recovers the correct absolute JD
when it adds `ln_get_julian_from_sys()` internally.

All existing drivers that call the old `TransformCelestialToTelescope` /
`TransformTelescopeToCelestial` interface — including `skywatcherAPIMount`,
`indi-celestronaux`, `eqmod`, `telescope_simulator`, `astrotrac`, `temmadriver`, `dsc` —
continue to work unchanged through the compat shims. Non-zero `JulianOffset` calls
(e.g. bracket tracking in `skywatcherAPIMount` and `celestronaux`) are correctly
handled: the shim resolves the absolute JD before dispatching.

### `libs/alignment/MathPlugin.h/.cpp`
- Add `TransformCelestialToTelescopeJD` and `TransformTelescopeToCelestialJD` as
  non-pure virtual methods with default fallback implementations.

### `libs/alignment/BasicMathPlugin.h/.cpp`, `NearestMathPlugin`, `SPKMathPlugin`, `DummyMathPlugin`
- Override the new `*JD` methods as the primary implementation (clock-free).
- Reduce the old `Transform*` overrides to one-liner shims.

### `libs/alignment/MathPluginManagement.h/.cpp`
- Add public `TransformCelestialToTelescopeJD` / `TransformTelescopeToCelestialJD`
  dispatch methods (direct virtual call on `pLoadedMathPlugin`).
- Convert old public `Transform*` methods to compat shims.
- Remove now-dead `pTransformCelestialToTelescope` / `pTransformTelescopeToCelestial`
  function pointer members (dispatch goes through direct virtual calls).

### `test/alignment/test_alignment_plugins.cpp`
- Remove the `joff = fixedJD - ln_get_julian_from_sys()` workaround from all test
  blocks; replace with direct `*JD` calls using `static constexpr double fixedJD`.
- Add two regression tests (`LegacyPlugin_JD_Fallback_CelestialToTelescope` and
  `LegacyPlugin_JD_Fallback_TelescopeToCelestial`) that verify the base-class fallback
  correctly converts an absolute JD to the offset expected by legacy external plugins.

## Test results

```
[==========] 110 tests from 2 test suites ran.
[  PASSED  ] 110 tests.
```

## Follow-up (not in this PR)

**Phase 2 — Deprecate:** Mark the old `Transform*` pure-virtual methods `[[deprecated]]`
to produce compiler warnings for external plugins that haven't migrated.

**Phase 3 — Driver API:** Thread a `JD` parameter (with default `= ln_get_julian_from_sys()`)
through the six public methods of `AlignmentSubsystemForDrivers`
(`SkyToTelescopeEquatorial`, `TelescopeEquatorialToSky`, etc.) and update the internal
call sites to use `*JD` directly. Migrate direct `MathPluginManagement` callers
(`skywatcherAPIMount`, `indi-celestronaux`) to use `*JD` with explicit JD values.

**Phase 4 — Remove:** Once no callers use the old interface, make `*JD` pure virtual and
remove the compat layer entirely.

### DIFF
--- a/libs/alignment/BasicMathPlugin.cpp
+++ b/libs/alignment/BasicMathPlugin.cpp
@@ -97,10 +97,8 @@ bool BasicMathPlugin::Initialise(InMemoryDatabase *pInMemoryDatabase)
                 case NORTH_CELESTIAL_POLE:
                 {
                     INDI::IEquatorialCoordinates DummyRaDec;
-                    //INDI::IHorizontalCoordinates DummyAltAz;
                     DummyRaDec.rightascension  = 0.0;
                     DummyRaDec.declination = 90.0;
-                    //EquatorialToHorizontal(&DummyRaDec, &Position, ln_get_julian_from_sys(), &DummyAltAz);
                     DummyActualDirectionCosine2   = TelescopeDirectionVectorFromEquatorialCoordinates(DummyRaDec);
                     DummyApparentDirectionCosine2 = DummyActualDirectionCosine2;
                     break;
@@ -108,10 +106,8 @@ bool BasicMathPlugin::Initialise(InMemoryDatabase *pInMemoryDatabase)
                 case SOUTH_CELESTIAL_POLE:
                 {
                     INDI::IEquatorialCoordinates DummyRaDec;
-                    //INDI::IHorizontalCoordinates DummyAltAz;
                     DummyRaDec.rightascension  = 0.0;
                     DummyRaDec.declination = -90.0;
-                    //EquatorialToHorizontal(&DummyRaDec, &Position, ln_get_julian_from_sys(), &DummyAltAz);
                     DummyActualDirectionCosine2   = TelescopeDirectionVectorFromEquatorialCoordinates(DummyRaDec);
                     DummyApparentDirectionCosine2 = DummyActualDirectionCosine2;
                     break;
--- a/libs/alignment/BasicMathPlugin.cpp
+++ b/libs/alignment/BasicMathPlugin.cpp
@@ -366,6 +366,15 @@ bool BasicMathPlugin::TransformCelestialToTelescope(const double RightAscension,
         double JulianOffset,
         TelescopeDirectionVector &ApparentTelescopeDirectionVector)
 {
+    return TransformCelestialToTelescopeJD(RightAscension, Declination,
+                                           ln_get_julian_from_sys() + JulianOffset,
+                                           ApparentTelescopeDirectionVector);
+}
+
+bool BasicMathPlugin::TransformCelestialToTelescopeJD(double RightAscension, double Declination,
+        double JulianDate,
+        TelescopeDirectionVector &ApparentTelescopeDirectionVector)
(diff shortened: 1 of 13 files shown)


knro:
Great work! I agree this would make all the testing and simulation deterministic now. You can mark the shim functions as deprecated. At the same time, another PR for all affected drivers should be prepared to migrate them to the new functions.

ckemper67:
Yes I will send the follow-up PRs once we have this submitted. Do you want the depreciation as part of this pull request or in a follow up as well?

knro:
Sure you can add it to this PR as well.

ckemper67:
On second thought, I would prefer to combine the deprecation declaration with the PR that fixes the drivers. That way we don't introduce deprecation warnings/errors into the build. What do you think?

#2388 is the followup. I based it on master, so it contains this PR as well. The last commit are the deprecation changes and the cleanups.

PR #2388, opened 2026-05-10
Alignment: deprecate old Transform* shims, migrate all in-tree callers to *JD

ckemper67:
## Reviewer note

This PR is stacked on #2387 and targets that branch as its base. After #2387 merges to master, this branch will be rebased onto master and the
PR base retargeted accordingly.

## Summary

Follow-up to #2387. That PR added `TransformCelestialToTelescopeJD` /
`TransformTelescopeToCelestialJD` to the math plugin stack and converted the old
`TransformCelestialToTelescope` / `TransformTelescopeToCelestial` methods in
`MathPluginManagement` into compat shims. This PR completes Phase 2 and Phase 3 of
the migration plan described in that PR's follow-up section:

- **Phase 2** -- Mark the old shim methods `[[deprecated]]` in `MathPluginManagement.h`
  so any caller that has not yet migrated gets a compiler warning.
- **Phase 3** -- Thread an explicit `JD` parameter through
  `AlignmentSubsystemForDrivers` and update every in-tree driver and example call
  site to use the `*JD` variants directly.

### `libs/alignment/MathPluginManagement.h`

Added `[[deprecated("Use TransformCelestialToTelescopeJD with an explicit Julian Date")]]`
to both old public shim methods. Their implementations are unchanged.

Note: the old pure-virtual methods in `MathPlugin.h` are intentionally left
un-deprecated. External math plugins must still implement them; their deprecation is
Phase 4 work after the broader ecosystem migrates.

### `libs/alignment/AlignmentSubsystemForDrivers.h/.cpp`

Added `double JD = ln_get_julian_from_sys()` as a trailing default parameter to all
six public helper methods:

- `AddAlignmentEntryEquatorial` / `AddAlignmentEntryAltAz` -- use `JD` for
  `ObservationJulianDate` (previously always captured `ln_get_julian_from_sys()`)
- `SkyToTelescopeEquatorial` / `SkyToTelescopeAltAz` -- call
  `TransformCelestialToTelescopeJD(ra, dec, JD, TDV)` instead of the old shim
- `TelescopeEquatorialToSky` / `TelescopeAltAzToSky` -- call
  `TransformTelescopeToCelestialJD(TDV, ra, dec, JD)` instead of the old shim

Existing callers that omit the `JD` argument see no source change and identical
behavior (the default evaluates `ln_get_julian_from_sys()` at the call site).
Callers that want simulated time support can now pass an explicit JD.

### Driver and example call sites

All remaining in-tree uses of the old `MathPluginManagement` shims are updated to
call `*JD` directly with `ln_get_julian_from_sys()`:

| File | Sites |
|------|-------|
| `drivers/telescope/skywatcherAPIMount.cpp` | 9 |
| `drivers/telescope/astrotrac.cpp` | 2 |
| `drivers/telescope/telescope_simulator.cpp` | 2 |
| `drivers/telescope/temmadriver.cpp` | 3 |
| `drivers/telescope/dsc.cpp` | 2 |
| `examples/tutorial_seven/simple_telescope_simulator.cpp` | 3 |

`skywatcherAPIMount` non-zero offset sites (`JulianOffset` axis property and bracket
tracking) now compute the absolute JD explicitly
(`ln_get_julian_from_sys() + JDOffset`) instead of passing a relative offset through
the shim. The `JDnow` variable used by the bracket tracking else-branch is hoisted
before the if/else so both branches sample the clock at the same instant.

## Backward compatibility

No behavior change for any existing caller. All drivers that go through
`AlignmentSubsystemForDrivers` pick up the default `JD` argument and continue to
call the system clock as before. Third-party drivers (`indi-celestronaux`, `indi-eqmod`,
etc.) that call `MathPluginManagement` directly will see deprecation warnings on
their next rebuild but continue to function correctly through the shim.

## Test results

```
1/2 Test  #9: test-alignment ...................   Passed    0.50 sec
2/2 Test #10: test-alignment-plugins ...........   Passed    0.36 sec

100% tests passed, 0 tests failed out of 2
```

## Follow-up (not in this PR)

**indi-3rdparty driver migration:** Third-party drivers that call
`TransformCelestialToTelescope` / `TransformTelescopeToCelestial` directly
(notably `indi-celestronaux` and `indi-eqmod`) will see deprecation warnings on
their next rebuild. A follow-up PR to `indi-3rdparty` should update those call
sites to use `*JD` with explicit `ln_get_julian_from_sys()`, mirroring the
pattern applied here.

**Phase 4 -- Remove:** Once external plugins have migrated, the compat layer can be
fully removed: make `*JD` pure virtual in `MathPlugin.h`, remove the old pure
virtuals and their shim overrides in built-in plugins, remove the deprecated
`MathPluginManagement` shim methods. The full sequence is described in the
follow-up section of #2387.

### DIFF
--- a/drivers/telescope/astrotrac.cpp
+++ b/drivers/telescope/astrotrac.cpp
@@ -682,7 +682,7 @@ bool AstroTrac::ReadScopeStatus()
     m_MountInternalCoordinates.declination = de;
     TDV = TelescopeDirectionVectorFromEquatorialCoordinates(m_MountInternalCoordinates);
 
-    if (TransformTelescopeToCelestial(TDV, skyRA, skyDE))
+    if (TransformTelescopeToCelestialJD(TDV, skyRA, skyDE, ln_get_julian_from_sys()))
     {
         // double lst = get_local_sidereal_time(LocationNP[LOCATION_LONGITUDE].getValue());
         // double dHA = rangeHA(lst - skyRA);
@@ -710,7 +710,7 @@ bool AstroTrac::ReadScopeStatus()
 bool AstroTrac::getTelescopeFromSkyCoordinates(double ra, double de, INDI::IEquatorialCoordinates &telescopeCoordinates)
 {
     TelescopeDirectionVector TDV;
-    if (TransformCelestialToTelescope(ra, de, 0.0, TDV))
+    if (TransformCelestialToTelescopeJD(ra, de, ln_get_julian_from_sys(), TDV))
     {
         EquatorialCoordinatesFromTelescopeDirectionVector(TDV, telescopeCoordinates);
         LOGF_DEBUG("TransformCelestialToTelescope: RA=%lf DE=%lf, TDV (x :%lf, y: %lf, z: %lf), local hour RA %lf DEC %lf",
--- a/drivers/telescope/dsc.cpp
+++ b/drivers/telescope/dsc.cpp
@@ -468,7 +468,7 @@ INDI::IEquatorialCoordinates DSC::TelescopeEquatorialToSky()
         eq.declination = encoderEquatorialCoordinates.declination;
         TDV    = TelescopeDirectionVectorFromLocalHourAngleDeclination(eq);
 
-        if (!TransformTelescopeToCelestial(TDV, RightAscension, Declination))
+        if (!TransformTelescopeToCelestialJD(TDV, RightAscension, Declination, ln_get_julian_from_sys()))
         {
             RightAscension = encoderEquatorialCoordinates.rightascension;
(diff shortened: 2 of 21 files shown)


ckemper67:
How would we fix the drivers in indi-3rdparty? i.e. what is the sequence of events to make a clean transition? If we merge this, then a build with `-Werror` would fail if it encounters the deprecated methods, right? I think some of our CI pipelines do that.

Merge #2387, then fix 3rdparty, then merge #2388?
Or remove the deprecation from #2388, merge both PRs, then update 3rd-party, then add the deprecation after everything is clean?

knro:
Thanks. Now the 3rd party drivers needs to be fixed next

PR #2389, opened 2026-05-10
Fix updating of rotator backlash & reverse direction

escribana:
This PR fixes two signifiicant errors introduced with https://github.com/indilib/indi/commit/b74820236c65456d32591f330c50cb1989efc828.

The switches "ReverseRotatorSP" and "RotatorBacklashSP" didn't update correctly: Both were sticky when switched on. The bug is mentioned [here](https://indilib.org/forum/astro-arch/15662-what-about-derotating-feature-for-alt-az-user.html?start=48#107479).

I couldn't test "RotatorBacklash", but it has the very same structure like "ReverseRotator". So I decided to adapt this code too.


knro:
THat is indeed a critical error! I found a couple more so I'll close this and submit all fixes.. thanks for discovering this!!

PR #2390, opened 2026-05-10
TCP retry on establishing connection

miceno:
This is a PR to refactor the Connection::TCP class to allow reconnect in case of network failures. It implements exponential backoff delay, with suitable limits to prevent blocking the driver.

It adds two config params to the Connection:

- Number of retries.
- Base backoff delay.


knro:
Thank you. How would a connection loss be detected in this case? It appears this only applies to the initial handshake?

miceno:
Yes, it is only for the initial connection. There is no method on the TCP object to read data, so it is not possible as of now that the TCP class will reconnect in case a read/write error on the network occurs. Network operations are delegated to clients of the class, for example, Dome interface creates a TCP connection but it does not read nor write data to the port, it is an implementation like Rolloffino class that sends and receives the data.

I feel we can add status to the TCP connection, by adding a sort of keep-alive, but it requires modifying the interface, not sure how the approach should be, but I have a look at it if you consider it the right way.

On the other hand there is also the approach of adding the reconnect logic to the client class. In this case, it will be like:
1. detect read or write operation didn't succeed due to a network disconnect.
2. disconnect
3. reconnect
4. try again
I feel part of this logic can be encapsulated in the TCP class too.

Not sure about the right approach.

PR #2405, opened 2026-05-30
Pegasus Flatmaster Neo Driver

DomMer2587:
Base driver for Flatmaster Neo. 
Dustcap, lightbox and weather Interface Integration.

More Features will be integrated after I will get the command List.


knro:
Thank you! Please change copyright header to include you. Also please submit [Driver Documentation](https://github.com/indilib/drivers-docs) as required for all new drivers.

This was tested already on the hardware? which command sets are pending?

DomMer2587:
documentation is in work.
At the moment some commands for manual dew control are missing. Complete WLAN control.
From the device status some information is missing which i think is linked to dew control.
 
Main functions like lightbox control, intesity control, park/unpark dustcap and set dustcap angle are working, also the automatic dew control.

PR #2409, opened 2026-06-05
Add MX-HD telescope driver

botchASTRO:
This PR adds a new INDI telescope driver for the MX-HD equatorial mount.

Summary
- Supports USB serial and Bluetooth RFCOMM serial connections
- Implements Goto / Sync
- Implements Park / Unpark
- Implements HOME via driver-specific MXHD_HOME action
- Implements pulse guiding
- Implements standard TELESCOPE_TRACK_MODE
- Implements HA-based TELESCOPE_PIER_SIDE for GEM operation
- Implements INDI-provided site/time synchronization
- Implements standard slew-rate selection

Implementation notes
- The mount does not report pier side directly, so TELESCOPE_PIER_SIDE is derived from hour angle.
- Unpark is implemented as a HOME-return sequence before clearing the parked state, because MX-HD requires a known HOME pose to restore a valid mount attitude/position solution for reliable remote operation.
- Home is exposed as a driver-specific MXHD_HOME action due to client compatibility issues observed with standard TELESCOPE_HOME handling.
- The driver supports Bluetooth when the operating system exposes the mount as a serial RFCOMM device.

Tested with
- Real MX-HD hardware
- Raspberry Pi 5
- libindi 2.2.1
- KStars/Ekos
- USB serial (/dev/ttyUSB0)
- Bluetooth RFCOMM serial (/dev/rfcomm0)

Documentation
- Included driver documentation in drivers/telescope/mxhd_telescope.md

Reference
- Project repository: https://github.com/botchASTRO/mx-hd_indi
- Manufacturer/product page: http://gototelesco.co.jp/mxhd.html
-

### DIFF
--- a/drivers/telescope/mxhd_serial.cpp
+++ b/drivers/telescope/mxhd_serial.cpp
@@ -17,14 +17,30 @@
 *******************************************************************************/
 #include "mxhd_serial.h"
 
+#include <indicom.h>
+
 #include <cerrno>
-#include <chrono>
 #include <cstring>
-#include <sys/select.h>
 #include <termios.h>
 #include <unistd.h>
 
 MXHDSerial::MXHDSerial(int fd) : m_fd(fd) {}
+
+static void setTTYError(int rc, std::string &err)
+{
+    char msg[1024] = {0};
+    tty_error_msg(rc, msg, sizeof(msg));
+    err = msg;
+}
+
+static void splitTimeoutMs(int timeoutMs, long &seconds, long &microseconds)
+{
+    if (timeoutMs < 0)
+        timeoutMs = 0;
+
+    seconds      = timeoutMs / 1000;
+    microseconds = static_cast<long>(timeoutMs % 1000) * 1000L;
+}
 
 bool MXHDSerial::discardInput(std::string &err)
 {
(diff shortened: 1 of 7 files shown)


drivers/telescope/mxhd_serial.cpp
```diff
@@ -0,0 +1,228 @@
...
+        }
+        if (r == 0)
+        {
+            err = "read timeout";
+            return false;
+        }
+        return true;
+    }
+}
+
```
knro, line 90: is there you're using these low level calls when you can use indicom standard serial calls?

botchASTRO: You are right. I initially kept a small protocol-local serial wrapper because MX-HD mixes hash-terminated ASCII replies, 1-byte ACK replies, and a fixed 3-byte binary status response.

However, I agree that for indi-core it is better to use the standard indicom tty_* helpers wherever possible. I will refactor the low-level read/write/select paths to use tty_write/tty_write_string/tty_read/tty_nread_section and keep only the protocol-specific framing logic in the wrapper.

botchASTRO: Thanks for pointing this out. I refactored the MX-HD serial helper to use the standard indicom tty_* calls wherever possible instead of direct low-level read/write/select handling.

I kept only the protocol-specific framing logic for MX-HD replies:
- hash-terminated ASCII replies
- one-byte ACK replies
- fixed 3-byte binary status replies
- RX flush before specific command/query patterns

I rebuilt the updated driver in the standalone MX-HD driver tree on Raspberry Pi and re-tested it on real MX-HD hardware. Connect, RA/Dec updates, Goto, Abort, Park/Unpark, and pulse guiding all worked as expected.

knro:
Thank you! Can you please submit [Driver Documentation](https://github.com/indilib/drivers-docs) as it is required for all new driver submissions.

botchASTRO:
Thank you. I have now submitted the driver documentation PR as well:

https://github.com/indilib/drivers-docs/pull/26

Please let me know if anything should be adjusted there.

botchASTRO:
Thank you. I have added the GOTO INC manufacturer logo to the driver documentation PR as well:

https://github.com/indilib/drivers-docs/pull/26

PR #2410, opened 2026-06-06
driverio: Fix multi-device BLOB interleave crash in unix IO path

cfuture81:
## Problem

When multiple device threads in a HotPlug multi-device driver simultaneously write BLOB data, the INDI XML protocol stream between the driver and indiserver can get corrupted. This causes XML parse errors followed by a segfault in libindidriver.so, crashing both the driver and KStars.

## Root Cause

In the unix IO path (`driverio_init_unix()`), the `stdout_mutex` was only acquired lazily inside `driverio_flush()`, not at initialization. When the output buffer exceeds `OUTPUTBUFF_FLUSH_THRESOLD` (64KB), an intermediate flush sends a partial XML message. Between this flush and the final flush at `driverio_finish()`, another thread's message can be sent, interleaving the XML stream.

The stdout path (`driverio_init_stdout()`) already locks the mutex upfront — this fix brings the unix path in line with that behavior.

## Failing scenario

- 3 SVBONY cameras (`SV605CC` + `SV605MC` + `SV305C`) in one `indi_svbony_ccd` process (HotPlug multi-device)
- Guide camera (SV305C) in streaming mode, two imaging cameras capturing at 0.5s
- Crashes after 6–60 minutes
- Reproduced on both x86_64 (Ubuntu 25.04) and aarch64 (RPi 5 / StellarMate)
- could be not vendor /  SVBONY SDK-specific — any multi-device HotPlug driver could be affected in a similar setup

## Symptoms Before Crash

```
indi_svbony_ccd XML error: Line 2: Bogus tag char <
indi_svbony_ccd XML error: Line 2: Bogus tag char /
indi_svbony_ccd dispatch error: oneNumber requires 'device' attribute (×11)
```

Followed by segfault in `libindidriver.so` (thread 79, the streaming worker).

## Fix

Acquire `stdout_mutex` at `driverio_init_unix()` (2-line change), ensuring the entire XML generation + flush + sendmsg sequence is atomic per message. The existing `driverio_flush()` already checks `dio->locked` before attempting to lock, so there's no double-lock risk.

## Testing

pending feedback from Discord channel user

### DIFF
--- a/libs/indibase/indidriverio.c
+++ b/libs/indibase/indidriverio.c
@@ -327,10 +327,11 @@ static void driverio_init_unix(driverio * dio)
     dio->user = (void*)dio;
     dio->joins = NULL;
     dio->joinSizes = NULL;
-    dio->locked = 0;
     dio->joinCount = 0;
     dio->outBuff = NULL;
     dio->outPos = 0;
+    pthread_mutex_lock(&stdout_mutex);
+    dio->locked = 1;
 }
 
 static void driverio_finish_unix(driverio * dio)


cfuture81:
confirmed by tests by the bug reporter, segfaults and kstars crash are now prevented.

knro:
Thanks, were there any performance regressions?

cfuture81:
> Thanks, were there any performance regressions?

Hey!  its hard to measure the diff in a real world scenario, as it would be extremely minimal.

There is no measurable performance impact expected:

the situation Before: The unix IO path acquired the mutex lazily — only inside driverio_flush() when the buffer exceeded 64KB. Small messages never locked at all. This left a window where two threads could flush partial XML simultaneously, corrupting the stream.

After: The mutex is acquired at driverio_init_unix() and released at driverio_finish_unix(), matching the existing driverio_init_stdout() behavior. The lock now covers the full message lifetime (buffer build + flush).

Why impact is negligible: On the unix IO path, BLOB payloads travel via shared memory (SCM_RIGHTS fd passing). The sendmsg() call only carries the XML envelope (a few hundred bytes) plus the file descriptor — not the multi-MB frame data. So the actual serialization window per message is: vsnprintf of XML tags + one sendmsg of ~hundreds of bytes. That's microseconds for property updates and low single-digit ms for BLOB metadata envelopes.

The only scenario where one thread blocks another is if they emit messages concurrently — and even then, the wait is bounded by one XML envelope write, not by frame size.

PR #2423, opened 2026-06-19
Myfocuserpro2 - Support for devices without Temperature Probes

TerminalMan:
The current MyFocuserPro2:readTemperature() function expects a response otherwise it waits for a timeout, which is not necessary.

The MyFocuserPro2 Devices expose their temperature support through :83# -> 1: temp support 0: no temp support.

The driver should only query for a temp, if :83# returns 1. 

Without this modification the driver waits until it reaches a timeout, which is unnecessary.


knro:
Thank you!

TerminalMan:
Thanks for accepting it so quickly. :)

PR #2427, opened 2026-06-21
fix: suppress all strncpy warnings.

TallFurryMan:
This is a global run on all strncpy issues in the repository.


knro:
I think the recommendation is to drop it completely and replace it with snprintf which I did partially in a previous PR.

TallFurryMan:
Sure, that's even better indeed.

PR #2429, opened 2026-06-25
Alignment refactor

knro:
Refactor alignment to comply with INDI new style properties (partially done) and to move away from cursor-based interface. Also, send JSON payload when the model is loaded or updated so that clients always have an up-to-date list of all sync points.

### DIFF
--- a/libs/alignment/InMemoryDatabase.cpp
+++ b/libs/alignment/InMemoryDatabase.cpp
@@ -274,16 +274,6 @@ static const char *MountAlignmentToString(MountAlignment_t alignment)
     }
 }
 
-// Helper: map JSON string to MountAlignment_t enum
-static MountAlignment_t StringToMountAlignment(const std::string &str)
-{
-    if (str == "NORTH_CELESTIAL_POLE")
-        return NORTH_CELESTIAL_POLE;
-    if (str == "SOUTH_CELESTIAL_POLE")
-        return SOUTH_CELESTIAL_POLE;
-    return ZENITH;
-}
-
 std::string InMemoryDatabase::SerializeModelToJSON(MountAlignment_t MountAlignment) const
 {
     using json = nlohmann::json;
--- a/libs/alignment/InMemoryDatabase.cpp
+++ b/libs/alignment/InMemoryDatabase.cpp
@@ -10,6 +10,7 @@
 
 #include "basedevice.h"
 #include "indicom.h"
+#include "../indijson.hpp"
 
 #include <cerrno>
 #include <cstdlib>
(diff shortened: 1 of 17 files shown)


knro:
@ckemper67 Given you made substantial changes to the INDI Alignment Subsytem recently, it would be great if you can review this or even suggest a better alternative if you have anything in mind.

ckemper67:
> @ckemper67 Given you made substantial changes to the INDI Alignment Subsytem recently, it would be great if you can review this or even suggest a better alternative if you have anything in mind.

I will take a look

ckemper67:
I must admit, that I never really paid attention to how the InMemoryDataBase works and interacts with the properties. Here are some first general observations and questions:

I think the new approach is clearly better than the cursor, but the interaction between the different properties and the overall sequencing of events still makes me nervous without some tests that exercise that behavior. Can you add some tests for the MapPropertiesToInMemoryDatabase?

Are we using the properties as some form of RPC here? i.e. if the AlignmentSaveSP property is set to on that causes a save of the database? My first reaction would be to have some form of "command property" instead of having a property per action, but let me think about that a bit. Do we have other similar interactions through properties in indi? i.e. anything where we manipulate state through multiple properties? I remember I had a property where I continued to publish a legacy value but writes used a new format and you proposed to delete that and just use the new format instead.

For updates could we simply rely on the complete state to be read and written instead? Now that we have the blob property that should be easy, right? i.e. instead of removing an entry at an index through a property, remove the entry client side and push the blob? Since we immediately publish the blob after an update we already have that traffic over the wire.

Another interaction question: Are clients always making changes and then immediately trigger a save in practice? That would mean than we could "autosave" instead to simplify the interaction.

knro:
While I can't say for sure, this is most likely not used by any clients I know of. Furthermore, unless you are trying to edit and delete points yourself, then the current cursor-based interface isn't really used that much in practice. The only property I have issues with is the BLOB because it sends a BLOB (json) after each update with the list of all points. I added support for handling this in KStars, but other clients would need to update to incorporate these BLOBs if they find them useful.

So basically what happens if that you just plate solve, and after SYNC is triggered, a new point is passively added to the model. There are no explicit commands from the client, at least from KStars as far as I know.

ckemper67:
I don't understand how a new alignment BLOB makes it into the alignment module. I can see the deserialization happening, so we do get a new alignment database, but I think we need to call Initialize() somewhere after loading.

libs/alignment/InMemoryDatabase.cpp
```diff
@@ -257,5 +258,115 @@ void InMemoryDatabase::SetLoadDatabaseCallback(LoadDatabaseCallbackPointer_t Cal
...
+                entry.TelescopeDirection.y = v.value("y", 0.0);
+                entry.TelescopeDirection.z = v.value("z", 0.0);
+            }
+
+            entry.PrivateDataSize = 0;
+            MySyncPoints.push_back(entry);
+        }
+    }
+
+    return true;
```
ckemper67, line 368: I would expect this method to somehow call the initialization for the alignment module.

libs/alignment/MapPropertiesToInMemoryDatabase.cpp
```diff
@@ -14,242 +14,467 @@ namespace INDI
...
+            bool success = DeserializeModelFromJSON(JSONData);
+            if (success)
+            {
+                AlignmentModelBlob.setState(IPS_OK);
+                AlignmentModelBlob.apply();
+
+                // Re-initialize the math plugin with the new data
+                // Note: The callback set in AlignmentSubsystemForDrivers constructor
+                // will call Initialise(this). We trigger that here.
+                UpdateModelCount();
```
ckemper67, line 194: I don't understand how this initialization would work and in any case it is difficult to understand that a function called `UpdateModelCount()` has a side effect of initializing the alignment subsystem.

knro: Ok I'm hesitant to make substantial changes now to the Alignment subsystem that might end breaking the INDI drivers. You think we should leave as-is, or you have a proposal?

ckemper67: I am not proposing to change the way this works. I am mostly concerned that calling a function UpdateModelCount() seems to rely on magic to make the initialization happen and that could break when someone refactors something downstream. Ideally this would have a call that also looks like it initializes the alignment subsystem. I also think if you have tests, you can have the confidence that INDI drivers don't break going forward. Can you tests that show how setting properties affects the database?

- add an entry and see that the number of db entries change,
- set a blob and see the number of entries in the database
- remove an entry, clear the db

We also need tests that show that the alignment subsystem gets initialized. i.e. have a mock math plugin where you can observe that it gets initialized when you set the blob, remove an index and clear the database.

PR #2431, opened 2026-06-29
Make wanderer rotator much more robust

olsonse:
These patches improve the stability and robustness of the wanderer rotators, such that serial comms do not get confused anymore.  This patch also ensures that the rotator follows commanded angles, even if these change during the middle of a move operation.

### DIFF
--- a/drivers/rotator/wanderer_rotator_base.cpp
+++ b/drivers/rotator/wanderer_rotator_base.cpp
@@ -290,7 +290,17 @@ bool WandererRotatorBase::Handshake()
     this->reader.start(PortFD);
 
     if (!send_handshake())
-        return false;
+    {
+        /* For some reason, when this is the first time attempting to
+         * communicate to the wanderer rotator after connecting the USB, it does
+         * not respond to the first query immediately after connection.  We will
+         * attempt to send the handshake one more time after having waited for
+         * the handshake response to timeout.  If this does not work, there must
+         * be a real issue.
+         */
+        if (!send_handshake())
+            return false;
+    }
 
     //Device Model//////////////////////////////////////////////////////////////
     LOGF_INFO("Connected to device %s", this->getRotatorHandshakeName());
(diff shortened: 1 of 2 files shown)


knro:
@WandererAstro Can you please take a look?

olsonse:
I added one more commit to fix the rotator angle displayed to the user.  This last commit does not impact functionality.  I also fixed a compiler issue with an unused variable from my original pull request.

WandererAstro:
> @WandererAstro Can you please take a look?

I will take a look in a few days
Thanks all of you :)

olsonse:
Just pushed one more commit to make the initial connection more robust (after initially physically connecting USB).  The restores an attempt in the original code to try a second time to do a handshake--the reason *must* have been because of the same issue I have seen:  the wanderer firmware (at least for the mini v2) does not respond to any commands within about 1.5 seconds after physically connecting the USB and after the first serial connection is attempted.  

As long as the cable remains connected to the computer, it appears subsequent serial connections do not have this limitation and the device responds immediately.

WandererAstro:
@olsonse Hello,
move() currently can:

double delta = range180(abs_angle - current_angle);
0° → 270° gives range180(270) = -90° — turning backwards 90° instead of forwards 270°, taking the virtual mechanical angle below 0°.

On the INDI side this works fine by itself. However, it may cause problem with our ASCOM driver.
To keep both drivers interoperable, the virtual mechanical angle must stay strictly within [0°, 360°). 

Kindly modify this :)

olsonse:
@WandererAstro  I assume that your comment pertains to a desire to keep a functional similarity between the ASCOM drivers and INDI with respect to the Wanderer rotators.  Perhaps the functionality you seek is to ensure an automatic limitation of infinite rotations in either direction that would otherwise tangle cables?

WandererAstro:
> @WandererAstro I assume that your comment pertains to a desire to keep a functional similarity between the ASCOM drivers and INDI with respect to the Wanderer rotators. Perhaps the functionality you seek is to ensure an automatic limitation of infinite rotations in either direction that would otherwise tangle cables?

If the INDI driver makes the rotator move to negative angles, when the user connect this rotator to ASCOM or WandererEmpire, there will be problem.

PR #2436, opened 2026-07-03
Wanderer ETA: Fix first-connect serial garbage

cfuture81:
## Summary

Fix the Wanderer ETA M54 driver failing to connect on first attempt due to stale/garbage bytes in the serial buffer after port open.

## Problem

On initial connection, the USB-serial buffer contains stale bytes from device power-on or a prior session. The first `tty_read_section()` in the handshake reads a line prefixed with garbage, causing `parseDeviceData()` to fail with "Unknown device" and the connection to time out. A manual disconnect + reconnect always worked because the garbage was consumed on the first (failed) attempt.

## Fix

- Flush the serial buffer (`tcflush TCIOFLUSH`) on the first `getData()` call (which serves as the handshake)
- Retry up to 3 times with 3-second timeout each to get a valid status line
- If a read returns data that doesn't parse (partial/garbage), flush again and retry
- Suppress the "Unknown device" warning during handshake retries (use DEBUG level) to avoid confusing users — warning only fires for invalid data after successful initialization

## Testing

- Tested on hardware: firmware 20251123, StellarMate OS 2.2.1
- Verified: clean first-connect after fresh reboot (no manual reconnect needed)
- Verified: no WARNING message visible to user during normal handshake
- Compared with other Wanderer drivers — they use command/response model so don't hit this issue. The ETA uses passive streaming which is susceptible to mid-stream port opening.

### DIFF
--- a/drivers/auxiliary/wanderer_eta.cpp
+++ b/drivers/auxiliary/wanderer_eta.cpp
@@ -422,14 +422,12 @@ bool WandererETA::ISNewSwitch(const char *dev, const char *name, ISState *states
             double pos3 = PositionReadNP[POINT_3].getValue();
 
             // Calculate new targets
-            double newPos1 = pos1 + offset;
-            double newPos2 = pos2 + offset;
-            double newPos3 = pos3 + offset;
+            double targets[3] = {pos1 + offset, pos2 + offset, pos3 + offset};
 
             // Validate all within range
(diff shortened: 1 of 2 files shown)


cfuture81:
Closing — branch contained stale commits from already-merged PR #2433. Re-submitted as clean single-commit PR.

PR #2437, opened 2026-07-03
Wanderer ETA: Fix first-connect serial packages

cfuture81:
## Summary

Major refactor of the Wanderer ETA M54 driver move logic and first-connect reliability.

### 1. Non-blocking move architecture (replaces blocking waitForPosition)

Previously, all moves blocked the INDI event loop until the motor reached its target (or timed out). This prevented live UI updates and made the driver unresponsive during moves.

Now:
- `ISNewNumber`/`ISNewSwitch` sends the serial command and returns immediately
- UI shows IPS_BUSY (yellow indicator) during move
- `TimerHit()` polls device status every 2s, detects when target is reached
- When device reports target position, move completes and shows IPS_OK (green)
- Sequential multi-point moves (ZeroAll, ApplyOffset): one point at a time, each completing before the next starts
- Concurrent move rejection: warns user and resets indicator to IDLE

### 2. Proportional timeout (replaces fixed 15s/60s timeout)

Timeout is calculated from actual travel distance at measured motor speed (~0.015 mm/s) + 25s safety margin. A 0.2mm move gets ~38s, a full 1.2mm move gets ~105s. Only fires as safety fallback if position update never arrives.

### 3. First-connect serial buffer fix

Flush stale serial buffer on first connect with retry loop (up to 3 attempts x 3s). Suppresses "Unknown device" warning during handshake retries (DEBUG level only).

## Position readback behavior

The ETA device is **half-duplex** -- while processing a motor command, it does not stream status updates. `TimerHit()` polls the serial port every 2s, but the device only responds once the motor has stopped and it resumes its status stream. This means:

- During a move: yellow IPS_BUSY indicator is shown, but PositionReadNP stays at the pre-move value
- After motor stops: device sends a status line, `parseDeviceData()` updates PositionReadNP to the final position, `checkMoveProgress()` detects target reached and transitions to IPS_OK (green)

This is a hardware limitation, not a driver choice -- the device simply cannot report intermediate positions while driving a motor.

## Regression risk

Low. No changes to:
- Serial protocol or command format (`%d%.3f\n`)
- Property names, types, or structure (INDI API unchanged for clients)
- Config save/load behavior
- Backfocus offset or Zero All user-facing behavior

Key behavioral differences from merged version:

| Aspect | Before (merged) | After (this PR) |
|--------|-----------------|-----------------|
| Move handling | Blocking (ISNewNumber holds thread 15-100s) | Non-blocking (returns immediately) |
| UI during move | Frozen (no property updates) | Yellow indicator, updates on completion |
| Concurrent moves | Unguarded (serial corruption possible) | Rejected with user warning |
| First connect | Fails on stale buffer, needs reconnect | 3-attempt retry with flush |
| m_SendingCommand flag | Blocked getData() during moves | Removed (not needed) |

## Testing

- Firmware 20251123 and 20260701
- StellarMate OS 2.2.1
- Zero All from 0.95mm: all 3 points reached 0.000 (48s, 49s, 40s per point)
- Backfocus offset 0.55mm: all 3 points reached targets (29s, 30s, 29s per point)
- Individual point moves with concurrent move rejection verified
- Clean first-connect on fresh reboot confirmed
- No false stall/timeout warnings in final version

### DIFF
--- a/drivers/auxiliary/wanderer_eta.cpp
+++ b/drivers/auxiliary/wanderer_eta.cpp
@@ -151,10 +151,6 @@ bool WandererETA::updateProperties()
 ////////////////////////////////////////////////////////////////////////////////////////////////////////
 bool WandererETA::getData()
 {
-    // Skip reading while commands are being sent
-    if (m_SendingCommand)
-        return true;
-
     if (!serialPortMutex.try_lock_for(std::chrono::milliseconds(500)))
     {
         LOG_DEBUG("Serial port is busy, skipping status update");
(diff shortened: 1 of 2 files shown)


knro:
@WandererAstro OK from your side?

PR #2443, opened 2026-07-10
Add Askar WAF focuser driver

3255867418:
This PR adds an INDI Core focuser driver for the Askar WAF USB focuser
(Jiaxing Ruixing Optical Instrument Co., Ltd.).

The device uses a USB CDC virtual serial port and a simple "F…#" command set.
No third-party libraries are required.

Implemented:
- Absolute and relative moves, abort, sync
- Reverse direction and firmware backlash compensation
- Read/set max travel
- Firmware version and model name
- Registration in "drivers.xml" and CMake
- Driver documentation under "drivers/focuser/doc/askar_waf/"

Tested:
- Built on Linux
- Tested with real Askar WAF hardware (firmware 1.0.6)
- Verified connection, absolute/relative move, abort, sync, reverse, backlash, and max position in KStars/Ekos


knro:
Thank you! Please separate the documentation part and submit it to a separate repository for INDI driver documentation here: https://github.com/indilib/drivers-docs

PR #2446, opened 2026-07-12
weather: uranusmeteo: fix GPS altitude, TIME_UTC offset, add Lux

gstax:
## Summary

Three fixes to the Uranus Meteo Sensor driver, found while cross-checking the wire protocol against a hand-reverse-engineered decoder validated against real hardware (firmware 1.5).

- **GPS altitude**: `GP` is parsed as an 8-field response (fix, time, utc offset, lat, lon, sat count, speed, bearing), but the device actually sends **9 fields** — a GPS altitude sits between satellite count and speed, undocumented in Pegasus Astro's own command-set PDF. `LocationNP`'s elevation was filled from the barometric sensor instead, which drifts with weather (pressure changes of a few tens of hPa shift the estimate by tens of meters for a stationary site) since it isn't referenced against the site's true elevation. Now `GPSNP` exposes the real GPS altitude and `LocationNP` uses it.

- **TIME_UTC off by the device's UTC offset**: the code assumed `GP`'s Unix time field was local time and subtracted the offset to get UTC, per a 2022 comment. On real hardware the field is already UTC — it matched `date -u` to the second. The subtraction was double-converting (e.g. 2 hours off under CEST). This is more than cosmetic: `GPSInterface`'s `SystemTimeUpdateSP` can use this value to set the host clock. Fixed to use the raw value directly, and to source the UTC offset from the device's own `GP` field instead of the host's timezone.

- **SQ Lux**: the sixth field of `SQ` (raw lux) was read off the wire but never exposed. Unlike MPAS, which saturates near daylight (observed `0.44 mag/arcsec²` in full sun — not a meaningful night-sky reading), lux stays linear and gives a usable daylight/twilight signal.

## Test plan

Verified against a real Uranus Meteo Sensor (firmware 1.5):
- [x] GPS altitude read 441m / 439m / 433m across three separate fixes, agreeing with the barometric sensor's 432m reading under stable weather (as expected when both should roughly agree).
- [x] `TIME_UTC` matched wall-clock UTC after a fix (previously off by exactly the device's 2h offset).
- [x] `SQ` Lux read 72050 lux in full sunlight, while MPAS read a non-physical 0.44 mag/arcsec² at the same time (illustrating why lux is useful near daylight).
- [x] `scripts/check-codestyle.sh` clean on both changed files.
- [x] Builds cleanly (`INDI_BUILD_DRIVERS=ON`).

### DIFF
--- a/drivers/weather/uranusmeteo.cpp
+++ b/drivers/weather/uranusmeteo.cpp
@@ -94,6 +94,7 @@ bool UranusMeteo::initProperties()
     SkyQualityNP[FullSpectrum].fill("FullSpectrum",  "Full Spectrum", "%.2f", -1000, 1000, 10, 0);
     SkyQualityNP[VisualSpectrum].fill("VisualSpectrum",  "Visual Spectrum", "%.2f", -1000, 1000, 10, 0);
     SkyQualityNP[InfraredSpectrum].fill("InfraredSpectrum",  "Infrared Spectrum", "%.2f", 0, 1, 0.1, 0);
+    SkyQualityNP[Lux].fill("Lux",  "Lux", "%.2f", 0, 1e6, 10, 0);
     SkyQualityNP.fill(getDeviceName(), "SKYQUALITY", "Sky Quality", SKYQUALITY_TAB, IP_RO, 60, IPS_IDLE);
 
     SkyQualityUpdateNP[0].fill("VALUE", "Period (s)", "%.f", 0, 3600, 60, 60);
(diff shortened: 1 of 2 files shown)


knro:
Thank you. Are you sure about the UTC time? was this fixed by some later firmware update to the meteo sensor?

gstax:
Yes, I'm confident about the UTC handling — I re-tested it just now on the real device to be sure, in addition to the original test.

- Firmware: `MV` reports `1.5` in all tests below (both the original PR test and today's re-test), so it's not a firmware-version discrepancy.
- Original test (2026-07-11, no GPS fix yet): the raw `GP` Unix timestamp field matched `date -u` to the second.
- Re-test today (2026-07-21), no fix (`sat=1`): same result, raw field matches system UTC to the second.
- Re-test today with an actual 3D fix (`sat=6`, real position acquired): still matches system UTC exactly, not local time. So this isn't specific to the no-fix case either — the receiver decodes UTC time (from the GPS nav message) independently of computing a position fix, which is normal GNSS behavior (time sync only needs one satellite's nav subframe; position needs 3-4).

So across three independent measurements (with and without a fix, one week apart, same firmware), the `GP` time field is consistently UTC, not local time — confirming the original bug: subtracting the host's UTC offset from an already-UTC value shifts `TIME_UTC` by that offset.

I don't have data on any firmware newer than 1.5, so I can't rule out a fix on a later version — happy to test if you can point me to one.

knro:
I'm on firmware 1.2, and Unity software has an issue and failing to detect the device! At any rate, I'll try to figure out how to update the firmware. Thanks!

gstax:
FYI, I've hit the same detection issue with Pegasus's Unity software — it's pretty flaky about recognizing the device. What worked for me was trying a few different USB cables; one of them just wouldn't get picked up at all, another worked intermittently, and a third worked reliably. Might be worth trying a different cable before digging further.

PR #2448, opened 2026-07-22
WandererRotator: Add configurable move accuracy and fix false "not po…

pmneo:
…wered" error on zero-step moves

Adds a new "Accuracy" number property (ROTATOR_ACCURACY/ACCURACY) to WandererRotatorBase, shared by both the Mini and Lite drivers. It lets the user configure the granularity to which requested rotator angles are rounded before a move is issued (e.g. 0.5 degrees), or disable rounding entirely by setting it to 0.

Motivation: physical backlash/precision of the WandererRotator Mini/Lite mechanism makes sub-degree accuracy meaningless for most setups, so snapping goto requests to a coarser grid avoids unnecessary micro-moves and firmware round-trips. The value is persisted via saveConfigItems and restored on startup via IUGetConfigNumber, mirroring the existing Backlash property.

While testing, rounding surfaced a pre-existing bug: if the client requests an angle that rounds to the same position the rotator is already at, MoveRotator still computed a relative move of 0 steps and sent it to the device. The firmware does not report a new position after a zero-step command (since nothing physically moves), so the driver's TimerHit blocked on tty_read_section until it timed out and logged "Rotator not powered!", even though the rotator was working correctly and simply didn't need to move.

Fix: after applying accuracy rounding, MoveRotator now checks whether the resulting relative angle is ~0 and, if so, returns IPS_OK immediately without writing to the serial port, instead of issuing a no-op move command and waiting for a response that will never arrive.

### DIFF
--- a/drivers/rotator/wanderer_rotator_base.cpp
+++ b/drivers/rotator/wanderer_rotator_base.cpp
@@ -135,7 +135,6 @@ bool WandererRotatorBase::ISNewNumber(const char * dev, const char * name, doubl
             accuracy = AccuracyNP[ACCURACY].getValue();
             AccuracyNP.setState(IPS_OK);
             AccuracyNP.apply();
-            saveConfig(AccuracyNP);
             return true;
         }
 
@@ -259,17 +258,19 @@ bool WandererRotatorBase::Handshake()
 IPState WandererRotatorBase::MoveRotator(double angle)
 {
     if (accuracy > 0)
-        angle = round(angle / accuracy) * accuracy;
+        angle = std::round(angle / accuracy) * accuracy;
 
     angle = angle - GotoRotatorNP[0].getValue();
 
-    // Rounded target matches the current position: nothing to move, and the
-    // firmware won't report a position update for a zero-step command.
-    if (std::abs(angle) < 1e-6)
+    // Gate on the actual step delta the firmware will see. Anything that
+    // rounds to zero steps is a no-op the firmware won't acknowledge, so
+    // return immediately instead of blocking in TimerHit ("not powered").
+    int steps = static_cast<int>(std::round(angle * getStepsPerDegree()));
+    if (steps == 0)
         return IPS_OK;
 
     char cmd[16];
-    int position = (int)(angle * getStepsPerDegree() + 1000000);
+    int position = steps + 1000000;
     positionhistory = angle;
     snprintf(cmd, 16, "%d", position);
     Move(cmd);
--- a/drivers/rotator/wanderer_rotator_base.cpp
+++ b/drivers/rotator/wanderer_rotator_base.cpp
@@ -57,6 +57,11 @@ bool WandererRotatorBase::initProperties()
     BacklashNP[BACKLASH].fill( "BACKLASH", "Degree", "%.2f", 0, 3, 0.1, 0);
     BacklashNP.fill(getDeviceName(), "BACKLASH", "Backlash", MAIN_CONTROL_TAB, IP_RW, 60, IPS_IDLE);
 
+    // ACCURACY
+    IUGetConfigNumber(getDeviceName(), "ACCURACY", "ACCURACY", &accuracy);
+    AccuracyNP[ACCURACY].fill("ACCURACY", "Degree (0=off)", "%.2f", 0, 10, 0.1, accuracy);
+    AccuracyNP.fill(getDeviceName(), "ACCURACY", "Accuracy", MAIN_CONTROL_TAB, IP_RW, 60, IPS_IDLE);
+
     serialConnection->setDefaultBaudRate(Connection::Serial::B_19200);
 
 
@@ -71,11 +76,13 @@ bool WandererRotatorBase::updateProperties()
     {
         defineProperty(SetZeroSP);
         defineProperty(BacklashNP);
+        defineProperty(AccuracyNP);
     }
     else
     {
         deleteProperty(SetZeroSP);
         deleteProperty(BacklashNP);
+        deleteProperty(AccuracyNP);
     }
     return true;
 }
@@ -121,8 +128,26 @@ bool WandererRotatorBase::ISNewNumber(const char * dev, const char * name, doubl
             return true;
         }
 
+        // accuracy
+        if (AccuracyNP.isNameMatch(name))
+        {
+            AccuracyNP.update(values, names, n);
+            accuracy = AccuracyNP[ACCURACY].getValue();
+            AccuracyNP.setState(IPS_OK);
+            AccuracyNP.apply();
+            saveConfig(AccuracyNP);
+            return true;
+        }
+
     }
     return Rotator::ISNewNumber(dev, name, values, names, n);
+}
+
+bool WandererRotatorBase::saveConfigItems(FILE *fp)
+{
+    INDI::Rotator::saveConfigItems(fp);
+    AccuracyNP.save(fp);
+    return true;
 }
 
 bool WandererRotatorBase::Handshake()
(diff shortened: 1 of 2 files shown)


cfuture81:
@pmneo  Nice addition — the snapping is genuinely useful for these mechanisms, and thanks for tracking down the zero-step "not powered" cause. A few enhancement suggestions on `MoveRotator`, one of which is worth taking seriously.

### 1. The zero-move guard should be in *step* space, not a degree epsilon

The guard `std::abs(angle) < 1e-6` is finer than one motor step (`1 / getStepsPerDegree()` ≈ 0.00088° on the Mini). The command itself is built with a truncating cast:

```cpp
int position = (int)(angle * getStepsPerDegree() + 1000000);
```

So any residual move between `1e-6°` and roughly half a step still passes the guard but truncates to `position == 1000000` — i.e. a zero-step command is sent anyway, the firmware stays silent, and `TimerHit` blocks and logs "not powered" again. Same symptom you just fixed, just through a narrower door.

Computing the integer step delta first and gating on that closes it exactly, and using `round` instead of a truncating cast keeps sub-step requests honest:

```cpp
if (accuracy > 0)
    angle = std::round(angle / accuracy) * accuracy;

angle = angle - GotoRotatorNP[0].getValue();

// Gate on the actual step delta the firmware will see. Anything that
// rounds to zero steps is a no-op the firmware won't acknowledge, so
// return immediately instead of blocking in TimerHit ("not powered").
int steps = static_cast<int>(std::round(angle * getStepsPerDegree()));
if (steps == 0)
    return IPS_OK;

char cmd[16];
int position = steps + 1000000;
positionhistory = angle;
snprintf(cmd, 16, "%d", position);
Move(cmd);
return IPS_BUSY;
```

### 2. Minor: `round` → `std::round` for consistency

You already use `std::abs` two lines down; the bare `round` relies on `<cmath>` leaking into the global namespace, which is the same portability trap this driver has elsewhere.

### 3. Very minor / optional

`saveConfig(AccuracyNP)` in `ISNewNumber` writes the config file on every accuracy edit, which is a bit heavier than the rest of the driver (Backlash relies on `saveConfigItems` only). Harmless, just noting the inconsistency — the `saveConfigItems` override you added already covers persistence.

---

Also FYI: while reviewing this I found three unrelated pre-existing bugs in the same file (reverse-flag buffer mis-terminated in `Handshake()`, a `%s` used to log the firmware `int`, and `abs()`-on-double causing a divide-by-zero in `TimerHit` for sub-degree moves). I put those in a separate PR stacked on yours — #2451 — rather than muddy this one. The third is relevant here: your accuracy feature makes sub-degree residual moves more common, and on current master those still hit the `TimerHit` divide-by-zero, so the two changes compose well.

Happy to approve once you've had a look at #1.

drivers/rotator/wanderer_rotator_base.cpp
```diff
@@ -233,8 +258,16 @@ bool WandererRotatorBase::Handshake()
...
+
     angle = angle - GotoRotatorNP[0].getValue();
 
+    // Rounded target matches the current position: nothing to move, and the
+    // firmware won't report a position update for a zero-step command.
+    if (std::abs(angle) < 1e-6)
+        return IPS_OK;
+
     char cmd[16];
     int position = (int)(angle * getStepsPerDegree() + 1000000);
```
cfuture81: The zero-move guard is finer than one motor step (`1/getStepsPerDegree()` ≈ 0.00088° on the Mini) and the position is built with a truncating cast, so a residual move between `1e-6°` and ~half a step still sends a zero-step command → firmware stays silent → `TimerHit` blocks and logs "not powered" again. Gating on the integer step delta closes it exactly (and `round` instead of the truncating cast keeps sub-step requests honest):

```suggestion
    if (accuracy > 0)
        angle = std::round(angle / accuracy) * accuracy;

    angle = angle - GotoRotatorNP[0].getValue();

    // Gate on the actual step delta the firmware will see. Anything that
    // rounds to zero steps is a no-op the firmware won't acknowledge, so
    // return immediately instead of blocking in TimerHit ("not powered").
    int steps = static_cast<int>(std::round(angle * getStepsPerDegree()));
    if (steps == 0)
        return IPS_OK;

    char cmd[16];
    int position = steps + 1000000;
```

drivers/rotator/wanderer_rotator_base.cpp
```diff
@@ -121,10 +128,28 @@ bool WandererRotatorBase::ISNewNumber(const char * dev, const char * name, doubl
...
         }
 
+        // accuracy
+        if (AccuracyNP.isNameMatch(name))
+        {
+            AccuracyNP.update(values, names, n);
+            accuracy = AccuracyNP[ACCURACY].getValue();
+            AccuracyNP.setState(IPS_OK);
+            AccuracyNP.apply();
+            saveConfig(AccuracyNP);
```
cfuture81: Minor/optional: this writes the config file on every accuracy edit, a bit heavier than the rest of the driver (Backlash relies on `saveConfigItems` only). Harmless — your `saveConfigItems` override already covers persistence — just flagging the inconsistency.

cfuture81 (approved):
LGTM

pmneo:
@knro it was reviewed? why is it not getting merged?

knro:
@WandererAstro Can you please check?

drivers/rotator/wanderer_rotator_base.cpp
```diff
@@ -57,6 +57,11 @@ bool WandererRotatorBase::initProperties()
     BacklashNP[BACKLASH].fill( "BACKLASH", "Degree", "%.2f", 0, 3, 0.1, 0);
     BacklashNP.fill(getDeviceName(), "BACKLASH", "Backlash", MAIN_CONTROL_TAB, IP_RW, 60, IPS_IDLE);
 
+    // ACCURACY
+    IUGetConfigNumber(getDeviceName(), "ACCURACY", "ACCURACY", &accuracy);
```
knro: No need, you can add add AccurayNP.load() after AccuracyNP.fill(...)

drivers/rotator/wanderer_rotator_base.cpp
```diff
@@ -121,10 +128,27 @@ bool WandererRotatorBase::ISNewNumber(const char * dev, const char * name, doubl
...
 
+        // accuracy
+        if (AccuracyNP.isNameMatch(name))
+        {
+            AccuracyNP.update(values, names, n);
+            accuracy = AccuracyNP[ACCURACY].getValue();
+            AccuracyNP.setState(IPS_OK);
+            AccuracyNP.apply();
+            return true;
+        }
```
knro, line 139: Use the modern property update [like here](https://github.com/indilib/indi/blob/00b2a5eeeb9530ee77cadc17933e10ed233feba9/drivers/telescope/lx200am5.cpp#L390).

drivers/rotator/wanderer_rotator_base.h
```diff
@@ -74,4 +77,9 @@ class WandererRotatorBase : public INDI::Rotator
     {
         BACKLASH,
     };
+
+    enum
+    {
+        ACCURACY,
+    };
```
knro: If it is only 1 element, no need for enum.

drivers/rotator/wanderer_rotator_base.cpp
pmneo: Like this?
```
    // ACCURACY
    AccuracyNP[0].fill("ACCURACY", "Degree (0=off)", "%.2f", 0, 10, 0.1, 0);
    AccuracyNP.fill(getDeviceName(), "ACCURACY", "Accuracy", MAIN_CONTROL_TAB, IP_RW, 60, IPS_IDLE);
    AccuracyNP.load();
    accuracy = AccuracyNP[0].getValue();
``

knro: No need for last line, the property is already updated after .load()

WandererAstro:
> @WandererAstro Can you please check?

Looks good :)

WandererAstro:
You can merge it, we need to open another pull request recently regarding the new firmware with ID identification for multiple rotators.

PR #2449, opened 2026-07-22
Gemini Flatpanel: add per-filter brightness mode presets

pmneo:
The Presets tab already let users set a per-filter brightness value (FLAT_LIGHT_FILTER_INTENSITY, built dynamically from the snooped filter wheel's FILTER_NAME list via the shared INDI::LightBoxInterface). Gemini Rev2/Lite units also support a device-side Low/High brightness mode (BRIGHTNESS_MODE), but until now that mode was global and had no relation to the active filter, so switching filters only restored the brightness value, not the mode it was calibrated with.

Add a FILTER_BRIGHTNESS_MODE switch vector: one independent ISR_NOFMANY switch per known filter name (On = High, Off = Low), living alongside FLAT_LIGHT_FILTER_INTENSITY in the "Preset" tab. When the snooped filter wheel reports a new active slot, the matching mode is now applied to the device (in addition to the existing brightness-value application) and the global BRIGHTNESS_MODE switch is updated to reflect it. The property is only defined when the connected adapter reports supportsBrightnessMode() (Rev2/Lite), and values persist across restarts via saveConfigItems().

Since FilterIntensityNP's filter-name tracking (population from FILTER_NAME/FILTER_SLOT snoops) is private to LightBoxInterface, add two narrow, default-no-op virtual hooks - FilterNamesUpdated() and FilterSlotChanged() - so GeminiFlatpanel can mirror that per-filter bookkeeping without duplicating LightBoxInterface's snoop-parsing logic. These hooks are additive and inert for every other LightBoxInterface-based driver; verified indi_flipflat, indi_giotto and indi_snapcap still build unchanged.

Verified end-to-end against the simulation adapter: presets and modes are built per filter on snoop, toggling a filter's mode and switching the active filter slot correctly re-applies both the brightness value and the brightness mode, and both round-trip through config save/load.

### DIFF
--- a/drivers/auxiliary/gemini_flatpanel.cpp
+++ b/drivers/auxiliary/gemini_flatpanel.cpp
@@ -29,6 +29,19 @@ bool GeminiFlatpanel::initProperties()
     LI::initProperties(MAIN_CONTROL_TAB, LI::CAN_DIM);
     DI::initProperties(MAIN_CONTROL_TAB);
 
+    // Per-filter brightness mode presets (Low/High). Populated dynamically once filter
+    // names are known, mirroring LightBoxInterface's FilterIntensityNP.
+    FilterBrightnessModeSP.fill(
+        getDeviceName(),
+        "FILTER_BRIGHTNESS_MODE",
+        "Filter Brightness Mode",
+        "Preset",
+        IP_RW,
+        ISR_NOFMANY,
+        60,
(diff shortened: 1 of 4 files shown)


pmneo:
@knro any thougths on this?

pmneo:
@knro do you have time to look into this?

PR #2450, opened 2026-07-22
WandererRotator: fix handshake parse bugs and abs()-on-double hazards

cfuture81:
Three pre-existing bugs in `WandererRotatorBase`, found while reviewing #2448 (the accuracy property). None are introduced by that PR; they are independent and can be merged separately.

### 1. Reverse-flag buffer mis-terminated in `Handshake()`
The reverse-flag response buffer was null-terminated using the **angle** response length (`nbytes_read_M_angle`) instead of its own length (`nbytes_read_M_reverse`). If the two replies differ in length, the reverse state is parsed from a mis-terminated buffer. Same class of bug as the backlash-read fix in #2195.

### 2. `%s` format specifier used for an `int`
On the outdated-firmware path, the firmware version (an `int`) was logged with `%s`, reinterpreting the integer as a `char*` → garbage read / potential crash on the exact path meant to give the user a clear message. Changed to `%d`. Also removed a stray extra argument passed to a non-format `LOG_INFO` on the "no data received" path.

### 3. `abs()` on doubles truncates position / divides by zero
Rotator angles were passed through the C integer `abs()` (unqualified), which truncates sub-degree positions to whole degrees and, in `TimerHit`'s sign term `positionhistory / abs(positionhistory)`, divides by zero for any sub-1° move. Switched the position reads to `std::fabs()` and the sign term to `std::copysign(1.0, positionhistory)`. This matters more now that #2448 encourages sub-degree granularity.

---

7-line change, single file (`drivers/rotator/wanderer_rotator_base.cpp`), syntax-checked against the repo headers.

### DIFF
--- a/drivers/rotator/wanderer_rotator_base.cpp
+++ b/drivers/rotator/wanderer_rotator_base.cpp
@@ -155,7 +155,7 @@ bool WandererRotatorBase::Handshake()
         {
             char errorMessage[MAXRBUF];
             tty_error_msg(rc, errorMessage, MAXRBUF);
-            LOGF_INFO("No data received, the device may not be WandererRotator, please check the serial port!", "Updated");
+            LOG_INFO("No data received, the device may not be WandererRotator, please check the serial port!");
             LOGF_ERROR("Device read error: %s", errorMessage);
             return false;
         }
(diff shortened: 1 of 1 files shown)


cfuture81:
Superseded by #2451, which stacks the same three fixes on top of #2448 (the accuracy PR). Closing this master-based version to avoid duplication.

PR #2451, opened 2026-07-22
WandererRotator: fix handshake parse bugs and abs()-on-double hazards (on top of #2448)

cfuture81:
## Stacked on #2448

This PR is built **on top of #2448** (`WandererRotator: Add configurable move accuracy …`). The first commit here is that PR's commit; the diff will reduce to just the fix commit once #2448 is merged. Please merge #2448 first, or merge this after rebasing onto master once #2448 lands.

The fixes below are pre-existing bugs, independent of the accuracy feature, but they are stacked here because #2448's sub-degree accuracy makes bug 3 (below) considerably more likely to trigger in practice.

---

## Fixes (this PR's own commit)

Three pre-existing bugs in `WandererRotatorBase`:

### 1. Reverse-flag buffer mis-terminated in `Handshake()`
The reverse-flag response buffer was null-terminated using the **angle** response length (`nbytes_read_M_angle`) instead of its own length (`nbytes_read_M_reverse`). If the two replies differ in length, the reverse state is parsed from a mis-terminated buffer. Same class of bug as the backlash-read fix in #2195.

### 2. `%s` format specifier used for an `int`
On the outdated-firmware path, the firmware version (an `int`) was logged with `%s`, reinterpreting the integer as a `char*` → garbage read / potential crash on the exact path meant to give the user a clear message. Changed to `%d`. Also removed a stray extra argument passed to a non-format `LOG_INFO` on the "no data received" path.

### 3. `abs()` on doubles truncates position / divides by zero
Rotator angles were passed through the C integer `abs()` (unqualified), which truncates sub-degree positions to whole degrees and, in `TimerHit`'s sign term `positionhistory / abs(positionhistory)`, divides by zero for any sub-1° move. Switched the position reads to `std::fabs()` and the sign term to `std::copysign(1.0, positionhistory)`.

**This is why the fix is stacked here:** #2448 lets users request sub-degree accuracy (e.g. 0.5°), which makes the divide-by-zero in bug 3 reachable through normal use.

---

7-line change, single file (`drivers/rotator/wanderer_rotator_base.cpp`), syntax-checked against the repo headers with #2448 applied.

### DIFF
--- a/drivers/rotator/wanderer_rotator_base.cpp
+++ b/drivers/rotator/wanderer_rotator_base.cpp
@@ -180,7 +180,7 @@ bool WandererRotatorBase::Handshake()
         {
             char errorMessage[MAXRBUF];
             tty_error_msg(rc, errorMessage, MAXRBUF);
-            LOGF_INFO("No data received, the device may not be WandererRotator, please check the serial port!", "Updated");
+            LOG_INFO("No data received, the device may not be WandererRotator, please check the serial port!");
             LOGF_ERROR("Device read error: %s", errorMessage);
             return false;
         }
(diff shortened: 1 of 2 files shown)


cfuture81:
Superseded by #2452. These three fixes are pre-existing master bugs, independent of #2448 and touching none of its code, so they belong in a standalone PR against master rather than stacked on the accuracy work. Closing this stacked version.

PR #2453, opened 2026-07-23
indiccd: Add gradual warm-up state machine to INDI::CCD base class

pmneo:
When a driver turns the cooler off and a temperature ramp slope is configured, the base class now ramps the sensor target toward 30 °C instead of cutting power immediately.  Completion is detected by temperature stability (< 0.5 °C change over 60 s) so it works correctly with unidirectional TECs that cannot actively heat.

CoolerSP stays in each driver.  The base class exposes three protected helper methods that drivers call from their own ISNewSwitch handler:

  beginCoolerWarmup(currentTemperature)
    Saves the user's cooling target (m_TargetTemperature), then either
    cuts the cooler immediately (no ramp, or already ≥ 30 °C) or starts
    a ramped ascent.  The driver must set its own CoolerSP to IPS_BUSY
    before calling this.

  cancelCoolerWarmup()
    Clears m_CoolerWarmingUp and stops the temperature timer.  Call
    when the user re-enables the cooler mid warm-up.

  resumeCoolingAfterWarmup()
    If a valid target was saved, issues a new SetTemperature() ramp
    toward it.  Call after SetCoolerEnabled(true) succeeds.

  coolerWarmupTick(currentTemperature)
    Must be called from the driver's polling loop.  No-op when not
    warming up; fires SetCoolerEnabled(false) once temperature is stable.

The driver's SetCoolerEnabled(false) override is responsible for transitioning its own CoolerSP to IPS_IDLE on completion.

### DIFF
--- a/drivers/ccd/ccd_simulator.cpp
+++ b/drivers/ccd/ccd_simulator.cpp
@@ -368,7 +368,7 @@ bool CCDSim::updateProperties()
     return true;
 }
 
-int CCDSim::SetTemperature(double temperature)
+int CCDSim::SetTemperature(double temperature, bool enableCooler)
 {
     TemperatureRequest = temperature;
     if (std::abs(temperature - TemperatureNP[0].getValue()) < 0.1)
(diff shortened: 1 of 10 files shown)


pmneo:
@knro global implementation

pmneo:
@knro any thougths on this? I put a high effort into this to implement it globaly as you mentioned?

PR #2455, opened 2026-07-25
skywatcherAltAzMount: fix Alt-Az tracking direction reversal

bcolyn:
trackByRate() had two bugs preventing the altitude axis from tracking correctly whenever a direction reversal was needed:

1. The direction-change check compared AxesStatus[AXIS1] in both branches instead of AxesStatus[axis], so altitude's reversal decision was driven by azimuth's current direction rather than its own.

2. On restart after a stop, SetClockTicksPerMicrostep() (speed) was called before SetAxisMotionMode() (direction/mode). The mount firmware rejected this with a "Motor not stopped" error, even from a full standstill. The working manual-jog path (Slew() / PrepareForSlewing()) always sets motion mode before speed; reorder trackByRate()'s restart branch to match.

Without this fix, altitude would freeze whenever a reversal was needed and the tracking offset would grow unbounded. Tested on real hardware (AZ-EQ6, Wave 150i, and AZ-GTi) with both fixes applied: altitude and azimuth now track continuously with no errors, confirmed via encoder movement over several minutes and multiple GOTO targets. Tested on-sky tracking with Wave 150i.

### DIFF
--- a/drivers/telescope/skywatcherAPIMount.cpp
+++ b/drivers/telescope/skywatcherAPIMount.cpp
@@ -1615,8 +1615,8 @@ bool SkywatcherAPIMount::trackByRate(AXISID axis, double rate)
     if (AxesStatus[axis].FullStop && rate == 0)
         return true;
     // If rate is zero, or direction changed then we should stop.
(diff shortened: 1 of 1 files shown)


knro:
Thank you @bcolyn , this looks good to me.

PR #2456, opened 2026-07-29
    Fix -Wformat errors: format specifier type mismatches breaking Powerpc builds

ADITI-MISHRA310:
Using casts like **(long long)x** fixes the immediate warning, but only at that specific place in the code. Using <inttypes.h> is a more robust solution because it matches the actual type rather than assuming what it is. 

Macros like **PRId64, PRIu64, PRIx64** that are defined in **<inttypes.h>** automatically expand to the correct format specifier for the target architecture (%ld, %lld, I64d, etc.), so you don't have to worry about choosing between **%ld** and **%lld** on ppc64le or any other architecture in the future.

### DIFF
--- a/libs/fpack/fpackutil.c
+++ b/libs/fpack/fpackutil.c
@@ -20,6 +20,7 @@
 #include <fitsio.h>
 #include <fitsio2.h>
 #include "fpack.h"
+#include <inttypes.h>
 
 /* these filename buffer are used to delete temporary files */
 /* in case the program is aborted */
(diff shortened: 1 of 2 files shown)


knro:
Please use static_cast

ADITI-MISHRA310:
Done!!

libs/fpack/fpackutil.c
```diff
@@ -1879,15 +1875,10 @@ int fp_test (char *infits, char *outfits, char *outfits2, fpstate fpvar)
...
-            /* Microsoft Visual C++ 6.0 uses '%I64d' syntax  for 8-byte integers */
-            printf("\n File: %s, HDU %d,  %d cols X %I64d rows\n", infits, extnum, ncols, nrows);
-#elif (USE_LL_SUFFIX == 1)
-            printf("\n File: %s, HDU %d,  %d cols X %lld rows\n", infits, extnum, ncols, nrows);
-#else
-            printf("\n File: %s, HDU %d,  %d cols X %ld rows\n", infits, extnum, ncols, nrows);
-#endif
-            fp_test_table(inputfptr, outfptr, outfptr2, fpvar, &stat);
+            
+	    printf("\n File: %s, HDU %d,  %d cols X %" PRId64 " rows\n",
```
knro: We shouldn't have any printf in the library, so looks like we missed this before. Please remove it.

PR #2458, opened 2026-08-01
dome/nexdome_beaver: fix USB response truncation, add shutter and dome abort and rotator full-measurement

jaimealemany:
**USB/serial fixes:**
  - sendRawCommand(): flush stale input (tcflush) before resending a command on retry, so a delayed leftover reply from the previous attempt isn't misread as the answer to the retry.
  - echo(): guard against an unmatched version-string regex instead of crashing with std::length_error when the firmware returns an unexpected handshake response.
  - Handshake(): only enable tty_set_generic_udp_format() when the active connection is actually TCP/UDP. Previously it was set unconditionally, which broke serial reads whenever a reply arrived split across more than one USB packet (e.g. dome getaz, or the version handshake) — this was the actual root cause of intermittent USB truncation/hangs.
  (these issues surfaced when tested compatibility with the new Caterpillar controller)

**Rotator calibration:**
  - Add a "Full Measurement" button (autocalrot 2) alongside the existing Find Home / Measure Home controls - does one full rotation to measure steps/degree plus home position, then a home-measurement pass.
  - Fix rotatorFindHome()/rotatorMeasureHome()/rotatorFullMeasure() silently reporting success even when the firmware rejected the command (e.g. -101 when a calibration is already running) — these are now correctly reported as failures.
  - Add dedicated, always-available Abort buttons for the rotator and the shutter independently (!dome abort 0 1# / !dome abort 0 0 1#), since re-clicking a busy calibration switch to abort depends on client-widget click timing and isn't reliable.

**Tested with real hardware**

### DIFF
--- a/drivers/dome/nexdome_beaver.cpp
+++ b/drivers/dome/nexdome_beaver.cpp
@@ -301,7 +301,7 @@ bool Beaver::ISNewSwitch(const char *dev, const char *name, ISState *states, cha
         /////////////////////////////////////////////
         if (RotatorCalibrationSP.isNameMatch(name))
         {
-            bool wasBusy = RotatorCalibrationSP.getState() == IPS_BUSY;
+            bool wasMoving = getDomeState() == DOME_MOVING;
             RotatorCalibrationSP.update(states, names, n);
             bool rc = false;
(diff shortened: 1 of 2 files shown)


knro:
Great, thank you!

PR #2461, opened 2026-08-02
Fix warnings on modern gcc

returntoreality:
This fixes some gcc warnings I ran into, which caused the build to fail.

### DIFF
--- a/drivers/focuser/celestronauxpacket.cpp
+++ b/drivers/focuser/celestronauxpacket.cpp
@@ -184,7 +184,7 @@ bool Communicator::readPacket(int portFD, Packet &reply)
         return false;
     }
 
-    int len = rxbuf[0];
+    size_t len = static_cast<unsigned char>(rxbuf[0]);
     buffer packet;
     packet.reserve(2 + len + 1);
     packet.push_back(Packet::AUX_HDR);
--- a/drivers/telescope/pmc8driver.cpp
+++ b/drivers/telescope/pmc8driver.cpp
@@ -2194,7 +2194,7 @@ bool set_pmc8_radec(int fd, double ra, double dec)
 
 bool get_pmc8_coords(int fd, double &ra, double &dec, INDI::Telescope::TelescopePierSide &sop)
 {
-    int racounts, deccounts;
+    int racounts, deccounts = 0;
     bool rc;
 
     if (pmc8_simulation)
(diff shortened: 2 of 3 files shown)


drivers/telescope/astrotrac.h
```diff
@@ -139,13 +139,13 @@ class AstroTrac :
...
-                 * @param cmd_len if -1, it is assumed that the @a cmd is a null-terminated string. Otherwise, it would write @a cmd_len bytes from
+                 * @param cmd_len if PTRDIFF_MAX, it is assumed that the @a cmd is a null-terminated string. Otherwise, it would write @a cmd_len bytes from
                  * the @a cmd buffer.
-                 * @param res_len if -1 and if @a res is not nullptr, the function will read until it detects the default delimiter DRIVER_STOP_CHAR
+                 * @param res_len if PTRDIFF_MAX and if @a res is not nullptr, the function will read until it detects the default delimiter DRIVER_STOP_CHAR
                  *  up to DRIVER_LEN length. Otherwise, the function will read @a res_len from the device and store it in @a res.
                  * @return True if successful, false otherwise.
         */
-        bool sendCommand(const char * cmd, char * res = nullptr, int cmd_len = -1, int res_len = -1);
+        bool sendCommand(const char * cmd, char * res = nullptr, size_t cmd_len = PTRDIFF_MAX, size_t res_len = PTRDIFF_MAX);
```
knro: This change is dangerous. You changed to PTRDDIFF_MAX but still check for > 0 below. This essentially break all calls to this command.

returntoreality: I fixed the conditions and actually use DRIVER_MAX now.

drivers/telescope/astrotrac.cpp
```diff
@@ -1457,12 +1457,12 @@ bool AstroTrac::sendCommand(const char *cmd, char *res, int cmd_len, int res_len
...
 /////////////////////////////////////////////////////////////////////////////
-void AstroTrac::hexDump(char *buf, const char *data, int size)
+void AstroTrac::hexDump(char *buf, const char *data, size_t size)
 {
-    for (int i = 0; i < size; i++)
+    for (size_t i = 0; i < size; i++)
         sprintf(buf + 3 * i, "%02X ", static_cast<uint8_t>(data[i]));
 
-    if (size > 0)
+    if (size < DRIVER_LEN)
```
knro: What happens when size == 0 ? this would crash.

returntoreality: Good point, fixed it.

returntoreality:
@knro I dropped the changes from astrotrac, I think the rest is pretty uncontroversial

PR #2466, opened 2026-08-06
Add Proxisky UMi telescope driver (indi_lx200_proxisky)

nitr57:
## Summary

Adds `indi_lx200_proxisky`, a driver for [Proxisky](https://www.proxisky.com) UMi mounts (UMi17X, UMi20S and relatives).

These mounts run OnStep-derived firmware, so standard telescope behaviour is already handled
correctly by `LX200_OnStep`. What they add on top is a set of vendor-specific `:P...` commands that
stock OnStep does not implement - travel limits, an anti-collision system, servo PID gains and
several board settings. Until now those were unreachable from INDI and had to be set from Proxisky's
Windows-only tool.

`LX200_Proxisky` subclasses `LX200_OnStep` and adds **only** those commands. It overrides no motion,
tracking, guiding or parking behaviour.

Full user documentation is in `drivers/telescope/doc/lx200_proxisky/index.md`.

## Why a separate driver rather than a model check inside `lx200_OnStep`

Flagging this up front since it is the obvious review question.

- Per-variant subclassing is the established pattern here - `lx200generic.cpp` already dispatches
  ~20 classes on `getProgName()`, with no existing case of a variant folded in behind a model check.
- `LX200_OpenAstroTech`, the other OnStep-derived firmware in the tree, is likewise a separate driver.
- `lx200_OnStep.cpp` is already 5,555 lines; folding in ~1,900 more would put UMi-only commands and
  tabs in every OnStep user's path, gated on a runtime check that has to be correct.
- The `:P...` command space is entirely disjoint from OnStep's, so there is no shared logic to factor.

The other OnStep variants in `drivers.xml` (WarpAstron, EM31Pro, Juwei) point straight at
`indi_lx200_OnStep` precisely because they add nothing - which argues for the split, not against it.

## What it adds

19 properties across four `Proxisky *` tabs - Settings, Limits, ACS and Advanced - covering travel
limits, anti-collision (enable, thresholds, collision counters), servo PID gains per axis,
Supervised GOTO/Home, and board settings such as the status LED and power-loss memory.

Each control is published **only if the mount answers the corresponding query**, so the panel
reflects what the connected model actually supports. Every write is read back and verified; PID
writes are refused unless the mount is stopped, because the commit restarts the axis.

## Files

| File | Change |
|---|---|
| `drivers/telescope/lx200_proxisky.cpp` | new, 1925 lines |
| `drivers/telescope/lx200_proxisky.h` | new, 199 lines |
| `drivers/telescope/doc/lx200_proxisky/index.md` | new - user documentation |
| `drivers/telescope/CMakeLists.txt` | add source + install symlink |
| `drivers/telescope/lx200generic.cpp` | add dispatch branch |
| `drivers.xml` | point the existing Proxisky UMi entry at the new driver |

## ⚠️ Migration note for existing UMi users

The `Proxisky / UMi` entry in `drivers.xml` already existed and pointed at `indi_lx200_OnStep`. This
PR repoints it at `indi_lx200_proxisky` and corrects the manufacturer spelling (`ProxiSky` ->
`Proxisky`).

That changes the device name from `LX200 OnStep` to `Proxisky UMi`, and INDI keys saved state by
device name - so `~/.indi/<name>_config.xml` and the `<device name=…>` block in `ParkData.xml` are
orphaned. Existing users need to set port, configuration and park position once more on first
connect. Nothing is deleted, and selecting `LX200 OnStep` manually still works as before.

If you would rather avoid the migration, the alternative is a second catalogue entry leaving the
existing one alone - happy to change it.

## Testing

Verified end to end against a **Proxisky UMi20S, firmware 1.0.6, serial 9600 8N1**: connect /
reconnect / disconnect, first-connect property publication (19/19), capability probing, every write
path, setter refusals, out-of-range rejection on all five numeric properties, the PID gate and its
post-commit axis-restart readback.

Builds clean; `astyle` with the project's options reports zero diff.

**Not tested:** models other than the UMi20S - the driver probes per feature rather than assuming a
model, so a UMi17X should expose a superset, unverified.

**Simulation mode exposes no vendor properties**, by design - every vendor property mirrors a value
read back from hardware. Documented in `index.md` so it is not filed as a bug.

## Note for reviewers

The capability probes run from `updateProperties()` rather than `getBasicData()`. This is deliberate
and load-bearing: `LX200_OnStep::updateProperties()` reaches `getBasicData()` *before* it selects the
transport timeout, so probing there would use the 100 ms serial default on a TCP link and hide every
vendor property on first connect.

### DIFF
--- a/drivers.xml
+++ b/drivers.xml
@@ -43,9 +43,9 @@
             <driver name="LX200 OnStep">indi_lx200_OnStep</driver>
             <version>1.27</version>
         </device>
-        <device label="UMi" manufacturer="ProxiSky">
-            <driver name="LX200 OnStep">indi_lx200_OnStep</driver>
-            <version>1.27</version>
+        <device label="UMi" manufacturer="Proxisky">
+            <driver name="Proxisky UMi">indi_lx200_proxisky</driver>
+            <version>1.0</version>
         </device>
         <device label="EM31Pro" manufacturer="Emcan Astro">
             <driver name="LX200 OnStep">indi_lx200_OnStep</driver>
--- a/drivers/telescope/CMakeLists.txt
+++ b/drivers/telescope/CMakeLists.txt
@@ -77,7 +77,8 @@ add_executable(indi_lx200generic
     ioptronHC8406.cpp
     eq500x.cpp
     lx200_esp32go.cpp
-    toup_sh20.cpp)
+    toup_sh20.cpp
+    lx200_proxisky.cpp)
 
 if(CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
     target_compile_definitions(indi_lx200generic PRIVATE -D_XOPEN_SOURCE=700)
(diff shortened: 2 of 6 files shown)


knro:
Thank you! Please add the driver documentation to [INDI Documentation Repository](https://github.com/indilib/drivers-docs) as it is required for all new drivers.

knro:
@nitr57 Can you please submit the driver documentation as requested above? Thanks!

knro:
@nitr57 All OK? Do you require any help with the documentation?

knro:
@s-pouya-sh Can you please create a stub for this?

s-pouya-sh:
Done
https://drivers.indilib.org/mounts/proxisky/umi/umi/

PR #2489, opened 2026-09-08
Add Homebrew Core submission workflow

deb-mohan:
Adds a server-only indiserver formula, macOS Homebrew validation, and automation that opens or updates a Homebrew Core pull request on release.\n\nAI-assisted contribution: OpenAI Codex (GPT-5) assisted with preparing this change. I have reviewed it and will respond to maintainer feedback.


rlancaste:
I was just taking a look over this.  We have built INDI with homebrew for a number of years.  When I first ported KStars to MacOS that was our only way to build it all and we used that for years until we started doing it with craft.  We still do use homebrew for testing code regularly, so it is still very useful.  I agree that having a homebrew recipe would be very useful.

I notice that you said (and also I noticed in the code) that it does not build the drivers?  I am confused by why this would be the case.  INDI Server needs to build and install the drivers to be useful.  Also, the "third party" drivers are almost as important as the core INDI drivers.  Is there a reason that you chose not to do so?

PR #2492, opened 2026-09-20
Fix alpaca dome

44z6b:
Rewriting of the indi_alpaca_dome driver in order to fully comply with Alpaca API specifications for dome-type devices (rather than merely providing partial support for a simple roll-off roof).

This has been fully tested against a dome simulator (Alpaca Interface version 1), and partially with a real Alpaca compliant specific dome (Alpaca Interface version 3, from the Saint-Veran Paul Felenbok Observatory, France). The driver still needs some improvements and it seems to be fully operational. Already planned add-ons : HTTPS support with several available security options, API authentication (at least API key support)... Maybe an extension of the Connection namespace might be preferred for HTTP(S) support.


knro:
Thanks, why did you remove CAN_PARK from the capabilities?

44z6b:
> Thanks, why did you remove CAN_PARK from the capabilities?

I removed CAN_PARK from the set of capabilities in the constructor since this one is returned by the /canpark API call. All the capabilities reported by the remote device are now set in a single step within the Connect method. This does not apply to CAN_ABORT which is the only one flag I kept in the constructor. But your comment suggests me that maybe CAN_ABORT could also be integrated into Connect, alongside CAN_ABS_MOVE, CAN_REL_MOVE or HAS_SHUTTER (and depending on them). I don't think it's essential but maybe it would be more consistent. Let me know.

